PR #25, opened 2017-09-20
Support for auto generation of spec objects from a spec dir and misc changes

harsh-px:
1) Add support for parsing application specs from a spec dir
2) Add missing support for Service 
3) Change torpedo k8s deployment to use pod instead of jobs to avoid restart issue 
4) Add postgres and nginx apps support 
5) Add missing hostkey issue for ssh node driver 
6) Perform deep copy of spec objects in factory and modify clones during schedule


harsh-px:
I have kept the PR open but merged the changes to hasten the process.

PR #34, opened 2017-09-20
Ssh changes

varunturlapati:
I have fixed the connectivity part and added an attribute to track a usable address instead of querying each time.
I also modified task to now return string and error to support returning stdout/any output and an error.

Before diverging much further I think we should do a review.


harsh-px:
@varunturlapati  The travis build failed.

drivers/node/node.go
```diff
@@ -20,23 +20,32 @@ const (
...
+type ConnectionOpts struct {
+	Timeout         time.Duration
+	TimeBeforeRetry time.Duration
 }
 
 // RebootNodeOpts provide additional options for reboot operation
 type RebootNodeOpts struct {
-	Force bool
+	Force          bool
+	ConnectionOpts ConnectionOpts
```
harsh-px: Use an anonymous filed here. So something like
```
type RebootNodeOpts struct {
 	Force bool
        ConnectionOpts
}
```
So users can use it as ops.Timeout instead of ops.ConnectionOpts.Timeout.

drivers/node/node.go
```diff
@@ -20,23 +20,32 @@ const (
...
 }
 
 // ShutdownNodeOpts provide additional options for shutdown operation
 type ShutdownNodeOpts struct {
-	Force bool
+	Force          bool
+	ConnectionOpts ConnectionOpts
```
harsh-px: Same as above

drivers/node/ssh/ssh.go
```diff
@@ -73,31 +107,33 @@ func (s *ssh) Init(sched string) error {
...
 		return &ErrFailedToTestConnection{
 			Node:  n,
 			Cause: fmt.Sprintf("failed to get node address due to: %v", err),
 		}
 	}
 
-	t := func() error {
-		return s.doCmd(addr, "hostname", false)
-	}
+	/*
```
harsh-px: Since this is dead code, this should not stay here commented. Delete the lines.

drivers/node/ssh/ssh.go
```diff
@@ -178,66 +216,88 @@ func (s *ssh) doCmd(addr string, cmd string, ignoreErr bool) error {
...
+	logrus.Infof("STDOUT set here")
 
 	chOut := make(chan string)
 	go func() {
 		var bufout bytes.Buffer
 		io.Copy(&bufout, stdout)
 		chOut <- bufout.String()
 	}()
-
+	logrus.Infof("go func loop here")
```
harsh-px: Remove all log statements used for development/debugging.

drivers/node/ssh/ssh.go
```diff
@@ -178,66 +216,88 @@ func (s *ssh) doCmd(addr string, cmd string, ignoreErr bool) error {
...
+	out = fmt.Sprintf("%v", byteout)
+	if ignoreErr == false && err != nil {
+		return out, &ErrFailedToRunCommand{
 			Addr:  addr,
 			Cause: fmt.Sprintf("failed to run command due to: %v", err),
 		}
 	}
-
-	return nil
+	logrus.Infof(fmt.Sprintf("Command %s has output %s", cmd, out))
```
harsh-px: No need to print this. We should only print errors/warnings and important events in the test. Not traces.

drivers/node/ssh/ssh.go
```diff
@@ -178,66 +216,88 @@ func (s *ssh) doCmd(addr string, cmd string, ignoreErr bool) error {
...
+	return addr, err
+}
+
+func (s *ssh) getOneUsableAddr(n node.Node, options node.ConnectionOpts) (string, error) {
+	logrus.Infof("Here are the addresses: %v", n.Addresses)
+	for _, addr := range n.Addresses {
+		t := func() (string, error) {
+			return s.doCmd(addr, "hostname", false)
+		}
+		logrus.Infof("Will try task with addr: %s", addr)
```
harsh-px: Remove this log line

drivers/node/ssh/ssh.go
```diff
@@ -178,66 +216,88 @@ func (s *ssh) doCmd(addr string, cmd string, ignoreErr bool) error {
...
 	}
 
-	addr := n.Addresses[0] // TODO don't stick to first address
-	return addr, nil
+	addr, err := s.getOneUsableAddr(n, options)
+	return addr, err
+}
+
+func (s *ssh) getOneUsableAddr(n node.Node, options node.ConnectionOpts) (string, error) {
+	logrus.Infof("Here are the addresses: %v", n.Addresses)
```
harsh-px: Remove

pkg/task/task.go
```diff
@@ -2,42 +2,48 @@ package task
...
 			default:
-				err := t()
+				out, err := t()
 				if err == nil {
+					logrus.Infof("Task done: %v\nOutput is: %s", t, out)
 					done <- true
 					return
 				}
-
+				logrus.Infof("Will retry task")
```
harsh-px: Remove. This is going to just flood the logs on each retry,

pkg/task/task.go
```diff
@@ -2,42 +2,48 @@ package task
...
 				if q {
+					logrus.Infof("Timed out, quitting")
 					return
 				}
 
 			default:
-				err := t()
+				out, err := t()
 				if err == nil {
+					logrus.Infof("Task done: %v\nOutput is: %s", t, out)
```
harsh-px: Remove

pkg/task/task.go
```diff
@@ -2,42 +2,48 @@ package task
...
 )
 
+//TODO: export the type: type Task func() (string, error)
+
 // ErrTimedOut is returned when an operation times out
 var ErrTimedOut = errors.New("timed out performing task")
 
 // DoRetryWithTimeout performs given task with given timeout and timeBeforeRetry
-func DoRetryWithTimeout(t func() error, timeout, timeBeforeRetry time.Duration) error {
+func DoRetryWithTimeout(t func() (string, error), timeout, timeBeforeRetry time.Duration) (string, error) {
```
harsh-px: Why do we need `DoRetryWithTimeout(...)` to return a string output? Unless we have a requirement, let's keep the interface minimal.

harsh-px: Change returning type to `interface{}` instead of a string.

harsh-px:
Please go over all review comments.

varunturlapati:
Updated the PR after resolving conflicts and pegging down the ssh package version in glide

drivers/node/ssh/ssh.go
```diff
@@ -138,11 +160,11 @@ func (s *ssh) ShutdownNode(n node.Node, options node.ShutdownNodeOpts) error {
 		shutdownCmd = "halt"
 	}
 
-	t := func() error {
+	t := func() (interface{}, error) {
```
harsh-px: These `t` functions don't have to have return type as interface{}. Only the generic `DoRetryWithTimeout` task package function should have generic return types. `t` is a user function that knows it's returning type is a string. So `t` should have a return type `(string, error)`

Do the same for other places in the review.

drivers/node/ssh/ssh.go
```diff
@@ -198,46 +220,59 @@ func (s *ssh) doCmd(addr string, cmd string, ignoreErr bool) error {
...
+}
+
+func (s *ssh) getOneUsableAddr(n node.Node, options node.ConnectionOpts) (string, error) {
+	for _, addr := range n.Addresses {
+		t := func() (interface{}, error) {
+			return s.doCmd(addr, "hostname", false)
+		}
+		if _, err := task.DoRetryWithTimeout(t, options.Timeout, options.TimeBeforeRetry); err == nil {
+			n.UsableAddr = addr
+			logrus.Infof("Updated field UsableAddr. Now node is %#v", n)
```
harsh-px: Remove trace log.

harsh-px:
This PR has changes already in master.

Rebase using `git rebase -i origin/master` to isolate changes only in this PR.

drivers/node/ssh/ssh.go
varunturlapati: I think that should be done on a case to case basis because currently none of the user tasks t() are using the string part yet.

harsh-px: An `interface{}` should be used when you defined a generic API or struct that users will consume and you don't know what data type you might want to use. `t` here is a function variable, neither an API nor a generic data structure. API's like `DoRetryWithTimeout` are good candidates of using `interface{}` as return type since it provides a generic functionality and doesn't make assumptions about what callers might want.

In 
```
 t := func() (interface{}, error) {
      return s.doCmd(addr, shutdownCmd, true)
} 
```

You exactly know the return type since  `s.doCmd(...)` returns the string type. Infact, any other return type when defining `t` won't work.

While the code does compile, one shouldn't use `interface{}` as a placeholder for all types when we exactly know the returning type.

adityadani:
Instead of merging branches. It would be easier if you rebase your private branch with master.
This PR now has 40 or so commits and 13000 changed files. It becomes very hard to isolate your changes and review them.
I usually follow this procedure

```
$ git checkout -b <new_private_branch>
$ git commit 1
.....
$ git commit 2
$ git push <new_private_branch>
....
// Merge changes from master
// On branch <new_private_branch>

$ git rebase origin/master
// This will rebase all your commits on top of master commits. You might have to resolve conflicts here. After resolving conflicts

$ git push -f <new_private_branch>
```

varunturlapati:
Rebase privately hasn't helped. I went through the cycle 2 times resolving each intermediate commit.  I just declined this PR. I'll create a new one with a new branch name.

PR #35, opened 2017-09-21
Aws initial

jessicapwx:


harsh-px:
I'll continue review later tonight.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil
+}
+
+func (aws *aws) TestConnection(n node.Node, options node.TestConectionOpts) error {
+	return nil
+}
+
+func (aws *aws) RebootNode(n node.Node, options node.RebootNodeOpts) error {
+	var err error
+	_, err = aws.getAddrToConnect(n)
```
harsh-px: `getAddrToConnect` isn't required for the aws driver. There isn't any useful logic being performed by this routine. Remove it.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	reservations := resp.Reservations
+	//If resp is empty, then return empty []
+	instances := []*ec2.Instance{}
+	for _, resv := range reservations {
+		for _, ins := range resv.Instances {
+			instances = append(instances, ins)
+		}
+	}
+	fmt.Printf("Found %d instances in %s\n", len(instances), AwsRegion)
```
harsh-px: Don't print traces. Remove.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (aws *aws) getAddrToConnect(n node.Node) (string, error) {
+	if n.Addresses == nil || len(n.Addresses) == 0 {
+		return "", fmt.Errorf("no address available to connect")
+	}
+	addr := n.Addresses[0] // TODO don't stick to first address
+	return addr, nil
+}
+
+func (aws *aws) getAllInstancesInEast() []*ec2.Instance {
```
harsh-px: Why does this function name say East? Within the function, there is nothing specific about the east region.

In general, the code needs to be generic for aws and variables like regions should be provided during init.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	params := &ec2.DescribeInstancesInput{}
+	resp, err := aws.svc.DescribeInstances(params)
+	if err != nil {
+		fmt.Println("there was an error listing instances in", AwsRegion, err.Error())
+		log.Fatal(err.Error())
```
harsh-px: Remove all log.Fatal. Replace with just returning error objects.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+		for _, ins := range resv.Instances {
+			instances = append(instances, ins)
+		}
+	}
+	fmt.Printf("Found %d instances in %s\n", len(instances), AwsRegion)
+	return instances
+}
+
+func (aws *aws) getNodeIDByPrivAddr(n node.Node) (string, error) {
+	fmt.Printf("Get instance ID of node whose private ip is %s\n", n.Addresses[0])
```
harsh-px: Remove all traces.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, i := range aws.Instances {
+		if aws_pkg.StringValue(i.PrivateIpAddress) == n.Addresses[0] {
```
harsh-px: Fragile code. We cannot simply compare to `n.Addresses[0]`. Since the AWS private IP can easily be the 2nd, 3rd, ... address. Compare with all `n.Addresses`.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+			return aws_pkg.StringValue(i.InstanceId), nil
+		}
+	}
+	return "none", fmt.Errorf("Failed to get instanceID by privateIP")
```
harsh-px: Don't send `"none"`. Send an empty string rather.

Sending special strings like "none" defines custom protocol between callers and callee which can be unintuitive. A rather standard protocol is callers always check for error objects which you are already returning. It is also intuitive for callers to check for empty strings.

drivers/node/aws/errors.go
```diff
@@ -0,0 +1,16 @@
+package aws
```
harsh-px: Merge `ssh/errors.go` and `aws/errors.go` and move it to `node/errors.go`.

This will remove redundant error objects.

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"fmt"
+	"github.com/portworx/torpedo/drivers/node"
+	//"github.com/portworx/torpedo/drivers/scheduler"
+	aws_pkg "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"log"
```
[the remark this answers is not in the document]
jessicapwx: fixed

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
+package aws
+
+import (
+	"fmt"
+	"github.com/portworx/torpedo/drivers/node"
+	//"github.com/portworx/torpedo/drivers/scheduler"
```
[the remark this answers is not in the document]
jessicapwx: fixed

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	AwsRegion = "us-east-1"
+)
+
+type aws struct {
+	node.Driver
+	session     *session.Session
+	credentials *credentials.Credentials
+	config      *aws_pkg.Config
+	svc         *ec2.EC2
+	Instances   []*ec2.Instance
```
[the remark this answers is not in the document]
jessicapwx: All us-east-1 instances would be save as aws.Instances, which would be used for any specific instance searching

drivers/node/aws/aws.go
jessicapwx: removed

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	_, err = aws.getAddrToConnect(n)
+	if err != nil {
+		return &ErrFailedToRebootNode{
+			Node:  n,
+			Cause: fmt.Sprintf("failed to get node address due to: %v", err),
+		}
+	}
+	instanceID, err := aws.getNodeIDByPrivAddr(n)
+	if err != nil {
+		fmt.Println("there was an error get instance ID by its private IP ", n.Addresses[0], err.Error())
```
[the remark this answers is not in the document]
jessicapwx: log printing removed, if err!= nil then return ErrFailedToRebootNode

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+		log.Fatal(err.Error())
```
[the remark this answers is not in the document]
jessicapwx: fixed

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	//Reboot the instance by its InstanceID
+	rebootInput := &ec2.RebootInstancesInput{
+		InstanceIds: []*string{
+			aws_pkg.String(instanceID),
+		},
+	}
+	fmt.Printf("Now reboot instance %+v\n", rebootInput)
```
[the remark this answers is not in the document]
jessicapwx: printf removed

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,128 @@
...
+	_, err = aws.svc.RebootInstances(rebootInput)
+	if err != nil {
+		return fmt.Errorf("there was an error reboot instances %s, %q", instanceID, err.Error())
```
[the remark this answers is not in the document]
jessicapwx: return ErrFailedToRebootNode now. Fixed

drivers/node/aws/aws.go
jessicapwx: Changed to getAllInstances()

drivers/node/aws/aws.go
jessicapwx: removed

drivers/node/aws/aws.go
jessicapwx: removed

drivers/node/aws/aws.go
jessicapwx: removed

drivers/node/aws/errors.go
jessicapwx: merged

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,119 @@
...
+	instances := aws.getAllInstances()
+	aws.instances = instances
+	return nil
+}
+
+func (aws *aws) TestConnection(n node.Node, options node.TestConectionOpts) error {
+	return nil
+}
+
+func (aws *aws) RebootNode(n node.Node, options node.RebootNodeOpts) error {
```
harsh-px: Change `aws` to `a` so we distinguish the instance and the struct name.

So, it'll be

`func (a *aws) ...`

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,119 @@
...
+func (aws *aws) ShutdownNode(n node.Node, options node.ShutdownNodeOpts) error {
+	return nil
+}
+
+func (aws *aws) getAllInstances() []*ec2.Instance {
+	params := &ec2.DescribeInstancesInput{}
+	resp, err := aws.svc.DescribeInstances(params)
+	if err != nil {
+		logrus.Errorf("there was an error listing instances in %s. Error: %q", aws.region, err.Error())
+		logrus.Fatal(err.Error())
```
harsh-px: No helper routine just kill the test. Helper routines should only return error objects.

So remove logrus and return error. Change the function definition to `getAllInstances() ([]*ec2.Instance, error)`

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (aws *aws) getNodeIDByPrivAddr(n node.Node) (string, error) {
+	for _, i := range aws.instances {
+		for _, addr := range n.Addresses {
+			if aws_pkg.StringValue(i.PrivateIpAddress) == addr {
+				return aws_pkg.StringValue(i.InstanceId), nil
+			}
+		}
+	}
+	return "", fmt.Errorf("Failed to get instanceID by privateIP")
```
harsh-px: Include the node name (`n.Name`) here in the Errorf.

drivers/node/aws/aws.go
jessicapwx: changed func(aws *aws) to func(a *aws)

drivers/node/aws/aws.go
jessicapwx: fixed

drivers/node/aws/aws.go
jessicapwx: fixed as
return "", fmt.Errorf("Failed to get instanceID of %s by privateIP", n.Name)

drivers/node/aws/aws.go
harsh-px: I don't see that updated here yet. Note that all of them need to have `func(a *aws)`

jessicapwx: not checked-in yet, I would check in soon

drivers/node/aws/aws.go
```diff
@@ -0,0 +1,123 @@
...
+func (a *aws) ShutdownNode(n node.Node, options node.ShutdownNodeOpts) error {
+	return nil
+}
+
+func (a *aws) getAllInstances() ([]*ec2.Instance, error) {
+	instances := []*ec2.Instance{}
+	params := &ec2.DescribeInstancesInput{}
+	resp, err := a.svc.DescribeInstances(params)
+	if err != nil {
+		logrus.Errorf("there was an error listing instances in %s. Error: %q", a.region, err.Error())
```
harsh-px: Don't log here. We want to have logging done only by higher routines and not low-level helpers. Return fmt.Errorf that includes this custom message and the original error.

jessicapwx: changed to
 return instances, fmt.Errorf("there was an error listing instances in %s. Error: %q", a.region, err.Error())

Issue #11, opened 2017-09-13
Add command line argument parser for torpedo


harsh-px:
Use a command line arg parser that has flags to take in scheduler, volume driver, testname, version, help etc.

harsh-px:
@varunturlapati Can you wrap this up? Reassigning from @jessicapwx as she appears occupied with aws work.

adityadani:
Done!

PR #37, opened 2017-09-21
Add a cli interface to torpedo.

adityadani:
#11

### DIFF
--- a/cmd/torpedo/torpedo.go
+++ b/cmd/torpedo/torpedo.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
 	"github.com/portworx/torpedo/drivers/node"
 	_ "github.com/portworx/torpedo/drivers/node/ssh"
 	"github.com/portworx/torpedo/drivers/scheduler"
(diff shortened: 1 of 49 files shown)


cmd/torpedo/torpedo.go
```diff
@@ -592,39 +593,74 @@ func (t *torpedo) run(testName string) error {
...
+				},
+				cli.StringFlag{
+					Name:  "node-driver,n",
+					Usage: "Name of the node driver to use",
+					Value: "ssh",
+				},
+				cli.StringFlag{
+					Name:  "test,t",
+					Usage: "Name of the test to run",
+					Value: "all",
```
harsh-px: The default value is all? If it's all, the current torpedo will not run any test. It runs all tests if `testName` is empty.

Also, this should be tests instead of test and have a csv value which we can parse into list of tests.

cmd/torpedo/torpedo.go
```diff
@@ -592,39 +593,74 @@ func (t *torpedo) run(testName string) error {
...
+				},
+			},
+		},
 	}
+	app.Run(os.Args)
+
+}
+
+func fireTorpedo(c *cli.Context) {
+	s := c.String("scheduler")
```
harsh-px: Better if the cli flags like "scheduler" are constants that are also referenced while creating the cli flags above. So if someone changes one, other changes automatically.

adityadani: addressed in subsequent commit.

PR #40, opened 2017-09-27
Gather support bundle - part1

varunturlapati:
Hooks in place so far.


cmd/torpedo/torpedo.go
```diff
@@ -26,13 +27,15 @@ const (
 	nodeDriverCliFlag    = "node-driver,n"
 	storageDriverCliFlag = "storage,v"
 	testsCliFlag         = "tests,t"
+	logLocationCliFlag   = "logLocation"
```
piyush-nimbalkar: we should have the param name as log-location as this is a cli param and to make it consistent with other params

cmd/torpedo/torpedo.go
```diff
@@ -335,15 +338,32 @@ func (t *torpedo) validateContext(ctx *scheduler.Context) error {
...
 	}
 	if err = t.s.WaitForRunning(ctx); err != nil {
+		GenerateSupportBundle(t, ctx)
 		return err
 	}
 	return err
 }
 
+// GenerateSupportBundle gathers logs and any artifacts pertinent to the scheduler and dumps them in the defined location
+func GenerateSupportBundle(t *torpedo, ctx *scheduler.Context) error {
```
piyush-nimbalkar: why not make this a method on torpedo instance? also it could be a private method, right?

cmd/torpedo/torpedo.go
```diff
@@ -335,15 +338,32 @@ func (t *torpedo) validateContext(ctx *scheduler.Context) error {
...
+	if t == nil || ctx == nil || t.s == nil {
+		return fmt.Errorf("Invalid context or torpedo object. Cannot generate support bundle")
+	}
+	out, err := t.s.GenerateSupportBundle(ctx)
+	err = ioutil.WriteFile(fmt.Sprintf("%s/supportbundle%v.log", t.logLoc, time.Now().Format(time.RFC3339)), []byte(out), 0644)
+	if err != nil {
+		return fmt.Errorf("Couldn't write the support bundle file because of err: %v", err)
+	}
+	return nil
+
```
piyush-nimbalkar: nit: extra line?

drivers/scheduler/k8s/k8s.go
```diff
@@ -616,6 +616,72 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
 	return result, nil
 }
 
+func (k *k8s) GenerateSupportBundle(ctx *scheduler.Context) (string, error) {
+	var output string
+	var err error
+	for _, core := range ctx.App.SpecList {
```
piyush-nimbalkar: now that we don't have a core and storage component in the app spec, should we call this just spec?

drivers/scheduler/k8s/k8s.go
```diff
@@ -616,6 +616,72 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
...
+				for _, conStat := range pod.Status.ContainerStatuses {
+					output += insertLineBreak(conStat.Name)
+					output += fmt.Sprintf("%v\n", conStat)
+					output += insertLineBreak("END Container")
+				}
+				output += insertLineBreak("END Pod")
+			}
+			output += insertLineBreak("END StatefulSet")
+		} else {
+			output += insertLineBreak("UNKNOWN/UNSUPPORTED OBJ TYPE")
```
piyush-nimbalkar: you will have to remove the else block because now we are parsing the entire list of specs which include storage specs. They are not from the above list but still valid specs. We can rely on the parser that it is not going to let unsupported objects in the scheduler.

pkg/k8sutils/k8sutils.go
```diff
@@ -504,14 +528,21 @@ func DeletePods(pods []v1.Pod) error {
...
 
-// GetPodsByOwner returns pods for the given owner and namespace
-func GetPodsByOwner(ownerName string, namespace string) ([]v1.Pod, error) {
+//GetPods returns pods for the given namespace
+func GetPods(namespace string) (*v1.PodList, error) {
+	var pods *v1.PodList
 	client, err := GetK8sClient()
 	if err != nil {
-		return nil, err
+		return pods, err
```
piyush-nimbalkar: what is the empty value of this? why did we remove nil?

cmd/torpedo/torpedo.go
varunturlapati: Will do.

cmd/torpedo/torpedo.go
varunturlapati: What are we trying to prevent accessing this? IMO In fact this should be available eventually if we expose APIs outside too.

drivers/scheduler/k8s/k8s.go
varunturlapati: Will do.

drivers/scheduler/k8s/k8s.go
varunturlapati: We have the entire list but we don't have the handling for all kinds yet. That handling means more of the APIs for those kinds need to be developed and integrated. That should be a separate task.

cmd/torpedo/torpedo.go
```diff
@@ -336,14 +339,29 @@ func (t *torpedo) validateContext(ctx *scheduler.Context) error {
 	}
 
 	if err := t.validateVolumes(ctx); err != nil {
+		generateSupportBundle(t, ctx)
 		return err
 	}
 	if err = t.s.WaitForRunning(ctx); err != nil {
+		generateSupportBundle(t, ctx)
```
harsh-px: Add a warning log if failed to generate support bundle.

drivers/scheduler/k8s/k8s.go
harsh-px: What @piyush-nimbalkar means to say is `SpecList` will contain all sorts of spec (including `PersistentVolumeClaim`, `StorageClass`. This change was made recently. So your current code, will cause that to come to this else case and add the unsupported type message. So remove this else block.

drivers/scheduler/k8s/k8s.go
```diff
@@ -616,6 +616,72 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
...
+			if depStatus, err = k8sutils.DescribeDeployment(obj); err != nil {
+				output += fmt.Sprintf("%v", &ErrFailedToValidateApp{
+					App:   ctx.App,
+					Cause: fmt.Sprintf("Failed to get status of deployment: %v. Err: %v", obj.Name, err),
+				})
+			}
+			//Dump depStatus
+			output += fmt.Sprintf("%v\n", depStatus)
+			pods, _ := k8sutils.GetDeploymentPods(obj)
+			for _, pod := range pods {
```
harsh-px: Create common function for generating output from a list of pods. This block and line 658..667 have common code.

drivers/scheduler/k8s/k8s.go
```diff
@@ -616,6 +616,72 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
...
+			}
+			output += insertLineBreak("END StatefulSet")
+		} else {
+			output += insertLineBreak("UNKNOWN/UNSUPPORTED OBJ TYPE")
+			output += fmt.Sprintf("%v\n", &ErrFailedToValidateApp{
+				App:   ctx.App,
+				Cause: fmt.Sprintf("Unsupported app type: %v", obj),
+			})
+			output += insertLineBreak("END UNSUPPORTED OBJ TYPE")
+		}
```
harsh-px: We need to dump PVC and Service status too.

drivers/scheduler/k8s/k8s.go
```diff
@@ -616,6 +616,72 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
...
 
+func (k *k8s) GenerateSupportBundle(ctx *scheduler.Context) (string, error) {
+	var output string
+	var err error
+	for _, spec := range ctx.App.SpecList {
+		if obj, ok := spec.(*apps_api.Deployment); ok {
+			output += insertLineBreak(obj.Name)
+			var depStatus apps_api.DeploymentStatus
+			if depStatus, err = k8sutils.DescribeDeployment(obj); err != nil {
+				output += fmt.Sprintf("%v", &ErrFailedToValidateApp{
```
harsh-px: Why is `GenerateSupportBundle` returning error type for validating app? It should have it's own error type.

cmd/torpedo/torpedo.go
```diff
@@ -657,6 +680,7 @@ func fireTorpedo(c *cli.Context) {
 	v := c.String("storage")
 	n := c.String("node-driver")
 	tests := c.String("tests")
+	logLoc := c.String("logLocation")
```
harsh-px: The flag name appears to be `log-location`. Why are you getting the value from `logLocation`?

cmd/torpedo/torpedo.go
```diff
@@ -667,12 +691,15 @@ func fireTorpedo(c *cli.Context) {
 	} else if nodeDriver, err := node.Get(n); err != nil {
 		logrus.Fatalf("Cannot find node driver for %v. Err: %v\n", n, err)
 		os.Exit(-1)
+	} else if err := os.MkdirAll(logLoc, os.ModeDir); err != nil {
+		logrus.Fatalf("Cannot create path %s for saving support bundle. Error was: %v", logLoc, err)
```
harsh-px: nit: Instead of "Error was" just say "Err:".

drivers/scheduler/k8s/k8s.go
```diff
@@ -616,6 +616,72 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
 	return result, nil
 }
 
+func (k *k8s) GenerateSupportBundle(ctx *scheduler.Context) (string, error) {
+	var output string
```
harsh-px: Strings in golang are immutable. So this routine is going to allocate a new string each time you have `output +=`.

Use a byte array. Reference: https://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go

drivers/scheduler/k8s/k8s.go
varunturlapati: @harsh-px my original intent as I explained to @piyush-nimbalkar was to leave a log in place that points out an unsupported object and so we can identify how to add support to that. Currently PVC and Service are two such objects but when I did started this branch we hasn't yet thought of these objects.
To dump the status of these as well we will have to add underlying status calls. I can do that but I hadn't accounted for that for this effort. I'll work on adding those too.

drivers/scheduler/scheduler.go
```diff
@@ -69,6 +69,9 @@ type Driver interface {
 
 	// DeleteVolumes will delete a storage volume.
 	DeleteVolumes(*Context) error
+
+	// GenerateSupportBundle generates a bundle that can be used by support - logs, cores, states, etc
+	GenerateSupportBundle(*Context) (string, error)
```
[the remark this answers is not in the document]
varunturlapati: But Describe() is a k8s specific term which doesn't imply to anyone that possibly a log file will be generated and dumped somewhere. How will that make it look generic?

cmd/torpedo/torpedo.go
varunturlapati: Missed it. My bad.

cmd/torpedo/torpedo.go
varunturlapati: Done

cmd/torpedo/torpedo.go
```diff
@@ -645,6 +663,11 @@ func main() {
 					Usage: "Comma-separated list of tests. [Default:  runs all the tests]",
 					Value: "",
 				},
+				cli.StringFlag{
+					Name:  logLocationCliFlag,
+					Usage: "Path to save logs/artifacts upon failure. Default: /mnt/tsb",
+					Value: "/mnt/tsb",
```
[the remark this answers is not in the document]
varunturlapati: I thought 'torpedosupportbundle' will be too long. From user perspective 'tsb' is easy to type and/or tab complete.

I can change it to 'torpedo-support-dir'

cmd/torpedo/torpedo.go
```diff
@@ -336,14 +339,29 @@ func (t *torpedo) validateContext(ctx *scheduler.Context) error {
...
 	return err
 }
 
+// generateSupportBundle gathers logs and any artifacts pertinent to the scheduler and dumps them in the defined location
+func generateSupportBundle(t *torpedo, ctx *scheduler.Context) error {
+	if t == nil || ctx == nil || t.s == nil {
+		return fmt.Errorf("Invalid context or torpedo object. Cannot generate support bundle")
+	}
+	out, err := t.s.GenerateSupportBundle(ctx)
+	err = ioutil.WriteFile(fmt.Sprintf("%s/supportbundle%v.log", t.logLoc, time.Now().Format(time.RFC3339)), []byte(out), 0644)
```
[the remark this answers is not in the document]
varunturlapati: Will do. I still want to leave the timestamp to disambiguate in case of multiple files, perhaps, at a future point or for merely a chronology reference.

cmd/torpedo/torpedo.go
```diff
@@ -336,14 +339,29 @@ func (t *torpedo) validateContext(ctx *scheduler.Context) error {
...
 	}
 	if err = t.s.WaitForRunning(ctx); err != nil {
+		generateSupportBundle(t, ctx)
 		return err
 	}
 	return err
 }
 
+// generateSupportBundle gathers logs and any artifacts pertinent to the scheduler and dumps them in the defined location
+func generateSupportBundle(t *torpedo, ctx *scheduler.Context) error {
```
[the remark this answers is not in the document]
varunturlapati: Just curious: Outside of syntactical consistency is there a reason to do that?

cmd/torpedo/torpedo.go
varunturlapati: Done.

pkg/k8sutils/k8sutils.go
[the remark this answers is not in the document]
varunturlapati: That was how it was but then the code doesn't compile with an error like the below:
../../pkg/k8sutils/k8sutils.go:356: cannot use nil as type "github.com/portworx/torpedo/vendor/k8s.io/client-go/pkg/apis/apps/v1beta1".DeploymentStatus in return argument

../../pkg/k8sutils/k8sutils.go:505: cannot use nil as type "github.com/portworx/torpedo/vendor/k8s.io/client-go/pkg/apis/apps/v1beta1".StatefulSetStatus in return argument

Likewise for podList.

drivers/scheduler/k8s/k8s.go
varunturlapati: I was thinking of embedding the root error here but ended up neither here nor there. Added a new error type. ErrFailedToGetAppStatus.

drivers/scheduler/k8s/k8s.go
varunturlapati: Done

cmd/torpedo/torpedo.go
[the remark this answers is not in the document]
harsh-px: Primarily syntactical consistency.

cmd/torpedo/torpedo.go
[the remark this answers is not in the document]
harsh-px: torpedo-support-dir sounds good

drivers/scheduler/scheduler.go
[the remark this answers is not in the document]
harsh-px: The above function is not generating a log file. It is the caller of the function who uses the output to create the log file. The above function is serving the purpose of getting all details of the given context.

Yes, Describe() is a k8s operation but can be extended generally too since it aptly represents the purpose of the function.

pkg/k8sutils/k8sutils.go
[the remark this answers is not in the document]
harsh-px: Below compiles on my system

```
// GetPods gets pods
func GetPods(namespace string) (*v1.PodList, error) {
	return nil, nil
}
```

varunturlapati:
@harsh-px @piyush-nimbalkar - Updated the PR

cmd/torpedo/torpedo.go
```diff
@@ -335,15 +338,39 @@ func (t *torpedo) validateContext(ctx *scheduler.Context) error {
...
+			logrus.Warnf("In attempting to generate support bundle, Err: %v", gsbErr)
+		}
 		return err
 	}
 	return err
 }
 
+// generateSupportBundle gathers logs and any artifacts pertinent to the scheduler and dumps them in the defined location
+func (t *torpedo) generateSupportBundle(ctx *scheduler.Context) error {
+	if t == nil || ctx == nil || t.s == nil {
```
piyush-nimbalkar: no need to check t == nil now

cmd/torpedo/torpedo.go
```diff
@@ -335,15 +338,39 @@ func (t *torpedo) validateContext(ctx *scheduler.Context) error {
...
+		return fmt.Errorf("Invalid context or torpedo object. Cannot generate support bundle")
```
piyush-nimbalkar: *Invalid context or empty scheduler

cmd/torpedo/torpedo.go
```diff
@@ -645,6 +672,11 @@ func main() {
 					Usage: "Comma-separated list of tests. [Default:  runs all the tests]",
 					Value: "",
 				},
+				cli.StringFlag{
+					Name:  logLocationCliFlag,
+					Usage: "Path to save logs/artifacts upon failure. Default: /mnt/torpedo-support-dir",
```
piyush-nimbalkar: nit: Usually paths in linux do not have hyphen, you should have underscores - torpedo_support_dir

deployments/torpedo-k8s-ssh.yaml
```diff
@@ -51,7 +51,14 @@ spec:
   - name: torpedo
     image: harshpx/torpedo:latest
     #command: ["./torpedo", "fire", "--scheduler", "k8s", "--storage", "pxd", "--node-driver", "ssh" ]
-    command: ["./torpedo", "fire", "--scheduler", "k8s", "--storage", "pxd", "--node-driver", "ssh", "--tests", "testSetupTearDown" ]
+    command: ["./torpedo", "fire", "--scheduler", "k8s", "--storage", "pxd", "--node-driver", "ssh", "--tests", "testSetupTearDown", "--log-location", "/mnt/tsb" ]
```
piyush-nimbalkar: should we keep this file consistent with the default log location we have above?

drivers/scheduler/k8s/errors.go
```diff
@@ -79,6 +79,18 @@ func (e *ErrFailedToValidateApp) Error() string {
 	return fmt.Sprintf("Failed to validate app: %v due to err: %v", e.App.Key, e.Cause)
 }
 
+// ErrFailedToGetAppStatus error type for failing to validate an app
+type ErrFailedToGetAppStatus struct {
+	// App is the app that failed to destroy
```
piyush-nimbalkar: we need to change the comment here

drivers/scheduler/k8s/errors.go
```diff
@@ -138,3 +150,27 @@ type ErrFailedToGetVolumesParameters struct {
 func (e *ErrFailedToGetVolumesParameters) Error() string {
 	return fmt.Sprintf("Failed to get volume parameters for app: %v due to err: %v", e.App.Key, e.Cause)
 }
+
+// ErrFailedToGetPvcStatus error type for failing to get the status of a pvc
+type ErrFailedToGetPvcStatus struct {
+	// App is the pvc for which we couldn't get the status
```
piyush-nimbalkar: nit: *App whose PVC status could not be retrieved?

drivers/scheduler/k8s/errors.go
```diff
@@ -138,3 +150,27 @@ type ErrFailedToGetVolumesParameters struct {
...
+	Cause string
+}
+
+func (e *ErrFailedToGetPvcStatus) Error() string {
+	return fmt.Sprintf("Failed to get status of pvc: %v due to err: %v", e.App.Key, e.Cause)
+}
+
+// ErrFailedToGetScParams error type for failing to get the parameters of a storage class
+type ErrFailedToGetScParams struct {
+	// App is the storage class for which we couldn't get the params
```
piyush-nimbalkar: same as above..  we need to reword the comment

pkg/k8sutils/k8sutils.go
```diff
@@ -168,6 +168,23 @@ func GetService(svcName string, svcNS string) (*v1.Service, error) {
 
 }
 
+// DescribeService updates the service status and returns the status part
+func DescribeService(svc *v1.Service) (*v1.ServiceStatus, error) {
+	client, err := GetK8sClient()
+	if err != nil {
+		return nil, err
+	}
+	svc, err = GetService(svc.Name, svc.Namespace)
```
piyush-nimbalkar: by default the get does not return the status resource?

pkg/k8sutils/k8sutils.go
```diff
@@ -504,14 +555,21 @@ func DeletePods(pods []v1.Pod) error {
...
+func GetPods(namespace string) (*v1.PodList, error) {
+	var pods *v1.PodList
 	client, err := GetK8sClient()
 	if err != nil {
-		return nil, err
+		return pods, err
 	}
 
-	pods, err := client.Pods(namespace).List(meta_v1.ListOptions{})
+	pods, err = client.Pods(namespace).List(meta_v1.ListOptions{})
```
piyush-nimbalkar: nit: we can just say "return client.Pods(..."

pkg/k8sutils/k8sutils.go
```diff
@@ -550,6 +608,22 @@ func DeleteStorageClass(name string) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	sc, err = client.StorageV1().StorageClasses().Get(sc.Name, meta_v1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return sc.Parameters, nil
+
```
piyush-nimbalkar: nit: extra line

pkg/k8sutils/k8sutils.go
```diff
@@ -594,6 +668,22 @@ func DeletePersistentVolumeClaim(pvc *v1.PersistentVolumeClaim) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	result, err := client.PersistentVolumeClaims(pvc.Namespace).Get(pvc.Name, meta_v1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return &result.Status, nil
+
```
piyush-nimbalkar: nit: extraline

pkg/k8sutils/k8sutils.go
```diff
@@ -594,6 +668,22 @@ func DeletePersistentVolumeClaim(pvc *v1.PersistentVolumeClaim) error {
 	return client.PersistentVolumeClaims(pvc.Namespace).Delete(pvc.Name, &meta_v1.DeleteOptions{})
 }
 
+// GetPersistentVolumeClaimStatus returns the status of the given pvc
+func GetPersistentVolumeClaimStatus(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaimStatus, error) {
+	var err error
```
piyush-nimbalkar: we are not using this err variable.. let's remove it
we could remove it from GetStorageClassParams as well, if we use := on line 619

pkg/k8sutils/k8sutils.go
varunturlapati: For all of these my initial idea was to get the latest status and so the update status. Maybe I can remove the explicit update from this and other objects too.

drivers/scheduler/k8s/errors.go
piyush-nimbalkar: I think these comments were missed

pkg/k8sutils/k8sutils.go
```diff
@@ -168,6 +168,15 @@ func GetService(svcName string, svcNS string) (*v1.Service, error) {
 
 }
 
+// DescribeService updates the service status and returns the status part
```
piyush-nimbalkar: lets change the comment now.. as we are not updating it
same for describe deployment and statefulset

varunturlapati: I'll do that separately. Eventually we may have to put in an update query before returning the status too. That the Get is getting the latest status works for us for now.

Issue #26, opened 2017-09-20
Add support for multiple sections in the same k8s yaml spec file


harsh-px:
When parsing the yaml spec files, if we encounter a file like below, it doesn't work since it has a yaml file separator. We need to handle that.

```yaml
##### Portworx storage class
kind: StorageClass
apiVersion: storage.k8s.io/v1beta1
metadata:
    name: px-nginx-shared-sc
provisioner: kubernetes.io/portworx-volume
parameters:
   repl: "2"
   shared: "true"
---
##### Portworx persistent volume claim
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
   name: px-nginx-shared-pvc
   annotations:
     volume.beta.kubernetes.io/storage-class: px-nginx-shared-sc
spec:
   accessModes:
     - ReadWriteOnce
   resources:
     requests:
       storage: 1Gi
```

PR #42, opened 2017-09-28
Support multiple sections in kubernetes yaml spec file.

piyush-nimbalkar:
https://github.com/portworx/torpedo/issues/26


drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata.yaml
```diff
@@ -1,3 +1,11 @@
+kind: StorageClass
```
harsh-px: Keep `StorageClass` is it's separate file since that belongs to storage. While keeping in one file works, it's good to keep storage and app in separate files.

Do the the same for other specs.

adityadani: @harsh-px  But if we have them separate, does torpedo guarantee that PVC is created before the application?

drivers/scheduler/k8s/k8s.go
```diff
@@ -108,27 +111,34 @@ func (k *k8s) parseSpecs(specDir string, parser func(in interface{}) (interface{
...
-		}
+			if len(bytes.TrimSpace(specContents)) > 0 {
+				obj, _, err := scheme.Codecs.UniversalDeserializer().Decode([]byte(specContents), nil, nil)
+				if err != nil {
+					return nil, err
+				}
 
-		coreSpecs = append(coreSpecs, specObj)
+				specObj, err := parser(obj)
+				if specObj != nil {
```
adityadani: Catch the parser error here and return if not nil.

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata.yaml
harsh-px: Yes, Schedule implementation in k8s does that. It performs storage operations before app/core operations.

drivers/scheduler/k8s/k8s.go
piyush-nimbalkar: The parser does not throw any error right now. Probably we can get rid of the error from the parse method.

drivers/scheduler/k8s/k8s.go
```diff
@@ -24,7 +24,11 @@ import (
 )
 
 // SchedName is the name of the kubernetes scheduler driver implementation
-const SchedName = "k8s"
+const (
+	SchedName = "k8s"
```
adityadani: nit: SchedName should be a separate constant.

Define a different enum for core and storage

drivers/scheduler/k8s/k8s.go
```diff
@@ -130,40 +120,39 @@ func (k *k8s) parseSpecs(specDir string, parser func(in interface{}) (interface{
...
 			}
 		}
 	}
 
-	return coreSpecs, nil
+	return coreSpecs, storageSpecs, nil
 }
 
-func coreParser(in interface{}) (interface{}, error) {
+func checkSpecType(in interface{}) (interface{}, string, error) {
```
adityadani: Instead of string return the enum type here

PR #43, opened 2017-10-03
Adding test to verify that there are no resource leaks post destroy.

piyush-nimbalkar:
Currently we are checking that the k8s deletes the volume dir for pods.

### DIFF
--- a/cmd/torpedo/torpedo.go
+++ b/cmd/torpedo/torpedo.go
@@ -59,9 +59,11 @@ func (t *torpedo) testSetupTearDown() error {
 		}
 	}
 
+	opts := make(map[string]bool)
+	opts[scheduler.OptionsWaitForResourceLeakCleanup] = true
 	for _, ctx := range contexts {
 		logrus.Infof("[Test: %v] Tearing down %v", taskName, ctx.App.Key)
-		if err := t.tearDownContext(ctx); err != nil {
+		if err := t.tearDownContext(ctx, opts); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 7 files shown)


cmd/torpedo/torpedo.go
```diff
@@ -325,6 +325,39 @@ func (t *torpedo) testNodeReboot(allNodes bool) error {
 	return err
 }
 
+// testResourceCleanupOnTearDown checks if all the resources used by the application
+// have been cleaned up
+func (t *torpedo) testResourceCleanupOnTearDown() error {
```
harsh-px: Let's not add a new test. Resource cleanup would normally be default check for all tests and we might eventually do that if we have resources. So having a dedicated test for checking resource cleanup wouldn't be right. Rather, have one or more of the existing tests use the new functionality by invoking destroy with the new options.

drivers/scheduler/k8s/k8s.go
```diff
@@ -361,9 +364,16 @@ func (k *k8s) WaitForRunning(ctx *scheduler.Context) error {
...
 
-func (k *k8s) Destroy(ctx *scheduler.Context) error {
+func (k *k8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
+	var podList []v1.Pod
 	for _, spec := range ctx.App.SpecList {
 		if obj, ok := spec.(*apps_api.Deployment); ok {
+			if pods, err := k8sutils.GetDeploymentPods(obj); err != nil {
+				logrus.Warnf("[%v] Error getting deployment pods. Err: %v", ctx.App.Key, err)
+			} else {
+				podList = append(podList, pods...)
```
harsh-px: Looks like we need the `podList` only if `opts[scheduler.OptionsWaitForResourceLeakCleanup]` is set. So we avoid the fetching of pods for other use cases.

PR #44, opened 2017-10-04
Add access key to torpedo aws yaml file

jessicapwx:


deployments/torpedo-k8s-aws.yaml
```diff
@@ -50,6 +50,22 @@ spec:
   containers:
   - name: torpedo
     image: harshpx/torpedo:latest
+    env:
+      - name: AWS_ACCESS_KEY_ID
+        valueFrom:
+          secretKeyRef:
+            name: aws-region-accesskey
```
harsh-px: Use a secret with name `aws-access`.

jessicapwx: fixed

PR #46, opened 2017-10-04
Refactoring k8sutils to a singleton implementing sub interfaces

piyush-nimbalkar:
…-interfaces.


drivers/scheduler/k8s/k8s.go
```diff
@@ -245,9 +245,10 @@ func (k *k8s) Schedule(instanceID string, options scheduler.ScheduleOptions) ([]
 }
 
 func (k *k8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.AppSpec) (interface{}, error) {
+	k8sOps := k8sops.Instance()
```
adityadani: Rename the variable to something else? A bit confusing whether using the package or the instance.
Or you could just use k8sops.Instance() everywhere.

PR #48, opened 2017-10-09
Switch logging to ginkgo writer and misc changes

harsh-px:
1) Move logging to ginkgo writer
2) Add default CheckIfPathExists for aws node driver
3) Switch to failFast runs


util/util.go
```diff
@@ -0,0 +1,21 @@
...
+func nowStamp() string {
+	return time.Now().Format(time.StampMilli)
+}
+
+func log(level string, format string, args ...interface{}) {
+	fmt.Fprintf(ginkgo.GinkgoWriter, nowStamp()+": "+level+": "+format+"\n", args...)
+}
+
+// Logf logs info log to ginkgo writer
+func Logf(format string, args ...interface{}) {
```
piyush-nimbalkar: Instead of Logf, we should have individual methods like Infof, Warnf, Errorf

drivers/node/aws/aws.go
```diff
@@ -180,6 +180,11 @@ func (a *aws) ShutdownNode(n node.Node, options node.ShutdownNodeOpts) error {
 	return nil
 }
 
+// TODO add AWS implementation for this
+func (a *aws) CheckIfPathExists(path string, n node.Node, options node.ConnectionOpts) (bool, error) {
```
piyush-nimbalkar: I might change this in the current review. It should not matter a lot other that the signature change

pkg/k8sops/k8sops.go
```diff
@@ -640,18 +641,23 @@ func (k *k8sOps) ValidateTerminatedStatefulSet(statefulset *apps_api.StatefulSet
 			if matched, _ := regexp.MatchString(".+ not found", err.Error()); matched {
 				return "", nil
 			}
+
+			logrus.Errorf("[debug] error while getting stateful set: %v", err)
```
piyush-nimbalkar: why not use util.logf here?

scripts/do-gofmt.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+find_files() {
```
piyush-nimbalkar: let's remove the check-gofmt.sh if we are going to use this one

util/util.go
harsh-px: Updated in new incremental

scripts/do-gofmt.sh
harsh-px: Done

pkg/k8sops/k8sops.go
harsh-px: Added util.Errorf in new incremental

Makefile
```diff
@@ -19,7 +19,7 @@ GOFMT := gofmt
 
 .DEFAULT_GOAL=all
 
-all: vet lint build fmt
+all: clean vet lint build fmt
```
piyush-nimbalkar: you may not want to do a clean every time you build, right?
doing make clean && make all could be a better way around this one. What do you think?

util/util.go
piyush-nimbalkar: I was thinking of renaming Logf to Infof like the other methods or other logging libraries

Makefile
harsh-px: I was going for a default clean to happen every time. But `make clean && make all` works too. Removed it.

util/util.go
harsh-px: Yeah `Infof` makes more sense. Updated.

PR #49, opened 2017-10-09
Verify no data is written to vol dir after tests are run.

piyush-nimbalkar:


drivers/scheduler/k8s/k8s.go
```diff
@@ -458,17 +459,21 @@ func (k *k8s) waitForCleanup(ctx *scheduler.Context, podList []v1.Pod) error {
...
+		MaxDepth: 1,
+	}
 
 	for _, n := range k.GetNodes() {
 		if n.Type == node.TypeWorker {
-			if exists, err := driver.CheckIfPathExists(podVolDir, n, options); err != nil {
+			if volDir, err := driver.FindFiles(podVolDir, n, options); err != nil {
 				return err
-			} else if exists {
+			} else if strings.TrimSpace(volDir) != "" {
```
harsh-px: need to include `volDir` in the error message so it propagates up explicitly in the logs.

drivers/volume/portworx/schedops/errors.go
```diff
@@ -0,0 +1,28 @@
+package schedops
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ErrFailedToCleanupVolume error type for orphan pods or unclean vol directories
+type ErrFailedToCleanupVolume struct {
```
harsh-px: :+1:

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -30,6 +36,96 @@ func (k *k8sSchedOps) EnableOnNode(n node.Node) error {
...
+			}
+			if found {
+				continue
+			}
+			orphanPods = append(orphanPods, podUID)
+
+			// check if there are files under portworx volume
+			n := nodeMap[nodeName]
+			findFileOpts := node.FindOpts{
+				ConnectionOpts: connOpts,
```
harsh-px: Can you add an comment showing example why depth of 2 is selected here with the path regex *portworx-volume?

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -30,6 +36,96 @@ func (k *k8sSchedOps) EnableOnNode(n node.Node) error {
...
+
+func separateFilePaths(volDirList string) []string {
+	trimmedList := strings.TrimSpace(volDirList)
+	if trimmedList == "" {
+		return []string{}
+	}
+	return strings.Split(trimmedList, "\n")
+}
+
+func extractPodUID(volDirPath string) string {
```
harsh-px: Better to use a regex than just depending on the 5th item being pod uid.

drivers/scheduler/k8s/k8s.go
piyush-nimbalkar: the volDir is /var/lib/kubelet/pods/<pod_uid>/volumes/*
, where the only imp thing is the pod UID

harsh-px: Ok. I thought it has the volume name too. If not, no change is needed.

drivers/volume/portworx/schedops/k8s-schedops.go
piyush-nimbalkar: Done

drivers/volume/portworx/schedops/k8s-schedops.go
piyush-nimbalkar: Done

drivers/node/node.go
```diff
@@ -66,8 +74,8 @@ type Driver interface {
 	// ShutdownNode shuts down the given node
 	ShutdownNode(node Node, options ShutdownNodeOpts) error
 
-	// CheckIfPathExists checks whether the given path is present in the node
-	CheckIfPathExists(path string, node Node, options ConnectionOpts) (bool, error)
+	// FindFiles finds and returns the files for the given path regex and the node
+	FindFiles(path string, node Node, options FindOpts) (string, error)
```
harsh-px: Add a default implementation for aws node driver. Without that, the aws tests will start failing with FindFiles() method not supported error.

piyush-nimbalkar: oops missed it.. adding it.

PR #50, opened 2017-10-11
Check k8s labels on node for portworx volumes

piyush-nimbalkar:


drivers/volume/portworx/portworx.go
```diff
@@ -193,6 +193,14 @@ func (d *portworx) InspectVolume(name string, params map[string]string) error {
 		}
 	}
 
+	// Labels
+	if err := d.schedOps.ValidateLabels(vol); err != nil {
```
harsh-px: Let's have a generic API `ValidateVolume` and as part of the implementation we will validate the labels.

adityadani: We should actually have two APIs
- ValidateCreateVolume - checks if labels are applied on the node
- ValidateDeleteVolume - checks if labels are removed from the node.

drivers/volume/portworx/schedops/errors.go
```diff
@@ -26,3 +26,13 @@ func (e *ErrFailedToCleanupVolume) Error() string {
 	}
 	return strings.Join(cause, ", ")
 }
+
+// ErrLabelsMissingOnNode error type for missing volume labels on node
```
adityadani: Add a new error
ErrLabelNotRemovedOnNode

drivers/volume/portworx/portworx.go
piyush-nimbalkar: @harsh-px why not rename the InspectVolume to ValidateCreateVolume as aditya suggested, as it is doing the same. And keep the ValidateAddLabels (ValidateLabels) as part of it?
@adityadani I will add a ValidateDeleteVolume which will currently check the label removal, but as you suggested we can also verify that the volume is not present in px volume list call, etc.. probably will do that as part of a later check-in

harsh-px: Yes, add `InspectVolume` as a method in the schedops interface.

For volume delete validation, it can be merged with the existing `ValidateVolumeCleanup` method in the schedops interface. You can rename `ValidateVolumeCleanup` to `ValidateDeleteVolume`.

piyush-nimbalkar: The issue here is ValidateVolumeCleanup is called after suite and does not check volume specific checks. We will have to add a separate ValildateDeleteVolume in portworx.go which checks if a given volume is deleted/cleanup or not.

harsh-px: Sure. Go ahead with `ValildateDeleteVolume`.

drivers/scheduler/scheduler.go
```diff
@@ -77,7 +78,7 @@ type Driver interface {
 	InspectVolumes(*Context) error
 
 	// DeleteVolumes will delete a storage volume.
```
harsh-px: nit: update comment to say "DeleteVolumes will all storage volumes for the given context"

drivers/volume/portworx/portworx.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/portworx/torpedo/drivers/node"
 	"github.com/portworx/torpedo/drivers/scheduler"
 	torpedovolume "github.com/portworx/torpedo/drivers/volume"
+	vol "github.com/portworx/torpedo/drivers/volume"
```
harsh-px: The same pkg is imported on the line above

pkg/task/task.go
```diff
@@ -35,7 +35,7 @@ func DoRetryWithTimeout(t func() (interface{}, error), timeout, timeBeforeRetry
 					return
 				}
 
-				util.Infof("%v. Retry count: %v Next retry in: %v", err, count, timeBeforeRetry)
+				util.Warnf("%v. Retry count: %v Next retry in: %v", err, count, timeBeforeRetry)
```
harsh-px: Keep it info. Since this routine is used to operations like reboot, driver bounce where it is expected to do retries. Adding a warning will create a false alarm when someone is debugging a different issue.

drivers/volume/portworx/portworx.go
```diff
@@ -276,6 +285,30 @@ func (d *portworx) InspectVolume(name string, params map[string]string) error {
 	return nil
 }
 
+func (d *portworx) ValidateDeleteVolume(vol *vol.Volume) error {
+	name := d.schedOps.GetVolumeName(vol)
+	t := func() (interface{}, error) {
+		vols, err := d.volDriver.Inspect([]string{name})
```
adityadani: I think volDriver.Inspect will return openstorage's ErrEnoEnt error if it does not find a volume. You can add a specific check if ErrEnoEnt then return success

PR #51, opened 2017-10-11
Handle validation of multiple replica pods using non-shared volumes

harsh-px:
This would basically test the cybraics/vilas test case where 3 postgres pods will contend for a non-shared volume.

### DIFF
--- a/pkg/k8sops/k8sops.go
+++ b/pkg/k8sops/k8sops.go
@@ -930,7 +930,7 @@ func (k *k8sOps) GetPersistentVolumeClaimParams(pvc *v1.PersistentVolumeClaim) (
 	return params, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/k8sops/k8sops.go
```diff
@@ -892,6 +930,18 @@ func (k *k8sOps) GetPersistentVolumeClaimParams(pvc *v1.PersistentVolumeClaim) (
 	return params, nil
 }
 
+// IsPVCShared returns true if the PersistentVolumeClaim has been configured for use by multiple clients
```
piyush-nimbalkar: nit: *isPVCShared

harsh-px: Fixed

PR #53, opened 2017-10-13
Always use the correct endpoint address which is up for portworx driver

adityadani:
I usually test without Portworx service, which fails with a stale endpoint being used by torpedo. This will solve it.

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -33,12 +33,13 @@ const (
 )
 
 type portworx struct {
-	hostConfig     *dockerclient.HostConfig
-	clusterManager cluster.Cluster
-	volDriver      volume.VolumeDriver
-	schedDriver    scheduler.Driver
-	schedOps       schedops.Driver
-	nodeDriver     node.Driver
+	hostConfig      *dockerclient.HostConfig
+	clusterManager  cluster.Cluster
+	volDriver       volume.VolumeDriver
+	schedDriver     scheduler.Driver
+	schedOps        schedops.Driver
+	nodeDriver      node.Driver
+	refreshEndpoint bool
 }
 
 func (d *portworx) String() string {
@@ -64,7 +65,7 @@ func (d *portworx) Init(sched string, nodeDriver string) error {
 		return err
 	}
 
-	cluster, err := d.clusterManager.Enumerate()
+	cluster, err := d.getClusterManager().Enumerate()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -380,17 +382,24 @@ func (d *portworx) setDriver() error {
 	if err == nil {
 		endpoint = svc.Spec.ClusterIP
 		if err = d.testAndSetEndpoint(endpoint); err == nil {
+			d.refreshEndpoint = false
 			return nil
 		}
+		logrus.Debugf("testAndSetEndpoint failed for %v: %v", endpoint, err)
```
harsh-px: Use util.Logf instead of logrus.

Using ginkgo writer allows to use their non-verbose mode. The ways it works is if all tests pass, none of the ginkgo writer output gets displayed. But if a test fails, all of the output gets dumped.

drivers/volume/portworx/portworx.go
```diff
@@ -380,17 +382,24 @@ func (d *portworx) setDriver() error {
...
+	// and working driver if the endpoint we are hooked onto goes
+	// down
+	d.refreshEndpoint = true
 	for _, n := range nodes {
 		if n.Type == node.TypeWorker {
 			for _, addr := range n.Addresses {
 				if err = d.testAndSetEndpoint(addr); err == nil {
 					return nil
 				}
+				logrus.Debugf("testAndSetEndpoint failed for %v: %v", endpoint, err)
```
harsh-px: same as above

drivers/volume/portworx/portworx.go
```diff
@@ -417,8 +426,8 @@ func (d *portworx) testAndSetEndpoint(endpoint string) error {
 	}
 
 	d.volDriver = volumeclient.VolumeDriver(dClient)
-	d.clusterManager = clusterclient.ClusterManager(cClient)
-	util.Infof("Using %v as endpoint for portworx volume driver", pxEndpoint)
+	d.clusterManager = clusterManager
+	logrus.Infof("Using %v as endpoint for portworx volume driver", pxEndpoint)
```
harsh-px: same as above for logrus

drivers/volume/portworx/portworx.go
adityadani: ah! my changes are too old! Will fix this.

drivers/volume/portworx/portworx.go
adityadani: Fixed in incremental

drivers/volume/portworx/portworx.go
adityadani: Fixed in incremental

Issue #55, opened 2017-10-24
Add scale test


harsh-px:
Add a scale test that can exercise the current application specs but scaled by the given scale factor

PR #56, opened 2017-10-25
Add  scale test and fix volume driver down, app down test

harsh-px:


tests/basic/basic_test.go
```diff
@@ -124,14 +124,18 @@ var _ = Describe("Volume driver and app down", func() {
 			}
 		})
 
-		Step("destroy apps", func() {
+		Step("wait for destroy of apps and volumes", func() {
 			opts := make(map[string]bool)
 			opts[scheduler.OptionsWaitForDestroy] = true
 			for _, ctx := range contexts {
-				TearDownContext(ctx, opts)
+				Step(fmt.Sprintf("wait for destroy of app: %s", ctx.App.Key), func() {
```
piyush-nimbalkar: we don't need the opts for 'wait for destroy' on above line

tests/scale/scale_test.go
```diff
@@ -0,0 +1,45 @@
...
+var _ = BeforeSuite(func() {
+	InitInstance()
+})
+
+// This test performs scale test of starting an application and destroying it (along with storage)
+var _ = Describe("Scaled Setup and teardown", func() {
+	It("has to setup, validate and teardown large scale apps", func() {
+		var contexts []*scheduler.Context
+		for i := 0; i < Inst().ScaleFactor; i++ {
+			contexts = ScheduleAndValidate(fmt.Sprintf("scalesetupteardown-%d", i))
```
piyush-nimbalkar: it should be append

tests/scale/scale_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/portworx/torpedo/drivers/scheduler"
+	. "github.com/portworx/torpedo/tests"
+)
+
+func TestBasic(t *testing.T) {
```
piyush-nimbalkar: we have the same name 'TestBasic' in the basic tests
doesn't go conflict when two methods with same name are defined in the same package?

piyush-nimbalkar:
In scale test, are we verifying that more number of small apps work properly or a few large scale apps?

tests/scale/scale_test.go
harsh-px: You mean replace `fmt.Sprintf("scalesetupteardown-%d", i)` with string append? What's wrong with sprintf?

harsh-px:
> In scale test, are we verifying that more number of small apps work properly or a few large scale apps?
More number of small apps. For e.g, this gives a easy way to exercise 100 volumes or 100 pods where the window for scheduler races increases.

To test few large scale apps, one would basically invoke torpedo with custom specs which are large scale (For e.g replica 100 wordpress deployment)

tests/scale/scale_test.go
harsh-px: Fixed in incremental

harsh-px:
closes #55

PR #58, opened 2017-10-31
Remove k8s ops and vendor in portworx sched ops repo

harsh-px:
1) Remove k8s ops and vendor in portworx sched ops repo
2) Remove task package
3) Replace logrus package with smaller-case package


pkg/log/log.go
```diff
@@ -3,8 +3,8 @@ package log
 import (
 	"strings"
 
-	"github.com/Sirupsen/logrus"
 	"github.com/fatih/color"
+	"github.com/sirupsen/logrus"
```
piyush-nimbalkar: does it work with case change from upper case S to lower?

piyush-nimbalkar:
LGTM

pkg/log/log.go
harsh-px: Yes, it needs to be lower case now.

Reference: https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276

PR #59, opened 2017-10-31
add checks for px and k8s to be ready in init for the respective drivers

harsh-px:


drivers/scheduler/k8s/k8s.go
```diff
@@ -78,6 +78,9 @@ func (k *k8s) Init(specDir string, nodeDriverName string) error {
 
 	for _, n := range nodes.Items {
 		k.nodes[n.Name] = k.parseK8SNode(n)
+		if err := k.IsNodeReady(k.nodes[n.Name]); err != nil {
```
piyush-nimbalkar: should we wait if the node is not ready?

harsh-px: `IsNodeReady(...)` by default has a 5 minute timeout loop for retries. If after this time, node is not ready, we fail the init.

piyush-nimbalkar: oops.. missed that

piyush-nimbalkar:
LGTM

PR #61, opened 2017-11-03
Always use root user for running torpedo.

adityadani:
Let me know if I have to change any scripts for current torpedo runs in order for this change to take effect.

### DIFF
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -18,8 +18,6 @@ const (
 	DriverName = "ssh"
 	// DefaultUsername is the default username used for ssh operations
 	DefaultUsername = "root"
-	// DefaultPassword is the default password used for ssh operations
-	DefaultPassword = ""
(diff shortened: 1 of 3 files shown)


drivers/node/ssh/ssh.go
```diff
@@ -17,9 +17,9 @@ const (
 	// DriverName is the name of the ssh driver
 	DriverName = "ssh"
 	// DefaultUsername is the default username used for ssh operations
-	DefaultUsername = "torpedo"
-	// DefaultPassword is the default username used for ssh operations
-	DefaultPassword = "t0rped0"
+	DefaultUsername = "root"
+	// DefaultPassword is the default password used for ssh operations
+	DefaultPassword = ""
```
piyush-nimbalkar: why even have the default password field if we are not setting it up

adityadani: The other option was to set it down in this file as an empty string or add it as a constant up here. Just future proofing.

PR #62, opened 2017-11-03
Storing nodes in a registry instead of the schduler driver

piyush-nimbalkar:
- extracting out the service endpoint check in schedops from portworx driver


drivers/node/node_registry.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+var nodeRegistry = make(map[string]Node)
+
+// AddNode adds a node to the node collection
+func AddNode(n Node) error {
+	if n.uuid != "" {
+		return fmt.Errorf("UUID should not be set to add new node")
+	}
+	uuid, err := exec.Command("uuidgen").Output()
```
adityadani: instead of exec, you could use a golang library to generate UUID.
github.com/pborman/uuid

drivers/node/node_registry.go
```diff
@@ -0,0 +1,50 @@
+package node
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+var nodeRegistry = make(map[string]Node)
```
adityadani: Protect this map with a lock for concurrent access.

drivers/node/node.go
```diff
@@ -19,10 +19,12 @@ const (
 
 // Node encapsulates a node in the cluster
 type Node struct {
-	Name       string
-	Addresses  []string
-	UsableAddr string
-	Type       Type
+	uuid            string
+	VolDriverNodeID string
```
adityadani: VolDriverNodeID not used anywhere?

PR #63, opened 2017-11-07
DC/OS tests with torpedo

piyush-nimbalkar:
** Hardcoded nodes and marathon leader, will change in next review **

- setup - tear down test works with DC/OS
- creates supports application using px volumes
- creates application and waits for it to be ready
- deletes and validates deletion of application
- sample mysql app deployed using marathon
- no volume inspection
- torpedo runs as a metronome job


disrani-px:
Can you separate out the vendor changes into a separate commit.

piyush-nimbalkar:
@disrani-px do you mean just a separate commit or a separate review as well?

disrani-px:
Just a separate commit with the same PR

piyush-nimbalkar:
Done

drivers/volume/portworx/portworx.go
```diff
@@ -76,7 +80,25 @@ func (d *portworx) Init(sched string, nodeDriver string) error {
...
+func (d *portworx) updateNodes(pxNodes []api.Node) {
+	for _, n := range node.GetNodes() {
+		d.updateNode(n, pxNodes)
+	}
+}
+
+func (d *portworx) updateNode(n node.Node, pxNodes []api.Node) {
+	for _, address := range n.Addresses {
+		for _, pxNode := range pxNodes {
+			if address == pxNode.DataIp {
```
harsh-px: Add check for MgmtIP too.

drivers/scheduler/dcos/specs/mysql/px-mysql.json
```diff
@@ -0,0 +1,50 @@
...
+            ],
+            "privileged": false,
+            "parameters": [
+                {
+                    "key": "volume-driver",
+                    "value": "pxd"
+                },
+                {
+                    "key": "volume",
+                    "value": "size=20,repl=2,name=px_mysql_vol:/var/lib/mysql"
```
harsh-px: Let's use repl 3. Will give write quorum if we kill one of the nodes.

drivers/scheduler/dcos/marathon_ops.go
```diff
@@ -0,0 +1,110 @@
...
+	return nil
+}
+
+func (m *marathonOps) ValidateTerminatedApplication(name string) error {
+	t := func() (interface{}, error) {
+		if err := m.initMarathonClient(); err != nil {
+			return nil, err
+		}
+
+		_, err := m.client.Application(name)
```
harsh-px: Can be 

```
if _, err := m.client.Application(name); err != nil 
```

drivers/scheduler/dcos/marathon_ops.go
```diff
@@ -0,0 +1,110 @@
...
+	once.Do(func() {
+		instance = &marathonOps{}
+	})
+	return instance
+}
+
+// Initialize Marathon client if not initialized
+func (m *marathonOps) initMarathonClient() error {
+	if m.client == nil {
+		marathonURL := "http://192.168.65.90:8080"
```
harsh-px: How are we eventually going to find the marathon URL? (For e.g k8s gets it by the mounted px-account or the KUBE_CONFIG env variable).

drivers/scheduler/dcos/dcos.go
```diff
@@ -1 +1,234 @@
...
+	}
+
+	return nil
+}
+
+func getNodes() ([]node.Node, error) {
+	var nodes []node.Node
+	nodes = append(nodes, node.Node{
+		Name:      "a1.dcos",
+		Addresses: []string{"192.168.65.111"},
```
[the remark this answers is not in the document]
adityadani: If DCOS does not have such an API, can we pass it through ginkgo params or env variables?

drivers/scheduler/dcos/marathon_ops.go
```diff
@@ -0,0 +1,110 @@
...
+
+	return m.client.CreateApplication(app)
+}
+
+func (m *marathonOps) ValidateApplication(name string) error {
+	if err := m.initMarathonClient(); err != nil {
+		return err
+	}
+
+	return m.client.WaitOnApplication(name, 5*time.Minute)
```
adityadani: Is this a blocking call?
The callers are not expecting ValidateApplication to be a blocking call. Also the time period can be a constant.

drivers/scheduler/dcos/marathon_ops.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			if matched, _ := regexp.MatchString(".+ does not exist", err.Error()); !matched {
+				return nil, err
+			}
+		}
+
+		return nil, nil
+	}
+
+	if _, err := task.DoRetryWithTimeout(t, 5*time.Minute, 10*time.Second); err != nil {
```
adityadani: May be we could regulate the timeouts and wait period constants by defining them in a constant file. All these constants should be common for both k8s and dcos

drivers/volume/portworx/portworx.go
```diff
@@ -330,19 +352,29 @@ func (d *portworx) StopDriver(n node.Node) error {
...
 
-func (d *portworx) WaitStart(n node.Node) error {
-	var err error
-	// Wait for Portworx to become usable.
+// Wait for Portworx to be up
+func (d *portworx) WaitStart() error {
 	t := func() (interface{}, error) {
-		clusterManager := d.getClusterManager()
-		if status, _ := clusterManager.NodeStatus(); status != api.Status_STATUS_OK {
+		if status, _ := d.getClusterManager().NodeStatus(); status != api.Status_STATUS_OK {
```
adityadani: I don't understand the use of this function if we have WaitForNode. This function is only going to check the status of the node which is currently the endpoint.

If the intention of this function is to just check Cluster status then use
clusterManager.Enumerate() which returns api.Cluster object which has cluster status.

deployments/torpedo-dcos-ssh.json
```diff
@@ -0,0 +1,9 @@
+{
+    "id": "torpedo",
+    "description": "Run torpedo on DC/OS",
+    "run": {
+        "cmd": "docker run -e 'TORPEDO_SSH_PASSWORD=root' --entrypoint ginkgo piyushpx/torpedo:latest --slowSpecThreshold 180 -v bin/basic.test -- --spec-dir ../drivers/scheduler/dcos/specs --scheduler dcos",
```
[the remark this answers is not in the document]
piyush-nimbalkar: This is run as a job in dcos. The use will have to create it manually using the file and change the appropriate env variables.
For deployment from jenkins, we will have a shell script like we have for deploy-aws later which can take multiple params.

drivers/scheduler/dcos/dcos.go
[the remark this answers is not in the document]
piyush-nimbalkar: There is a way to get nodes from mesos. But no way to get the marathon or dcos master. So here the user will have to give us the master ip using env variable which we can use to get nodes or communicate to marathon.

drivers/scheduler/dcos/dcos.go
```diff
@@ -1 +1,234 @@
...
+
+type dcos struct {
+	specFactory *spec.Factory
+}
+
+func (d *dcos) Init(specDir string, nodeDriver string) error {
+	nodes, _ := getNodes()
+
+	for _, n := range nodes {
+		node.AddNode(n)
```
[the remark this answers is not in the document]
piyush-nimbalkar: done in incremental

drivers/scheduler/dcos/dcos.go
```diff
@@ -1 +1,234 @@
...
+			return nil, err
+		}
+
+		specs = append(specs, app)
+	}
+
+	return specs, nil
+}
+
+func (d *dcos) IsNodeReady(n node.Node) error {
```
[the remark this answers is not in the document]
piyush-nimbalkar: done in incremental

drivers/scheduler/dcos/marathon_ops.go
piyush-nimbalkar: done

drivers/scheduler/dcos/specs/mysql/px-mysql.json
piyush-nimbalkar: done

drivers/volume/portworx/portworx.go
piyush-nimbalkar: done

drivers/scheduler/dcos/marathon_ops.go
piyush-nimbalkar: Renamed the methods

drivers/scheduler/dcos/marathon_ops.go
piyush-nimbalkar: I agree that we should have constants for all these things. There are many such places, lets change it everywhere all at once.
Created an issue to track:
https://github.com/portworx/torpedo/issues/67

drivers/scheduler/dcos/dcos.go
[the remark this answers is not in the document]
piyush-nimbalkar: Actually wrong info. Mesos provides a nice set of dns names. We can use "leader.mesos" to get to the mesos host. Similarly for marathon.

drivers/volume/portworx/portworx.go
piyush-nimbalkar: done in incremental

drivers/scheduler/dcos/dcos.go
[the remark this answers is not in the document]
disrani-px: Marathon is running on the master node, so it'll be the same as master.mesos.

You can use mesos APIs to get the list of nodes. For example using curl: curl http://master.mesos:5050/slaves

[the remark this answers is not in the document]
piyush-nimbalkar: Yes, the mesos-go library seemed a bit complicated, hence decided to do that in a later commit.

drivers/scheduler/dcos/specs/mysql/px-mysql.json
```diff
@@ -0,0 +1,50 @@
...
+            ],
+            "privileged": false,
+            "parameters": [
+                {
+                    "key": "volume-driver",
+                    "value": "pxd"
+                },
+                {
+                    "key": "volume",
+                    "value": "size=10,repl=3,name=px_mysql_vol:/var/lib/mysql"
```
harsh-px: reduce size to 2. 10G will cause issues on dev systems which might have less storage.

tests/common.go
```diff
@@ -77,14 +78,14 @@ func ValidateContext(ctx *scheduler.Context) {
...
-		Step(fmt.Sprintf("validate %s app's volumes", ctx.App.Key), func() {
-			ValidateVolumes(ctx)
-		})
-
 		Step(fmt.Sprintf("wait for %s app to start running", ctx.App.Key), func() {
 			err := Inst().S.WaitForRunning(ctx)
 			expect(err).NotTo(haveOccurred())
 		})
+
+		Step(fmt.Sprintf("validate %s app's volumes", ctx.App.Key), func() {
```
harsh-px: Any reason to move volume check after app check? The volume check was before applications since volumes are a pre-requisite for a app container to start. So that itself fails, there is no need to `WaitForRunning`.

drivers/volume/portworx/portworx.go
```diff
@@ -330,19 +348,33 @@ func (d *portworx) StopDriver(n node.Node) error {
...
 			}
 		}
 
-		pxNode, err := clusterManager.Inspect(n.Name)
+		return cluster, nil
+	}
+
+	cluster, err := task.DoRetryWithTimeout(t, 2*time.Minute, 10*time.Second)
+	if err != nil {
+		return api.Cluster{}, err
```
harsh-px: In an error case, it's better to return a nil object than a default one.

deployments/torpedo-dcos-ssh.json
```diff
@@ -0,0 +1,15 @@
+{
+    "id": "torpedo",
+    "description": "Run torpedo on DC/OS",
+    "run": {
+        "cpus": 0.5,
+        "mem": 256,
+        "docker": {
+            "image": "piyushpx/torpedo:latest"
```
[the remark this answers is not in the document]
piyush-nimbalkar: done

drivers/scheduler/dcos/dcos.go
```diff
@@ -1 +1,247 @@
...
+				}
+			}
+			logrus.Infof("[%v] Destoyed application: %v", ctx.App.Key, obj.ID)
+		}
+	}
+
+	if value, ok := opts[scheduler.OptionsWaitForResourceLeakCleanup]; ok && value {
+		if err := d.WaitForDestroy(ctx); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
piyush-nimbalkar: done

drivers/scheduler/dcos/specs/mysql/px-mysql.json
piyush-nimbalkar: done

drivers/volume/portworx/portworx.go
piyush-nimbalkar: it was an object, hence could not return a nil
but changed it to a reference

tests/common.go
piyush-nimbalkar: done as per discussed.

piyush-nimbalkar:
Thanks for the review guys!

drivers/scheduler/dcos/specs/mysql/px-mysql.json
```diff
@@ -0,0 +1,50 @@
...
+                    "servicePort": 10002,
+                    "protocol": "tcp",
+                    "name": "default"
+                }
+            ],
+            "privileged": false,
+            "parameters": [
+                {
+                    "key": "volume-driver",
+                    "value": "pxd"
```
disrani-px: These specs should be templated out. The driver field should be populated depending on which volume driver is being tested. It looks like it is the same for the k8s specs. So they can be fixed together.

drivers/scheduler/dcos/marathon_ops.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// ApplicationOps is an interface to perform application operations
+type ApplicationOps interface {
+	// CreateApplication creates an application in Marathon
+	CreateApplication(*marathon.Application) (*marathon.Application, error)
+	// ValidateAppliation checks if the aplication is running and healthy
+	ValidateApplication(string) error
+	// DeleteApplication deletes the given application
+	DeleteApplication(string) error
```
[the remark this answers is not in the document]
piyush-nimbalkar: Follow up on the discussion - 
There are counter intuitive apis to do the stop and kill operations.
Stop is basically updating the app with instance count of 0
While kill operation calls a DELETE on the tasks of that app.

We can implement these as we need going forward.

drivers/scheduler/dcos/specs/mysql/px-mysql.json
piyush-nimbalkar: Created an issue to track this: https://github.com/portworx/torpedo/issues/70

PR #68, opened 2017-11-08
Get private agent nodes from mesos

piyush-nimbalkar:
- use domain name provided by mesos as marathon default url

### DIFF
--- a/drivers/scheduler/dcos/mesos_ops.go
+++ b/drivers/scheduler/dcos/mesos_ops.go
@@ -99,6 +99,8 @@ func (m *mesosOps) GetPrivateAgentNodes() ([]AgentNode, error) {
 
 	var nodes []AgentNode
 	for _, n := range agentsResp.Agents {
+		// Filtering out public agent nodes. Public agent nodes will have the
+		// public_ip attribute set to true to differentiate from other agents
 		if n.Attributes.PublicIP == "" || n.Attributes.PublicIP != "true" {
 			nodes = append(nodes, n)
 		}
--- a/drivers/scheduler/dcos/mesos_ops.go
+++ b/drivers/scheduler/dcos/mesos_ops.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	defaultMesosHostname = "master.mesos"
+	defaultMesosHostname = "leader.mesos"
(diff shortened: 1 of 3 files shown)


drivers/scheduler/dcos/mesos_ops.go
```diff
@@ -0,0 +1,108 @@
+package dcos
```
adityadani: Why don't we start with moving this package under the new portworx/sched-ops repo now itself? Instead of doing it later.

drivers/scheduler/dcos/mesos_ops.go
```diff
@@ -0,0 +1,108 @@
...
+
+	var agentsResp GetAgentsResponse
+	err = json.Unmarshal(body, &agentsResp)
+	if err != nil {
+		return nil, err
+	}
+
+	var nodes []AgentNode
+	for _, n := range agentsResp.Agents {
+		if n.Attributes.PublicIP == "" || n.Attributes.PublicIP != "true" {
```
adityadani: Can you add a doc line as to why this check is needed?
Does this mean that if the node has publicIP do not add that as an agent node?

drivers/scheduler/dcos/mesos_ops.go
adityadani: Also the marathon sched ops

piyush-nimbalkar: I wanted to get basic dcos tests running before I move all these ops in the other repo, as I am actively modifying the ops. It will be a hassle to keep maintaining the sched-ops up to date with every change I make here. Will migrate is soon after I have the tests passing.

drivers/scheduler/dcos/mesos_ops.go
piyush-nimbalkar: Yes, dcos has master nodes, private agent nodes and public agents nodes.. master is for most admin work, cluster management etc.
Most of the apps and services are run on private agents nodes (in our case portworx runs only on private agents. Public agents can be used to run services which need external facing address - something like load balancer, repoxy.

To avoid confusion I have named the method to GetPrivateAgentNodes. But I will add a comment anyway.

PR #69, opened 2017-11-08
Drive failure tests

adityadani:
Add utilities in node driver and volume driver to Fail and Recover drives.
- Add two new APIs in node driver -YankDrive and RecoverDrive
- Add implementations for ssh driver.
- Add two new APIs in volume driver - GetStorageDevices and RecoverDriver.

Add a ginkgo test to induce a drive failure and check apps.

Drive Failure implementation for AWS coming soon.

### DIFF
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -204,7 +204,7 @@ func (s *ssh) YankDrive(n node.Node, driveNameToFail string, options node.Connec
 	}
 
 	driveID = strings.TrimRight(driveID, "\n")
-	driveNameToFail = strings.TrimRight(strings.TrimLeft(driveNameToFail, "/"), "/")
+	driveNameToFail = strings.Trim(driveNameToFail, "/")
 	devices := strings.Split(driveNameToFail, "/")
 
 	// Disable the block device, so that it returns IO errors
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/portworx/sched-ops/task"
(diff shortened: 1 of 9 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -150,6 +154,78 @@ func (d *portworx) CleanupVolume(name string) error {
...
+
+	storageInfo, ok := pxNode.NodeData[storageInfoKey]
+	if !ok {
+		return []string{}, fmt.Errorf("Unable to find storage info for node: %v", n.Name)
+	}
+	storageInfoMap := storageInfo.(map[string]interface{})
+
+	resourcesMapIntf, ok := storageInfoMap[resourcesKey]
+	if !ok || resourcesMapIntf == nil {
+		return []string{}, fmt.Errorf("Unable to find resource info for node: %v", n.Name)
```
piyush-nimbalkar: same here.. can we send nil?

drivers/volume/portworx/portworx.go
```diff
@@ -150,6 +154,78 @@ func (d *portworx) CleanupVolume(name string) error {
...
+	}
+	resourcesMap := resourcesMapIntf.(map[string]interface{})
+
+	for _, v := range resourcesMap {
```
piyush-nimbalkar: nit: can we move the devPaths slice above this for loop from line 167.. it seems out of place there

tests/drive_failure/drive_failure_test.go
```diff
@@ -0,0 +1,105 @@
...
+			for _, ctx := range contexts {
+				ValidateAndDestroy(ctx, nil)
+			}
+		})
+
+	})
+}
+
+var _ = Describe("Induce drive failure on of the nodes", func() {
+	driveFailureTest("drivefailure")
```
piyush-nimbalkar: nit: instead of a separate method we could just put the 'it' block in this 'describe'
In reboot its done so because two Describes are using the same method with different params

drivers/node/ssh/ssh.go
```diff
@@ -181,6 +182,65 @@ func (s *ssh) ShutdownNode(n node.Node, options node.ShutdownNodeOpts) error {
...
+	addr, err := s.getAddrToConnect(n, options)
+	if err != nil {
+		return &node.ErrFailedToRecoverDrive{
+			Node:  n,
+			Cause: fmt.Sprintf("failed to get node address due to: %v", err),
+		}
+	}
+
+	// Enable the drive by rescaning
+	recoverCmd := "echo \" - - -\" > /sys/class/scsi_host/host" + driveUUIDToRecover + "/scan"
```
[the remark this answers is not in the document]
adityadani: No. 
It is no a subdirectory under host

drivers/volume/portworx/portworx.go
```diff
@@ -150,6 +154,78 @@ func (d *portworx) CleanupVolume(name string) error {
 	return nil
 }
 
+func (d *portworx) GetStorageDevices(n node.Node) ([]string, error) {
+	const (
+		storageInfoKey = "STORAGE-INFO"
+		resourcesKey   = "Resources"
+		pathKey        = "path"
+	)
+	pxNode, err := d.getClusterManager().Inspect(n.Name)
```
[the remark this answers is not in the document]
adityadani: good catch. I didn't run the tests after master rebase

drivers/volume/portworx/portworx.go
```diff
@@ -150,6 +154,78 @@ func (d *portworx) CleanupVolume(name string) error {
...
+	if err != nil {
+		return []string{}, err
```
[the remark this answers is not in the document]
adityadani: done

drivers/volume/portworx/portworx.go
adityadani: done

drivers/volume/portworx/portworx.go
adityadani: moved devPaths down in the function

tests/drive_failure/drive_failure_test.go
adityadani: Removed the wrapper function.

tests/drive_failure/drive_failure_test.go
```diff
@@ -0,0 +1,102 @@
...
+func TestDriveFailure(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Torpedo: DriveFailure")
+}
+
+var _ = BeforeSuite(func() {
+	InitInstance()
+})
+
+var _ = Describe("Induce drive failure on of the nodes", func() {
```
harsh-px: typo in description

tests/drive_failure/drive_failure_test.go
```diff
@@ -0,0 +1,102 @@
...
+					appNodes             []node.Node
+					nodeWithDrive        node.Node
+					driveToFail, driveID string
+				)
+
+				Step(fmt.Sprintf("get nodes where %s app is running", ctx.App.Key), func() {
+					appNodes, err = Inst().S.GetNodesForApp(ctx)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(appNodes).NotTo(BeEmpty())
+					nodeWithDrive = appNodes[0]
```
harsh-px: Instead of picking the first one, we can do this on a quorum on nodes.

tests/drive_failure/drive_failure_test.go
```diff
@@ -0,0 +1,102 @@
...
+				})
+
+				Step(fmt.Sprintf("get drive from node %v", nodeWithDrive), func() {
+					drives, err = Inst().V.GetStorageDevices(nodeWithDrive)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(drives).NotTo(BeEmpty())
+					driveToFail = drives[0]
```
harsh-px: Why just the first drive and not all?

tests/drive_failure/drive_failure_test.go
adityadani: I don't see a reason why we need to fail drives on more than one node. 
What I can check for is if the node whose drive I am failing has the replica of the volume being used.

tests/drive_failure/drive_failure_test.go
adityadani: Will add code to fail all the drives on one node.

tests/drive_failure/drive_failure_test.go
harsh-px: Yeah. We could end up off-lining drives on a node which doesn't have the replica of the volume.

adityadani: Currently it assumes that the volume replica is 3 and the cluster size is 3 (default settings).
I have created an issue to add a new API to handle the generic case where replica may not exist on all nodes - #83

PR #71, opened 2017-11-13
Inspect and delete volumes for DCOS torpedo test

piyush-nimbalkar:
- Updating vendor using 'glide up -v' which removed nested vendor directories and creates flat dependencies in vendor directory. Going forward everyone should use glide up -v to avoid creating nested dependencies.

- Using k8s.io/api for all k8s objects as it is going to be the official repo for all k8s objects.


drivers/scheduler/dcos/dcos.go
```diff
@@ -208,24 +220,124 @@ func (d *dcos) DeleteTasks(ctx *scheduler.Context) error {
...
-	// TODO: Implement this method
+	// DC/OS does not have volume objects like Kubernetes. We get the volume information from
+	// the app spec and get the options parsed from the respective volume driver
+	volDriver, err := volume.Get(d.volDriverName)
+	if err != nil {
+		return err
+	}
+
+	for _, spec := range ctx.App.SpecList {
+		if obj, ok := spec.(*marathon.Application); ok {
```
disrani-px: Duplicate code as above? Can this be moved into getVolumes() that can be called from here and above method?

drivers/scheduler/dcos/dcos.go
```diff
@@ -208,24 +220,124 @@ func (d *dcos) DeleteTasks(ctx *scheduler.Context) error {
...
-	return nil, nil
+	volDriver, err := volume.Get(d.volDriverName)
+	if err != nil {
+		return nil, err
+	}
+
+	var vols []*volume.Volume
+
+	for _, spec := range ctx.App.SpecList {
+		if obj, ok := spec.(*marathon.Application); ok {
```
disrani-px: Same comment as above, move to a common function.

drivers/scheduler/dcos/errors.go
```diff
@@ -53,3 +53,39 @@ type ErrFailedToValidateAppDestroy struct {
 func (e *ErrFailedToValidateAppDestroy) Error() string {
 	return fmt.Sprintf("Failed to validate destroy of app: %v due to err: %v", e.App.Key, e.Cause)
 }
+
+// ErrFailedToValidateVolume error type for failing to validate a volume
```
disrani-px: Why are these errors separate for the schedulers? There seems to be nothing scheduler specific and both the schedulers seem to have the same errors re-defined

drivers/scheduler/dcos/dcos.go
```diff
@@ -208,24 +220,124 @@ func (d *dcos) DeleteTasks(ctx *scheduler.Context) error {
...
+	volDriver, err := volume.Get(d.volDriverName)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]map[string]string)
+
+	for _, spec := range ctx.App.SpecList {
+		if obj, ok := spec.(*marathon.Application); ok {
+			params := *obj.Container.Docker.Parameters
```
disrani-px: This will need to be updated when testing with UCR/mesos containers: https://docs.portworx.com/scheduler/mesosphere-dcos/portworx_volumes.html#mesosucr-containers

piyush-nimbalkar: Yes we need to handle mesos containers. Creating a issue to track it.. will soon work on it.

drivers/scheduler/dcos/dcos.go
piyush-nimbalkar: done

drivers/scheduler/dcos/errors.go
piyush-nimbalkar: yes there is no need to have them separate.
There were some errors which were PVC or Storage Class specific. I made it generic and extracted in scheduler/errors

PR #76, opened 2017-11-21
Adding upgrade test

piyush-nimbalkar:
- upgrades to a version and then downgrades to a lower version
- fixing the add label retry logic; retry on label missing instead of retry on get labels

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -3,6 +3,7 @@ package portworx
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/libopenstorage/openstorage/api"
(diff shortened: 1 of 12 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -399,6 +400,40 @@ func (d *portworx) WaitForNode(n node.Node) error {
...
+
+		if pxNode.Status != api.Status_STATUS_OK {
+			return nil, &ErrFailedToWaitForPx{
+				Node: n,
+				Cause: fmt.Sprintf("px cluster is usable but node status is not ok. Expected: %v Actual: %v",
+					api.Status_STATUS_OK, pxNode.Status),
+			}
+		}
+
+		pxVersion := pxNode.NodeLabels["PX Version"]
```
adityadani: nit: use a constant.

drivers/volume/portworx/schedops/dcos-schedops.go
```diff
@@ -51,6 +51,10 @@ func (d *dcosSchedOps) GetServiceEndpoint() string {
 	return ""
 }
 
+func (d *dcosSchedOps) UpgradePortworx(version string) error {
+	return nil
```
adityadani: nit: comment for TODO ?

tests/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,63 @@
...
+var _ = BeforeSuite(func() {
+	InitInstance()
+})
+
+var _ = Describe("Upgrade volume driver", func() {
+	It("upgrade volume driver and ensure everything is running fine", func() {
+		contexts := ScheduleAndValidate("upgradevolumedriver")
+
+		Step("start the upgrade of volume driver", func() {
+			err := Inst().V.UpgradeDriver("1.2.11.4")
```
adityadani: Can the version be configurable? May be use env variables OR pass in to the upgrade_test binary as args ?

tests/common.go
```diff
@@ -28,21 +28,25 @@ import (
...
-	defaultNodeDriver     = "ssh"
-	defaultStorageDriver  = "pxd"
-	defaultLogLocation    = "/mnt/torpedo_support_dir"
-	defaultAppScaleFactor = 10
+	defaultScheduler                   = "k8s"
+	defaultNodeDriver                  = "ssh"
+	defaultStorageDriver               = "pxd"
+	defaultLogLocation                 = "/mnt/torpedo_support_dir"
+	defaultAppScaleFactor              = 10
+	defaultStorageDriverUpgradeVersion = "1.2.11.6"
```
adityadani: The defaults should be defined in the storage driver. This version is Portworx specific. The tests can pass in an empty version string as the default one and the storage driver can then react on it and use the default ones.

piyush-nimbalkar: The storage driver implementation should not worry about the default upgrade version required for the test. 
Here we are using 'pxd' as the default storage driver, so shouldn't it be okay to specify default to be px specific version? 
It is similar to when the user gives 'k8s' as the scheduler, they have to give the proper spec dir for that scheduler. 
What do you think?

adityadani: Setting the default storage driver/scheduler/node-driver here makes sense because this piece of code instantiates them.
But setting some specific driver's constants here is a bit fuzzy. Let the storage driver decide what their default versions are. In this way even for a new storage driver the upgrade tests will work even if the input args are not specified.. I don't think we set the default k8s directory here?

PR #78, opened 2017-11-21
Some changes for errors

disrani-px:


drivers/volume/portworx/portworx.go
```diff
@@ -404,13 +404,15 @@ func (d *portworx) setDriver() error {
 	var endpoint string
 
 	// Try portworx-service first
-	endpoint = d.schedOps.GetServiceEndpoint()
-	if endpoint != "" {
+	endpoint, err = d.schedOps.GetServiceEndpoint()
+	if err == nil {
```
adityadani: nit : add a check for endpoint != "" along with err == nil

disrani-px: Done

PR #84, opened 2017-12-12
Add support to parse and create snapshots

disrani-px:


drivers/scheduler/k8s/k8s.go
```diff
@@ -137,6 +141,21 @@ func (k *k8s) ParseSpecs(specDir string) ([]interface{}, error) {
...
+		if err := snap_v1.AddToScheme(scheme); err != nil {
+			return nil, err
+		}
+		codecs := serializer.NewCodecFactory(scheme)
+		obj, _, err = codecs.UniversalDeserializer().Decode([]byte(specContents), nil, nil)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return obj, err
```
piyush-nimbalkar: nit: could return nil for readability

drivers/scheduler/k8s/k8s.go
```diff
@@ -282,6 +303,18 @@ func (k *k8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.
...
+		obj.Metadata.Namespace = ns.Name
+		snap, err := k8sOps.CreateSnapshot(obj)
+		if err != nil {
+			return nil, &scheduler.ErrFailedToScheduleApp{
+				App:   app,
+				Cause: fmt.Sprintf("Failed to create Snapshot: %v. Err: %v", snap.Metadata.Name, err),
+			}
+		}
+
+		logrus.Infof("[%v] Created Ssnapshot: %v", app.Key, snap.Metadata.Name)
```
piyush-nimbalkar: nit: *Snapshot

drivers/scheduler/k8s/k8s.go
```diff
@@ -618,6 +660,16 @@ func (k *k8s) DeleteVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 				}
 			}
 			logrus.Infof("[%v] Destroyed PVC: %v", ctx.App.Key, obj.Name)
+		} else if obj, ok := spec.(*snap_v1.VolumeSnapshot); ok {
+			if err := k8sOps.DeleteSnapshot(obj.Metadata.Name, obj.Metadata.Namespace); err != nil {
+				if matched, _ := regexp.MatchString(".+ not found", err.Error()); !matched {
```
piyush-nimbalkar: nit: can we just use strings.Contains() here as well as above?

glide.yaml
```diff
@@ -23,14 +23,18 @@ import:
...
 - package: k8s.io/api
   version: kubernetes-1.8.2
 - package: k8s.io/client-go
   version: kubernetes-1.8.2
 - package: github.com/portworx/sched-ops
-  version: master
-  vcs: git
+  version: master 
   repo: git@github.com:portworx/sched-ops.git
+  vcs: git
```
piyush-nimbalkar: you can remove this now

drivers/scheduler/k8s/k8s.go
disrani-px: Done

drivers/scheduler/k8s/k8s.go
disrani-px: Done

drivers/scheduler/k8s/k8s.go
disrani-px: Done

glide.yaml
disrani-px: Done

glide.yaml
```diff
@@ -23,14 +23,17 @@ import:
...
 - package: k8s.io/api
   version: kubernetes-1.8.2
 - package: k8s.io/client-go
   version: kubernetes-1.8.2
 - package: github.com/portworx/sched-ops
-  version: master
-  vcs: git
+  version: master 
   repo: git@github.com:portworx/sched-ops.git
+  subpackages:
```
piyush-nimbalkar: why do we not need to subpackage task?

disrani-px: This was autogenerated using "glide up".

piyush-nimbalkar: I thought glide up does not update the glide.yaml file. It populates it only for the first time.
Can you try removing subpackages (or add task) and running glide up again?

disrani-px: glide.lock had both k8s and tasks package so it was fine. I've removed the subpackages from here

PR #92, opened 2018-01-16
add routine to collect kernel threads and journal output at the end of each run

harsh-px:
This is useful to debug issues where certain runs fail because docker is hung and system cannot move forward.  Gathering the kernel threads at that point can give some pointers on what could be the root cause.

### DIFF
--- a/drivers/scheduler/scheduler.go
+++ b/drivers/scheduler/scheduler.go
@@ -44,6 +44,9 @@ type Driver interface {
 	// IsNodeReady checks if node is in ready state. Returns nil if ready.
 	IsNodeReady(n node.Node) error
 
+	// GetNodes returns all nodes known the scheduler
+	GetNodes() ([]node.Node, error)
+
 	// GetNodesForApp returns nodes on which given app context is running
 	GetNodesForApp(*Context) ([]node.Node, error)
 
--- a/drivers/scheduler/scheduler.go
+++ b/drivers/scheduler/scheduler.go
@@ -44,9 +44,6 @@ type Driver interface {
 	// IsNodeReady checks if node is in ready state. Returns nil if ready.
 	IsNodeReady(n node.Node) error
 
-	// GetNodes returns all nodes known the scheduler
-	GetNodes() ([]node.Node, error)
-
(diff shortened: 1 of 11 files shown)


tests/common.go
```diff
@@ -205,7 +202,7 @@ func ValidateAndDestroy(ctx *scheduler.Context, opts map[string]bool) {
 }
 
 // generateSupportBundle gathers logs and any artifacts pertinent to the scheduler and dumps them in the defined location
-func generateSupportBundle(ctx *scheduler.Context) {
+/*func generateSupportBundle(ctx *scheduler.Context) {
```
adityadani: nit: delete commented code if not required.

tests/common.go
```diff
@@ -219,6 +216,28 @@ func generateSupportBundle(ctx *scheduler.Context) {
...
+
+			journalCmd := fmt.Sprintf(
+				"echo t > /proc/sysrq-trigger && journalctl -l > ~/all_journal_%v",
+				time.Now().Format(time.RFC3339))
+			for _, n := range nodes {
+				_, err := Inst().N.RunCommand(n, journalCmd, node.ConnectionOpts{
+					Timeout:         2 * time.Minute,
+					TimeBeforeRetry: 10 * time.Second,
+				})
+				expect(err).NotTo(haveOccurred())
```
adityadani: if this expect fails for one node will it try journalctl on the remaining nodes?

drivers/scheduler/scheduler.go
```diff
@@ -44,6 +44,9 @@ type Driver interface {
 	// IsNodeReady checks if node is in ready state. Returns nil if ready.
 	IsNodeReady(n node.Node) error
 
+	// GetNodes returns all nodes known the scheduler
+	GetNodes() ([]node.Node, error)
```
piyush-nimbalkar: we don't need the GetNodes() method in the scheduler as the node package already has it.
Check drivers/node/node_registry.go
The scheduler adds all nodes in the node registry at the start and all the consumers are using GetNodes, GetWorkerNodes from this.

piyush-nimbalkar:
LGTM

PR #94, opened 2018-01-17
don't delete k8s namespace in destroy implementation

### LINKED ISSUE #93: Destroy implementation in k8s driver deletes namespaces causing pvc's to also get deletd
The current destroy impl for k8s driver deletes the namespace. This causes all the pvcs in that namespace to get deleted which was not the intention of the caller. Only the application pods and services should get deleted.

harsh-px:
fixes #93

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -447,16 +447,6 @@ func (k *k8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
 		}
 	}
 
-	appNamespace := getAppNamespaceName(ctx.App, ctx.UID)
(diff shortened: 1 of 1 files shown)


disrani-px:
Deleting the namespace should be moved a cleanup routine then, because otherwise tests will leave back lots of stale namespaces.

harsh-px:
@disrani-px we already have https://github.com/portworx/torpedo/blob/master/scripts/k8s/k8s-torpedo-cleanup-all.sh that people can use to cleanup all resources including namespaces.

disrani-px:
That script should only be required in cases where tests are failing. If tests are passing everything should be cleaned up by the test programatically.

PR #109, opened 2018-02-08
Upgrade Portworx using talisman

### LINKED ISSUE #108: Use talisman to upgrade portworx
Use https://github.com/portworx/talisman/tree/master/cmd/talisman to upgrade Portworx

harsh-px:
Closes #108

This will now use https://github.com/portworx/talisman/tree/master/cmd/talisman for upgrading Portworx


Makefile
```diff
@@ -16,6 +16,7 @@ BASE_DIR := $(shell git rev-parse --show-toplevel)
 
 BIN :=$(BASE_DIR)/bin
 GOFMT := gofmt
++GLIDEPATH := $(shell command -v glide 2> /dev/null)
```
piyush-nimbalkar: what does the + mean?

drivers/volume/portworx/portworx.go
```diff
@@ -559,11 +559,21 @@ func (d *portworx) WaitForUpgrade(n node.Node, newVersion string) error {
 			}
 		}
 
+		// filter out first 3 octets from the tag
+		matches := regexp.MustCompile(`^(\d+\.\d+\.\d+)-.*`).FindStringSubmatch(tag)
```
piyush-nimbalkar: this may not work for portworx releases like 1.2.11.4.. you may want an additional \.d* after the first group

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -235,41 +238,104 @@ func (k *k8sSchedOps) GetServiceEndpoint() (string, error) {
...
+	}
+
+	// cleanup
+	err = inst.DeleteJob(job.Name, job.Namespace)
+	if err != nil {
+		return err
 	}
 
-	if _, err := task.DoRetryWithTimeout(t, 20*time.Minute, 30*time.Second); err != nil {
+	err = inst.DeleteJob(job.Name, job.Namespace)
```
piyush-nimbalkar: this is duplicate, we have already deleted the job

piyush-nimbalkar:
Can we separate the commits for vendor update and the upgrade?
Also, does everything work fine with the glide -v flag now?

harsh-px:
@piyush-nimbalkar 
> Can we separate the commits for vendor update and the upgrade?
Done

> Also, does everything work fine with the glide -v flag now?
Yes, it works.

Makefile
harsh-px: I had picked this up directly from https://github.com/lpabon/quartermaster/blob/dev/Makefile#L13

Doesn't appear this is needed. Removed it.

drivers/volume/portworx/portworx.go
harsh-px: Removed the - in new changes which will handle 1.2.11.4. Note that we only match the first 3 octets.

drivers/volume/portworx/schedops/k8s-schedops.go
harsh-px: Oops. Deleted.

PR #120, opened 2018-02-21
Wait a little bit longer for portworx to enter maintenance mode.

piyush-nimbalkar:
This test failed intermittently because it took little more time to enter maintenance mode. Increasing the retry time to 2 minutes.

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -34,6 +34,22 @@ const (
 	pxSystemdServiceName    = "portworx.service"
 )
 
+const (
+	defaultRetryInterval              = 10 * time.Second
+	maintenanceOpTimeout              = 1 * time.Minute
+	maintenanceWaitTimeout            = 2 * time.Minute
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -215,7 +215,7 @@ func (d *portworx) RecoverDriver(n node.Node) error {
 		return nil, true, fmt.Errorf("Node %v is not in Maintenance mode", n.Name)
 	}
 
-	if _, err := task.DoRetryWithTimeout(t, 1*time.Minute, 10*time.Second); err != nil {
+	if _, err := task.DoRetryWithTimeout(t, 2*time.Minute, 10*time.Second); err != nil {
```
harsh-px: Since you are at it, can you move timeouts in this file to  a single `const` section at the top. Basically something like https://github.com/portworx/talisman/blob/master/pkg/cluster/px/px.go#L86

piyush-nimbalkar: 👍 
Done!

PR #125, opened 2018-03-22
Add wordpress, add crash tests and allow tests to scale

harsh-px:
1) Add wordpress spec
2) Add crash test
3) Allow all tests to scale up
4) Allow users to specify which apps to run for test
5) Fix after suite resource cleanup
6) Remove scale test as all other tests can scale now


deployments/deploy-ssh.sh
```diff
@@ -129,7 +129,10 @@ function debug_log_then_exit {
 for i in $(seq 1 600) ; do
   printf .
   state=`kubectl get pod torpedo | grep -v NAME | awk '{print $3}'`
-  if [ "$state" == "Running" ] || [ "$state" == "Completed" ]; then
+  if [ "$state" == "Error" ]; then
+    echo "Error: Torpedo finished with $endState state"
```
piyush-nimbalkar: $state instead of $endState

drivers/volume/portworx/portworx.go
```diff
@@ -438,13 +438,27 @@ func (d *portworx) ValidateVolumeCleanup() error {
...
-		ConnectionOpts: node.ConnectionOpts{
-			Timeout:         stopDriverTimeout,
-			TimeBeforeRetry: defaultRetryInterval,
-		}})
+func (d *portworx) StopDriver(n node.Node, force bool) error {
+	var err error
+	if force {
+		pxCrashCmd := "sudo kill -9 px-storage"
+		_, err = d.nodeDriver.RunCommand(n, pxCrashCmd, node.ConnectionOpts{
+			Timeout:         2 * time.Minute,
```
piyush-nimbalkar: nit: extract constants

tests/common.go
```diff
@@ -278,30 +318,34 @@ func ParseFlags() {
...
 	flag.StringVar(&volUpgradeVersion, storageDriverUpgradeVersionCliFlag, defaultStorageDriverUpgradeVersion,
 		"Version of storage driver to be upgraded to")
 	flag.StringVar(&volBaseVersion, storageDriverBaseVersionCliFlag, defaultStorageDriverBaseVersion,
 		"Version of storage driver to be downgraded to")
+	flag.StringVar(&appListCSV, appListCliFlag, "", "Comma-separated list of apps to run as part of test. The names should match directories in the spec dir.")
 
 	flag.Parse()
 
+	appList := make([]string, 0)
+	if len(appListCSV) > 0 {
```
piyush-nimbalkar: doesn't the csv reader handle empty string?

tests/common.go
```diff
@@ -278,30 +318,34 @@ func ParseFlags() {
...
+		appList, err = splitCsv(appListCSV)
+		if err != nil {
+			logrus.Fatalf("failed to parse app lists: %v. err: %v", appListCSV, err)
```
piyush-nimbalkar: nit: list

tests/common.go
```diff
@@ -215,6 +230,30 @@ func StopVolDriverAndWait(appNodes []node.Node) {
...
+func CrashVolDriverAndWait(appNodes []node.Node) {
+	context(fmt.Sprintf("crashing volume driver %s", Inst().V.String()), func() {
+		Step(fmt.Sprintf("crash volume driver on nodes: %v", appNodes), func() {
+			for _, n := range appNodes {
+				err := Inst().V.StopDriver(n, true)
+				expect(err).NotTo(haveOccurred())
+			}
+		})
+
+		Step(fmt.Sprintf("Sleeping for %v for crash to take effect", waitVolDriverToCrash), func() {
```
piyush-nimbalkar: isn't this a lot of time to wait unconditionally? I feel the crash and wait time should be inside the storage driver implementation and not in the test.
and it would vary based on the CO as well

harsh-px:
@piyush-nimbalkar fixed review comments.

drivers/volume/portworx/portworx.go
```diff
@@ -438,13 +440,29 @@ func (d *portworx) ValidateVolumeCleanup() error {
...
+	var err error
+	if force {
+		pxCrashCmd := "sudo kill -9 px-storage"
+		_, err = d.nodeDriver.RunCommand(n, pxCrashCmd, node.ConnectionOpts{
+			Timeout:         crashDriverTimeout,
 			TimeBeforeRetry: defaultRetryInterval,
-		}})
+		})
+		if err != nil {
+			logrus.Warnf("failed to run cmd: %s. err: %v", pxCrashCmd, err)
```
piyush-nimbalkar: We should return if there is an error. Anyway the test will fail when we return the err at the end. It will save us the waiting time.

PR #127, opened 2018-03-27
 Add retry around k8s create objects, update libopenstorage for new portworx io profile

harsh-px:


drivers/scheduler/k8s/k8s.go
```diff
@@ -305,9 +337,11 @@ func (k *k8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.
 		obj.Namespace = ns.Name
 		pvc, err := k8sOps.CreatePersistentVolumeClaim(obj)
 		if err != nil {
-			return nil, &scheduler.ErrFailedToScheduleApp{
-				App:   app,
-				Cause: fmt.Sprintf("Failed to create PVC: %v. Err: %v", pvc.Name, err),
+			if !strings.Contains(err.Error(), "already exists") {
```
piyush-nimbalkar: nit: could just && the condition with the 'if' above

harsh-px: done

PR #132, opened 2018-03-31
Stop all nodes together and then wait

Rohit-PX:


drivers/volume/portworx/portworx.go
```diff
@@ -444,31 +444,34 @@ func (d *portworx) ValidateVolumeSetup(vol *torpedovolume.Volume) error {
...
-			Action: "stop",
-			ConnectionOpts: node.ConnectionOpts{
-				Timeout:         stopDriverTimeout,
-				TimeBeforeRetry: defaultRetryInterval,
-			}})
+		return err
 	}
-
-	return err
+	logrus.Infof("Sleeping for %v for crash to take effect", waitVolDriverToCrash)
```
piyush-nimbalkar: keep this inside the first if case .. we want to wait only when we abnormally shutdown px / crash it

drivers/volume/portworx/portworx.go
```diff
@@ -444,31 +444,34 @@ func (d *portworx) ValidateVolumeSetup(vol *torpedovolume.Volume) error {
...
-		logrus.Infof("Sleeping for %v for crash to take effect", waitVolDriverToCrash)
-		time.Sleep(waitVolDriverToCrash)
-	} else {
-		err = d.nodeDriver.Systemctl(n, pxSystemdServiceName, node.SystemctlOpts{
-			Action: "stop",
-			ConnectionOpts: node.ConnectionOpts{
-				Timeout:         stopDriverTimeout,
-				TimeBeforeRetry: defaultRetryInterval,
-			}})
+		return err
```
piyush-nimbalkar: move this return inside the else condition.. otherwise it will exit the for loop on the first iteration
return from the 'else' only if there is err, similar to the 'if' part

drivers/volume/portworx/portworx.go
```diff
@@ -444,31 +444,40 @@ func (d *portworx) ValidateVolumeSetup(vol *torpedovolume.Volume) error {
...
+				Timeout:         crashDriverTimeout,
+				TimeBeforeRetry: defaultRetryInterval,
+			})
+			if err != nil {
+				logrus.Warnf("failed to run cmd : %s. on node %s err: %v", pxCrashCmd, n.Name, err)
+				return err
+			}
+
+		} else {
+			logrus.Infof("Using systmectl stop.")
```
piyush-nimbalkar: nit: remove this log

drivers/volume/portworx/portworx.go
```diff
@@ -444,31 +444,40 @@ func (d *portworx) ValidateVolumeSetup(vol *torpedovolume.Volume) error {
...
-		err = d.nodeDriver.Systemctl(n, pxSystemdServiceName, node.SystemctlOpts{
-			Action: "stop",
-			ConnectionOpts: node.ConnectionOpts{
-				Timeout:         stopDriverTimeout,
-				TimeBeforeRetry: defaultRetryInterval,
-			}})
 	}
-
-	return err
+	logrus.Infof("Sleeping for %v for crash to take effect", waitVolDriverToCrash)
```
piyush-nimbalkar: move this sleep inside the 'if'.. we want to wait only for abnormal shutdown

drivers/volume/portworx/portworx.go

PR #143, opened 2018-04-16
Vol repl increase decrease tests

varunturlapati:


drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
 }
 
+func (d *portworx) GetReplicationFactor(vol *torpedovolume.Volume) (int64, error) {
+	name := d.schedOps.GetVolumeName(vol)
+	t := func() (interface{}, bool, error) {
+		vols, err := d.volDriver.Inspect([]string{name})
+		if err != nil && err == volume.ErrEnoEnt {
+			return 0, false, volume.ErrEnoEnt
+		} else if err != nil {
+			return 0, false, err
```
piyush-nimbalkar: In any other failure case return "true". This means we need to keep retrying.

drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+		}
+		if len(vols) == 1 {
+			return vols[0].Spec.HaLevel, false, nil
+		}
+		return 0, false, fmt.Errorf("Extra volumes with the same volume name/ID seen") //Shouldn't reach this line
+	}
+
+	tmp, err := task.DoRetryWithTimeout(t, defaultRetryInterval, defaultRetryInterval)
```
piyush-nimbalkar: change the second arg to use some retry timeout, instead of interval
nit: can we not call this tmp? may be call iRepFactor to indicate it is an interface

drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+}
+
+func (d *portworx) SetReplicationFactor(vol *torpedovolume.Volume, replFactor int64) error {
+	name := d.schedOps.GetVolumeName(vol)
+	t := func() (interface{}, bool, error) {
+		vols, err := d.volDriver.Inspect([]string{name})
+		if err != nil && err == volume.ErrEnoEnt {
+			return nil, false, volume.ErrEnoEnt
+		} else if err != nil {
+			return nil, false, err
```
piyush-nimbalkar: same as above, for all places where you want to actually retry

drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+		}
+		spec := &api.VolumeSpec{
+			HaLevel:          int64(replFactor),
+			SnapshotInterval: math.MaxUint32,
+			ReplicaSet:       &api.ReplicaSet{},
+		}
+
+		logrus.Infof("Replication passed here is: %d", replFactor)
```
piyush-nimbalkar: was this for debugging? If not, let's put it outside the retry function.

drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+		if len(vols) == 1 {
+			locator := &api.VolumeLocator{
```
piyush-nimbalkar: why not use the locator as it is and give it to set function?

drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+func (d *portworx) SetReplicationFactor(vol *torpedovolume.Volume, replFactor int64) error {
+	name := d.schedOps.GetVolumeName(vol)
+	t := func() (interface{}, bool, error) {
+		vols, err := d.volDriver.Inspect([]string{name})
+		if err != nil && err == volume.ErrEnoEnt {
+			return nil, false, volume.ErrEnoEnt
+		} else if err != nil {
+			return nil, false, err
+		}
+		spec := &api.VolumeSpec{
```
piyush-nimbalkar: we should get the existing spec from vols[0] and then only change the HaLevel, instead reinitializing the object, right?

also move this under -  if len(vols)== 1

drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+		}
+
+		logrus.Infof("Replication passed here is: %d", replFactor)
+		if len(vols) == 1 {
+			locator := &api.VolumeLocator{
+				Name:         vols[0].Locator.Name,
+				VolumeLabels: vols[0].Locator.VolumeLabels,
+			}
+			err = d.volDriver.Set(vols[0].Id, locator, spec)
+			logrus.Infof("Err after Set is %v", err)
```
piyush-nimbalkar: nit: remove this

drivers/volume/portworx/portworx.go
```diff
@@ -625,6 +627,81 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+			if err != nil {
+				return nil, false, err
+			}
+			return nil, false, nil
+		}
+		return 0, false, fmt.Errorf("Extra volumes with the same volume name/ID seen") //Shouldn't reach this line
+	}
+
+	_, err := task.DoRetryWithTimeout(t, validateReplicationUpdateTimeout, defaultRetryInterval)
```
piyush-nimbalkar: nit: can combine in one statement:
```
if _, err := task.DoRetry(); err != nil {
}
```

drivers/volume/volume.go
```diff
@@ -70,6 +70,13 @@ type Driver interface {
 
 	// GetStorageDevices returns the list of storage devices used by the given node.
 	GetStorageDevices(n node.Node) ([]string, error)
+
+	//GetVolumeReplicationFactor returns the current replication factor of the volume.
+	//TODO - Can we define the return type to be of the type *api.VolumeSpec.HaLevel?
```
piyush-nimbalkar: We don't need to do this TODO as api.VolumeSpec.HaLevel is very OSD specific. Let's keep this as generic as possible. int64 seems fine for now, unless we decide to only support OSD.
nit: spaces after // .. for SetRF as well

tests/repl/repl_test.go
```diff
@@ -0,0 +1,178 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Torpedo : Repl")
+}
+
+var _ = BeforeSuite(func() {
+	InitInstance()
+})
+
+// This test performs basic test of starting an application and destroying it (along with storage)
+var _ = Describe("SetupTeardown", func() {
```
piyush-nimbalkar: remove this test, as it already present in the basic_test suite.

tests/repl/repl_test.go
```diff
@@ -0,0 +1,178 @@
...
+							newRepl, err := Inst().V.GetReplicationFactor(v)
+							Expect(err).NotTo(HaveOccurred())
+							Expect(newRepl).To(Equal(expReplMap[v]))
+							Inst().V.ValidateVolumeSetup(v)
+						}
+					})
+			}
+		})
+
+		Step("destroy apps", func() {
```
piyush-nimbalkar: you can use ValidateAndDestroy() from common.go
It iterates over all the contexts for you.

tests/repl/repl_test.go
```diff
@@ -0,0 +1,178 @@
...
+						Inst().V.String(), ctx.App.Key, appVolumes),
+					func() {
+						for _, v := range appVolumes {
+							errExpected := false
+							currRep, err := Inst().V.GetReplicationFactor(v)
+							Expect(err).NotTo(HaveOccurred())
+							if currRep == 1 {
+								errExpected = true
+							}
+							expReplMap[v] = int64(math.Min(3, float64(currRep)-1))
```
piyush-nimbalkar: can we simply do:
```
expReplMap[v] = currRep
if currRep > 1 {
    expReplMap[v] = int64(math.Min(3, float64(currRep)-1))
    err = Inst().V.SetReplicationFactor(v, currRep-1)
}
```

tests/repl/repl_test.go
```diff
@@ -0,0 +1,178 @@
...
+					})
+				Step(
+					fmt.Sprintf("validate successful repl decrease"),
+					func() {
+						time.Sleep(1 * time.Minute)
+						for _, v := range appVolumes {
+							newRepl, err := Inst().V.GetReplicationFactor(v)
+							Expect(err).NotTo(HaveOccurred())
+							Expect(newRepl).To(Equal(expReplMap[v]))
+							Inst().V.ValidateVolumeSetup(v)
```
piyush-nimbalkar: no need to ValidateVolumeSetup because it does not check anything related to our current operation
calling ValidateContext should check everything
the ValidateAndDestroy below will call ValidateContext

vendor/github.com/portworx/sched-ops/k8s/k8s.go
```diff
@@ -261,6 +261,8 @@ type StorageClassOps interface {
 	// TODO: This is currently the same as GetStorageClass. If no one is using it,
 	// we should remove this method
 	ValidateStorageClass(name string) (*storage_api.StorageClass, error)
+	// UpdateStorageClassParams updates the pvc by the given name using params passed in
```
piyush-nimbalkar: nit: updates the Storage class
where is this method being used in the review?

tests/repl/repl_test.go
```diff
@@ -0,0 +1,178 @@
...
+							if !errExpected {
+								Expect(err).NotTo(HaveOccurred())
+							}
+
+						}
+					})
+				Step(
+					fmt.Sprintf("validate successful repl decrease"),
+					func() {
+						time.Sleep(1 * time.Minute)
```
harsh-px: why do we need to sleep?

varunturlapati:
Removed unintended vendor update. Renamed test.

tests/repl/repl_test.go
varunturlapati: This is to avoid any transient values/behavior impacting the test. I will work on adding more robust checks but I can do that in subsequent PRs. Thoughts?

drivers/volume/portworx/portworx.go
```diff
@@ -648,6 +650,78 @@ func (d *portworx) WaitForUpgrade(n node.Node, image, tag string) error {
...
+				SnapshotInterval: math.MaxUint32,
+				ReplicaSet:       &api.ReplicaSet{},
+			}
+			locator := &api.VolumeLocator{
+				Name:         vols[0].Locator.Name,
+				VolumeLabels: vols[0].Locator.VolumeLabels,
+			}
+			err = d.volDriver.Set(vols[0].Id, locator, spec)
+			if err != nil {
+				return nil, false, err
```
piyush-nimbalkar: should retry on this failure

tests/repl/repl_test.go
harsh-px: If we hit any transient values/behavior, let's approach it when we hit them. We will evaluate the remediation at that point. Remove the sleep for now.

tests/repl/repl_test.go
piyush-nimbalkar: Then at least verify that there is error with errExpected == true

tests/volume_ops/volume_ops_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+				Step(
+					fmt.Sprintf("repl increase volume driver %s on app %s's volumes: %v",
+						Inst().V.String(), ctx.App.Key, appVolumes),
+					func() {
+						for _, v := range appVolumes {
+							errExpected := false
+							currRep, err := Inst().V.GetReplicationFactor(v)
+							Expect(err).NotTo(HaveOccurred())
+							if currRep == 3 {
```
piyush-nimbalkar: This check for expecting error for repl more than 3 is very px specific.. not all volumes drives have a max repl factor of 3

varunturlapati: Not sure that is true. I think 3 is more a standard than exception. Even otherwise, elsewhere we have fine tuned APIs based on PX behavior like the return types of HaLevel. How is this different?

drivers/volume/portworx/portworx.go
varunturlapati: Why should this be retried? This is the failure of setting repl factor. Should flag it is an error and be triaged. If anything, overtime we may come up with a list of retry-able exceptions. Not sure we are there yet.

tests/volume_ops/volume_ops_test.go
varunturlapati: If anything, we may create a map of drivertype --> constants like this one. For PX, the max repl = 3 and min repl = 1. For XYZDriver it may be whatever. We don't have XYZDriver yet but defining the map/constants may make it easy to add different types later.

piyush-nimbalkar: HaLevel/repl factor is still an integer.. anyone can convert to int64 as it is not a custom type.
For now, I think keep the constant as is instead of making it complex. We or someone else will do it when needed.

tests/volume_ops/volume_ops_test.go
```diff
@@ -0,0 +1,158 @@
...
+							if !errExpected {
+								Expect(err).NotTo(HaveOccurred())
+							}
+
+						}
+					})
+				Step(
+					fmt.Sprintf("validate successful repl decrease"),
+					func() {
+						time.Sleep(1 * time.Minute)
```
harsh-px: let's remove the sleep.

tests/volume_ops/volume_ops_test.go
harsh-px: Create a volume driver interface `GetMaxSupportedReplicationFactor()` and have that return 3 for PX.

tests/volume_ops/volume_ops_test.go
```diff
@@ -0,0 +1,158 @@
...
+								Expect(err).NotTo(HaveOccurred())
+							}
+
+						}
+					})
+				Step(
+					fmt.Sprintf("validate successful repl increase"),
+					func() {
+						time.Sleep(1 * time.Minute)
+						for _, v := range appVolumes {
```
harsh-px: remove the sleep

drivers/volume/portworx/portworx.go
harsh-px: No need to retry now. If we hit intermittent errors in future in `Set(...)` we will evaluate whether to retry or not at that point.

varunturlapati:
Ping! The test works with the polling mechanism to see if the volume is clean and the num of nodes in replica set equals the new ha level. Checks passed here too.
@harsh-px @piyush-nimbalkar

tests/volume_ops/volume_ops_test.go
```diff
@@ -0,0 +1,161 @@
...
+						}
+					})
+				Step(
+					fmt.Sprintf("validate successful repl decrease"),
+					func() {
+						for _, v := range appVolumes {
+							newRepl, err := Inst().V.GetReplicationFactor(v)
+							Expect(err).NotTo(HaveOccurred())
+							Expect(newRepl).To(Equal(expReplMap[v]))
+							Inst().V.ValidateVolumeSetup(v)
```
harsh-px: The err is not being checked here. Without that the call to `Inst().V.ValidateVolumeSetup(v)` is useless.

harsh-px:
We can merge after the conflicts are resolved.

PR #144, opened 2018-04-17
Improved wordpress implementation

sam-myers:
- Pods will enter running state only when everything is actually ready
- Generates continuous load
- Automated installation


drivers/scheduler/k8s/specs/wordpress/wordpress.yaml
```diff
@@ -1,28 +1,16 @@
...
   labels:
     app: wordpress
+    tier: frontend
 spec:
   selector:
     matchLabels:
       app: wordpress
-  replicas: 3
+      tier: frontend
+  replicas: 2
```
harsh-px: let's do repl 2 here

drivers/scheduler/k8s/specs/wordpress/wordpress.yaml
```diff
@@ -48,7 +46,46 @@ spec:
...
           mountPath: /var/www/html
+        livenessProbe:
+          httpGet:
+            path: /
+            port: 80
+          initialDelaySeconds: 15
+          timeoutSeconds: 5
+        resources:
+          requests:
+            memory: "2Gi"
```
harsh-px: is this a must? I would refrain from requesting memory since this will run in a test context and will contend with other pods for memory.

drivers/scheduler/k8s/specs/wordpress/wordpress.yaml
```diff
@@ -48,7 +46,46 @@ spec:
...
+  ports:
+    - port: 80
+      name: web
+  selector:
+    app: wordpress
+    tier: frontend
+  type: NodePort
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
```
harsh-px: we can skip the ingress for torpedo since we never use it for any tests. Do any of the client pods talk to ingress?

drivers/scheduler/k8s/specs/wordpress/installer.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: v1
+kind: Pod
```
harsh-px: Use a Deployment instead of a Pod.

drivers/scheduler/sidecars/wp-cli.dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /home/www-data
+
+USER root
+RUN apk add --update bash less curl
+USER www-data
+
+COPY scripts/wp-install.sh .
+
+CMD bash ./wp-install.sh
```
harsh-px: do we need bash here ? wp-install.sh has a bash shebang.

drivers/scheduler/k8s/specs/wordpress/wordpress.yaml
sam-myers: Could you clarify? This is repl 2

drivers/scheduler/k8s/specs/wordpress/wordpress.yaml
sam-myers: This is our customer's recommended size for woocommerce sites. With lower values we were regularly seeing seige push the memory use high enough for Kubernetes to kill the pods regularly.

drivers/scheduler/k8s/specs/wordpress/wordpress.yaml
harsh-px: Oops. Sorry meant let's do repl 3 here.

PR #147, opened 2018-04-18
Return only running pods in GetNodesForApp

Rohit-PX:


drivers/scheduler/k8s/k8s.go
```diff
@@ -899,8 +899,9 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
 		if node.Contains(result, n) {
 			continue
 		}
-
-		result = append(result, n)
+		if k8s_ops.Instance().IsPodReady(p) {
```
harsh-px: Use `IsPodRunning`

PR #150, opened 2018-04-18
Using brackets for test names to make regex matches easy

piyush-nimbalkar:


harsh-px:
Do we need to update the jenkins jobs right away?

piyush-nimbalkar:
no need to update the job, if you are already using comma as delimiter. I think with the previous review we are already doing it.

PR #160, opened 2018-05-02
enable stork for elasticsearch and mysql

### LINKED ISSUE #105: Add option to use a particular scheduler
If a scheduler is specified use it for all the app specs. Use the default kube-scheduler otherwise.

harsh-px:


disrani-px:
Let's not close #105 with this, I think it would still be good to have an option to update the scheduler name in all apps.

PR #161, opened 2018-05-03
Use correct volume name for snapshot

disrani-px:


harsh-px:
@disrani-px I don't think this is needed after #174 . Can be closed.

disrani-px:
Please update vendor in stork and verify that the test passes there. The test looks for volume name as the key and this sets the volume I'd, so there might be an issue

harsh-px:
> Please update vendor in stork and verify that the test passes there. The test looks for volume name as the key and this sets the volume I'd, so there might be an issue


Sure which test do we need to check? stork-master?

disrani-px:
Any of the jobs. They are all using the master build of stork with different versions of PX

harsh-px:
All tests on all branches are passing with torpedo on master. Closing this one.

PR #164, opened 2018-05-09
Add DeepCopy for AppSpec

disrani-px:
The kubernetes APIs are no longer in 1.10 which causes issues
in projects that vendor torpedo. The recommended way is to generate
deepcopy functions, but AppSpec is simple enough.

### DIFF
--- a/drivers/scheduler/spec/factory.go
+++ b/drivers/scheduler/spec/factory.go
@@ -5,9 +5,9 @@ import (
 	"io/ioutil"
 	"path"
 
+	"github.com/mohae/deepcopy"
 	"github.com/portworx/torpedo/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"k8s.io/client-go/kubernetes/scheme"
 )
 
 // Factory is an application spec factory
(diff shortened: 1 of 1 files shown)


drivers/scheduler/spec/spec.go
```diff
@@ -14,3 +14,18 @@ type AppSpec struct {
...
+	if in == nil {
+		return nil
+	}
+	out := new(AppSpec)
+	out.Key = in.Key
+	out.Enabled = in.Enabled
+	out.SpecList = make([]interface{}, len(in.SpecList))
+	for _, spec := range in.SpecList {
+		out.SpecList = append(out.SpecList, spec)
+	}
```
piyush-nimbalkar: the spec if a pointer will still point to the same object, right?
most of our specs are pointers

disrani-px: Changed it to use a library which should do it correctly

drivers/scheduler/spec/factory.go
```diff
@@ -49,12 +44,7 @@ func (f *Factory) GetAll() []*AppSpec {
 	var specs []*AppSpec
 	for _, val := range appSpecFactory {
 		if val.Enabled {
-			valCopy, err := scheme.Scheme.DeepCopy(val)
-			if err != nil {
-				logrus.Errorf("Failed to create a copy of spec object: %#v Err: %v", val, err)
-				return nil
-			}
-
+			valCopy := deepcopy.Copy(val)
```
piyush-nimbalkar: may be have a nil check before we type cast it to avoid panic?
and not append if it is nil.
Same for Get, return nil

disrani-px: Added a type assertion, please review the update

drivers/scheduler/spec/factory.go
```diff
@@ -29,13 +29,11 @@ func (f *Factory) register(id string, app *AppSpec) {
...
-		if err != nil {
-			logrus.Errorf("Failed to create a copy of spec object: %#v Err: %v", d, err)
-			return nil, err
+		dCopy := deepcopy.Copy(d)
+		if copy, ok := dCopy.(*AppSpec); ok {
+			return copy, nil
 		}
-
-		return dCopy.(*AppSpec), nil
+		return nil, fmt.Errorf("error creating copy of sepc: %v", d)
```
piyush-nimbalkar: *spec

PR #165, opened 2018-05-14
Add DeepCopy for AppSpec

disrani-px:
The kubernetes APIs are no longer in 1.10 which causes issues
in projects that vendor torpedo. The recommended way is to generate
deepcopy functions, but AppSpec is simple enough.

### DIFF
--- a/drivers/scheduler/spec/spec.go
+++ b/drivers/scheduler/spec/spec.go
@@ -14,3 +14,18 @@ type AppSpec struct {
 	// Enabled indicates if the application is enabled in the factory
 	Enabled bool
 }
+
+// DeepCopy Creates a copy of the AppSpec
+func (in *AppSpec) DeepCopy() *AppSpec {
+	if in == nil {
+		return nil
+	}
+	out := new(AppSpec)
+	out.Key = in.Key
+	out.Enabled = in.Enabled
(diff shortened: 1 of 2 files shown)


drivers/scheduler/spec/spec.go
```diff
@@ -14,3 +14,18 @@ type AppSpec struct {
...
+func (in *AppSpec) DeepCopy() *AppSpec {
+	if in == nil {
+		return nil
+	}
+	out := new(AppSpec)
+	out.Key = in.Key
+	out.Enabled = in.Enabled
+	out.SpecList = make([]interface{}, 0)
+	for _, spec := range in.SpecList {
+		out.SpecList = append(out.SpecList, spec)
```
piyush-nimbalkar: not sure how this will work when spec is a pointer
as we discussed, in future when we run tests in parallel, and modify the specs, it could be a problem

disrani-px: Yes, will need to be changed when we start running tests in parallel.

piyush-nimbalkar: Okay; let's put a TODO here and/or create an issue to track it.

disrani-px: Created Issue #166

PR #168, opened 2018-06-05
Retry the check for PX running when waiting for driver to be up

Rohit-PX:


drivers/volume/portworx/portworx.go
```diff
@@ -624,10 +625,19 @@ func (d *portworx) WaitDriverUpOnNode(n node.Node) error {
 	}
 
 	// Check if PX pod is up
-	if !d.schedOps.IsPXAppRunningOnNode(n) {
-		return fmt.Errorf("PX pod is not up on node: %s", n.Name)
+	t = func() (interface{}, bool, error) {
+		if !d.schedOps.IsPXAppRunningOnNode(n) {
```
harsh-px: rename `IsPXAppRunningOnNode` is `IsPXReadyOnNode`

drivers/volume/portworx/portworx.go
```diff
@@ -624,10 +625,19 @@ func (d *portworx) WaitDriverUpOnNode(n node.Node) error {
...
+			return "", true, &ErrFailedToWaitForPx{
+				Node:  n,
+				Cause: fmt.Sprintf("PX is not running on %s after %v", n.Name, validatePXStartTimeout),
```
harsh-px: Change text to "PX is not ready on.."

drivers/volume/portworx/schedops/schedops.go
```diff
@@ -30,7 +30,7 @@ type Driver interface {
 	// UpgradePortworx upgrades portworx to the given docker image and tag
 	UpgradePortworx(image, tag string) error
 	//IsPXPodRunningOnNode returns true if PX pod is up on that node, else returns false
```
harsh-px: update the function comment too

PR #170, opened 2018-06-12
Add liveliness probe to Wordpress installer

sam-myers:
Signed-off-by: Sam Myers <sam@portworx.com>

### DIFF
--- a/drivers/scheduler/sidecars/scripts/wp-install.sh
+++ b/drivers/scheduler/sidecars/scripts/wp-install.sh
@@ -19,6 +19,7 @@ function am_i_first() {
 
 function sleep_forever() {
     log ":: Sleeping forever. Exec into me for debugging"
+    touch /wordpress/alive
     while true; do sleep 30; done;
 }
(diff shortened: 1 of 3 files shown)


drivers/scheduler/sidecars/scripts/wp-install.sh
```diff
@@ -19,6 +19,7 @@ function am_i_first() {
 
 function sleep_forever() {
     log ":: Sleeping forever. Exec into me for debugging"
+    touch /wordpress/alive
```
harsh-px: Let's not put the `alive` file on /wordpress since that is a persisted path. Put it on /tmp so on each pod restart, the alive file isn't present.

harsh-px:
@sam-myers can you cherry pick to the 1.7 branch too?

sam-myers:
@harsh-px done

https://github.com/portworx/torpedo/commit/a674075ac8e4237657fa507b9d4ec42461f48acd

PR #174, opened 2018-06-30
Fix stork snap check

harsh-px:


drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -244,12 +245,21 @@ func (k *k8sSchedOps) validateVolumeClone(parent *api.Volume, parentAnnotation s
...
+		if ok && snapName == snapshotAnnotation {
+			return nil
+		}
+	}
+
 	parentName := parent.Locator.Name
-	if parentName != snapshotAnnotation {
-		return fmt.Errorf("Parent name [%s] does not match the snapshot annotation "+
-			"[%s] on the clone PVC", parentName, snapshotAnnotation)
+	if parentName == snapshotAnnotation {
```
disrani-px: This will never match, right?

harsh-px: I think old stork version will match this. That's why this check was present before.

PR #184, opened 2018-08-24
Run mysql app using encrypted volumes.

adityadani:


drivers/scheduler/k8s/specs/nginx/px-nginx-storage.yaml
```diff
@@ -6,14 +15,16 @@ metadata:
...
 ---
 ##### Portworx persistent volume claim
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
    name: px-nginx-pvc
    annotations:
+     px/secret-name: volume-secrets
+     px/secret-namespace: portworx
+     px/secret-key: nginx-secret
```
harsh-px: we support encryption for shared volumes?

adityadani: No we don't. This storage class does not have shared flag.

drivers/scheduler/k8s/specs/nginx/px-nginx-storage.yaml
```diff
@@ -6,14 +15,16 @@ metadata:
 provisioner: kubernetes.io/portworx-volume
 parameters:
    repl: "3"
-   shared: "true"
```
harsh-px: we can't do this. This reduces shared volumes coverage.

harsh-px: Use other existing apps like mysql, postgres or elasticsearch which already don't have shared:true

adityadani: Isn't wordpress currently using shared?

harsh-px: Yes, but we need multiple shared apps. We already have postgres, mysql, elasticsearch and wordpress-mysql for regular volumes. So need equal coverage for shared volumes.  

The current torpedo`wordpress` stack is a complex beast and we have to disable that many times due to app issues. Also it has shared and non-shared volumes in the same stack so we need nginx for only shared volume tests.

adityadani: It does not make sense to have nginx only on shared just because wordpress is using both shared and non-shared together. From a testing shared volumes perspective, I don't see any difference in a stack using both volume type vs a stack that using only one.
I am assuming we are running multiple apps together anyway.

If wordpress is disabled, then yes we loose coverage for shared volumes.
I will create another spec directory nginx-enc for this use case.

adityadani:
I have updated the PR to use mysql.

PR #190, opened 2018-09-25
Fixed px driver init fail if one of node does not  have portworx installed

ram-infrac:
This PR adds check before updating node in px driver if corresponding node has px/enabled=false/  kubernetes.dcos.io/node-type=public  /  or node has any taints 

This is how I tested changes ->
- I created 4 node kubernetes cluster , disabled px installation on one of worker node using px/enabled=false and installed portworx on remaining two worker node.
- Applied label kubernetes.dcos.io/node-type=public 
- Marked node as tainted using `kubectl taint nodes node-name key:value:NoSchedule`

`
[root@ram-horn-seer-0 ~]# kubectl describe  node ram-horn-seer-1

Name:               ram-horn-seer-1
Roles:              ingress,node
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/os=linux
                    elasticdata-vol-elasticsearch-data-0=true
                    kubernetes.dcos.io/node-type=private
                    kubernetes.io/hostname=ram-horn-seer-1
                    mysql-data=true
                    node-role.kubernetes.io/ingress=true
                    node-role.kubernetes.io/node=true
                    postgres-data=true
                    px-nginx-pvc=true
                    px/enabled=true
                    wp-pv-claim=true
Annotations:        node.alpha.kubernetes.io/ttl=0
                    volumes.kubernetes.io/controller-managed-attach-detach=true
CreationTimestamp:  Tue, 25 Sep 2018 09:05:33 -0700
Taints:             dedicated=foo:PreferNoSchedule
`

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -97,8 +97,10 @@ func (d *portworx) Init(sched string, nodeDriver string) error {
 	}
 
 	for _, n := range node.GetWorkerNodes() {
-		if err := d.WaitDriverUpOnNode(n); err != nil {
-			return err
+		if n.IsDriverInstalled {
+			if err := d.WaitDriverUpOnNode(n); err != nil {
+				return err
+			}
 		}
 	}
 
@@ -127,15 +129,22 @@ func (d *portworx) updateNodes(pxNodes []api.Node) error {
 
 func (d *portworx) updateNode(n node.Node, pxNodes []api.Node) error {
 	for _, address := range n.Addresses {
+		isPX := d.schedOps.IsPXInstalled(n)
 		for _, pxNode := range pxNodes {
 			if address == pxNode.DataIp || address == pxNode.MgmtIp || n.Name == pxNode.Hostname {
 				n.VolDriverNodeID = pxNode.Id
+				n.IsDriverInstalled = isPX
+				node.UpdateNode(n)
+				return nil
+			} else if !isPX {
+				n.IsDriverInstalled = false
 				node.UpdateNode(n)
 				return nil
 			}
 		}
 	}
-
+	// This is when we get node where label is not applied but for some reason
+	// PX not installed
 	return fmt.Errorf("failed to find px node for node: %v PX nodes: %v", n, pxNodes)
 }
 
@@ -499,7 +508,7 @@ func (d *portworx) StopDriver(nodes []node.Node, force bool) error {
 	for _, n := range nodes {
 		logrus.Infof("Stopping volume driver on %s.", n.Name)
 		if force {
-			pxCrashCmd := "sudo kill -9 px-storage"
+			pxCrashCmd := "sudo pkill -9 px-storage"
 			_, err = d.nodeDriver.RunCommand(n, pxCrashCmd, node.ConnectionOpts{
 				Timeout:         crashDriverTimeout,
 				TimeBeforeRetry: defaultRetryInterval,
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -96,11 +96,9 @@ func (d *portworx) Init(sched string, nodeDriver string) error {
 		return err
 	}
 
-	for _, n := range node.GetWorkerNodes() {
-		if n.IsDriverInstalled {
-			if err := d.WaitDriverUpOnNode(n); err != nil {
-				return err
-			}
+	for _, n := range node.GetStorageDriverNodes() {
+		if err := d.WaitDriverUpOnNode(n); err != nil {
+			return err
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -499,7 +508,7 @@ func (d *portworx) StopDriver(nodes []node.Node, force bool) error {
 	for _, n := range nodes {
 		logrus.Infof("Stopping volume driver on %s.", n.Name)
 		if force {
-			pxCrashCmd := "sudo kill -9 px-storage"
+			pxCrashCmd := "sudo pkill -9 px-storage"
```
ram-infrac: this should be pkill right? or either grep process ID of px-storage and then use kill -9?

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -479,6 +483,25 @@ func (k *k8sSchedOps) IsPXReadyOnNode(n node.Node) bool {
 	return true
 }
 
+// IsPXInstalled returns true/false if px label is enabled/disabled on kubernetes node
+func (k *k8sSchedOps) IsPXInstalled(n node.Node) bool {
+	isPxInstalled := true
+	kubeNode, err := k8s.Instance().GetNodeByName(n.Name)
+	if err != nil {
```
piyush-nimbalkar: return error if you are not able to find the node
also add a small retry before you error out; may be 30 seconds or 1 min at an interval of 5 seconds

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -479,6 +483,25 @@ func (k *k8sSchedOps) IsPXReadyOnNode(n node.Node) bool {
...
+		logrus.Errorf("Failed to get node %v", err)
+		return false
+	}
+
+	// if node has px/enabled label set to false or node-type public or
+	// has any taints then make above assumption wrong
+	if kubeNode.Labels[pxenabled] == "false" || kubeNode.Labels[nodeType] == "public" || len(kubeNode.Spec.Taints) > 0 {
```
piyush-nimbalkar: @harsh-px do we want to return false on any taint? or only for NoSchedule/PreferNoSchedule?

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -479,6 +483,25 @@ func (k *k8sSchedOps) IsPXReadyOnNode(n node.Node) bool {
...
+		logrus.Infof("[%v] Px Disabled: ", n.Name)
+		isPxInstalled = false
+	}
+	logrus.Infof("PX status on node %v %v\n", n.Name, isPxInstalled)
```
piyush-nimbalkar: nit: no need to add \n at the end

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -479,6 +483,25 @@ func (k *k8sSchedOps) IsPXReadyOnNode(n node.Node) bool {
...
+func (k *k8sSchedOps) IsPXInstalled(n node.Node) bool {
+	isPxInstalled := true
+	kubeNode, err := k8s.Instance().GetNodeByName(n.Name)
+	if err != nil {
+		logrus.Errorf("Failed to get node %v", err)
+		return false
+	}
+
+	// if node has px/enabled label set to false or node-type public or
+	// has any taints then make above assumption wrong
```
piyush-nimbalkar: nit: ... has any taints then px is disabled on this node

drivers/volume/portworx/schedops/schedops.go
```diff
@@ -29,8 +29,10 @@ type Driver interface {
 	GetServiceEndpoint() (string, error)
 	// UpgradePortworx upgrades portworx to the given docker image and tag
 	UpgradePortworx(image, tag string) error
-	//IsPXReadyOnNode returns true if PX pod is up on that node, else returns false
+	// IsPXReadyOnNode returns true if PX pod is up on that node, else returns false
 	IsPXReadyOnNode(n node.Node) bool
+	// IsPXInstalled returns true is px/enable flag is disabled/not applied on node
```
piyush-nimbalkar: nit: return true if portworx is enabled on given node
keep it generic

also can we rename this method to IsPxEnabled() instead of IsPxInstalled()

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -37,7 +37,11 @@ const (
 	//   the name of the stork volume snapshot
 	storkSnapshotNameKey = "stork-snap"
 	// pvcLabel is the label used on volume to identify the pvc name
-	pvcLabel               = "pvc"
+	pvcLabel = "pvc"
+	// pxenable is label used to check whethere px installation is enabled/disabled on node
+	pxEnabled = "px/enabled"
+	// nodeType is label used to check kubernetes node-type
+	nodeType               = "kubernetes.dcos.io/node-type"
```
harsh-px: rename to `dcosNodeType`

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -479,6 +483,35 @@ func (k *k8sSchedOps) IsPXReadyOnNode(n node.Node) bool {
...
+	if err != nil {
+		logrus.Errorf("Failed to get node %v", err)
+		return false, err
+	}
+
+	kubeNode := node.(*corev1.Node)
+	// if node has px/enabled label set to false or node-type public or
+	// has any taints then px is disabled on node
+	if kubeNode.Labels[pxEnabled] == "false" || kubeNode.Labels[nodeType] == "public" || len(kubeNode.Spec.Taints) > 0 {
+		logrus.Infof("[%v] Px Disabled: ", n.Name)
```
harsh-px: you can remove this log since `isPxEnabled` is already getting printed below

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -479,6 +483,35 @@ func (k *k8sSchedOps) IsPXReadyOnNode(n node.Node) bool {
...
+		isPxEnabled = false
+	}
+	logrus.Infof("PX status on node %v %v", n.Name, isPxEnabled)
```
harsh-px: change log to be more explicit.
```
if isPxEnabled {
  logrus.Infof("PX is enabled on node %v.", n.Name)`

} else {
  logrus.Infof("PX is not enabled on node %v. Will be skipped for tests.", n.Name)`
}
```

ram-infrac: don't need ixPXEnabled here, we can directly return from label/px check

PR #192, opened 2018-09-28
Add back readiness and liveness probe for mysql specs in torpedo

thiguetta:


drivers/scheduler/k8s/specs/mysql/px-mysql-app.yaml
```diff
@@ -40,6 +40,19 @@ spec:
           value: password
         ports:
         - containerPort: 3306
+        livenessProbe:
+          exec:
+            command: ["sh", "-c", "mysqladmin -u root -p$MYSQL_ROOT_PASSWORD ping"]
+          initialDelaySeconds: 360
```
harsh-px: We can't use such a high `initialDelaySeconds`. This would mean mysql will stay 0/1 until 360 seconds. Use a higher failureThreshold instead. This will ensure that if mysql is ready earlier, it will make it 1/1.

Refer to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#probe-v1-core

PR #193, opened 2018-09-28
Install wordpress plugins from local zip files

Rohit-PX:


vendor/github.com/portworx/sched-ops/k8s/k8s.go
```diff
@@ -860,6 +860,7 @@ func (k *k8sOps) WaitForPodDeletion(uid types.UID, namespace string, timeout tim
 		p, err := k.GetPodByUID(uid, namespace)
 		if err != nil {
 			if err == ErrPodsNotFound {
+				logrus.Debug("RK=> Pod Not found error found. Exiting retry loop.")
```
piyush-nimbalkar: remove the sched-ops vendor

drivers/scheduler/sidecars/scripts/wp-install.sh
```diff
@@ -60,14 +60,14 @@ wp core install --url="${WORDPRESS_URL}" --title="TestWordPress" --admin_user="a
...
     log ":: Installing plugin ${plugin_name}"
-    wp plugin install "${plugin_name}" --activate
+    wp plugin install /home/www-data/"${plugin_name}" --activate
 done
 
 log ":: Installing theme..."
 wp theme install "${WORDPRESS_THEME}" --activate
 
-sleep_forever
+#sleep_forever
```
piyush-nimbalkar: you can keep this for debugging
else remove the line and the function entirely

PR #194, opened 2018-10-01
Check for cores on all nodes after torpedo job is done

thiguetta:


drivers/node/node.go
```diff
@@ -100,6 +111,8 @@ type Driver interface {
 
 	// RecoverDrive recovers the given drive from failure on the given node.
 	RecoverDrive(node Node, driveNameToRecover string, driveUUID string, options ConnectionOpts) error
+
```
harsh-px: Add comment on what the `SystemCheck` method should do

drivers/node/errors.go
```diff
@@ -83,3 +83,13 @@ type ErrFailedToRecoverDrive struct {
...
 }
+
+// ErrFailedToSystemCheck error type when we fail to check for core files
+type ErrFailedToSystemCheck struct {
+	Node Node
+	Cause string
+}
+
+func (e *ErrFailedToSystemCheck) Error() string {
+	return fmt.Sprintf("Failed to check core file on: %v. Cause: %v", e.Node.Name, e.Cause)
```
harsh-px: this should have details about "core file". The core file details will be in the Cause.

This should be `fmt.Sprintf("System check failed on: %v. Cause: %v", e.Node.Name, e.Cause)`

tests/basic/basic_test.go
```diff
@@ -301,6 +301,7 @@ var _ = Describe("{AppScaleUpAndDown}", func() {
 })
 
 var _ = AfterSuite(func() {
+	CheckForCoreFiles()
```
harsh-px: Add `CheckForCoreFiles` in `AfterSuite` of other tests too (tests/reboot, tests/volume_ops, tests/upgrade, tests/drive_failure)

tests/common.go
```diff
@@ -295,6 +295,23 @@ func CollectSupport() {
 	})
 }
 
+// CheckForCoreFiles check if cores files are present on each node
+func CheckForCoreFiles() {
```
harsh-px: Instead of `CheckForCoreFiles`, call this `PerformSystemCheck`.

Checking of core files is just an implementation detail for the node driver. At a high level here, we are just doing a system check.

tests/common.go
```diff
@@ -295,6 +295,23 @@ func CollectSupport() {
...
+	context(fmt.Sprintf("checking for core files..."), func() {
+		Step(fmt.Sprintf("verifying if core files are present on each node"), func() {
+			nodes := node.GetWorkerNodes()
+			expect(nodes).NotTo(beEmpty())
+			for _, n := range nodes {
+				file, err := Inst().N.SystemCheck(n)
+				if err != nil {
+					logrus.Warnf("failed to check system. err: %v", err)
```
harsh-px: don't ignore `err` here. Fail the test using an expect.

deployments/deploy-ssh.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash -x
+FOCUS_TESTS=VolumeReplicationIncrease,VolumeReplicationDecrease
+SCALE_FACTOR=1
+APP_LIST=elasticsearch,mysql,nginx,postgres,wordpress
+TORPEDO_IMG=thiagogonzaga/torpedo:v1.1.3
```
harsh-px: Don't change any of these variables. Looks like these are from your testing script.

PR #196, opened 2018-10-04
Add test to stop kubelet and check for new pods to start on other nodes

thiguetta:


tests/sched/sched_test.go
```diff
@@ -0,0 +1,91 @@
...
+					Expect(err).NotTo(HaveOccurred())
+					Expect(appNodes).NotTo(BeEmpty())
+				})
+
+				Step(fmt.Sprintf("stop scheduler service"), func() {
+					for _, nodeToStopService := range appNodes {
+						err := Inst().S.StopSchedOnNode(nodeToStopService)
+						Expect(err).NotTo(HaveOccurred())
+					}
+					Step("wait for the service to stop", func() {
```
harsh-px: Did this test pass on your run?

For a repl 3 application (e.g nginx), `appNodes` will be all the nodes in the cluster. If you stop the scheduler on all nodes, the app has nowhere to go.

The test needs to 

1. pick only one node for an app
2. Stop scheduler on that node
3. Wait and validate the app is back healthy (time for eviction threshold to cross and new pod to come up)
4. Start scheduler on that node
5. Move to the next app ctx

tests/sched/sched_test.go
```diff
@@ -0,0 +1,91 @@
...
+					})
+				})
+			}
+		})
+
+		ValidateAndDestroy(contexts, nil)
+	})
+})
+
+var _ = Describe("{StartScheduler}", func() {
```
harsh-px: We don't need  a test to start scheduler. The start scheduler interface is only useful in above test where after stopping, you need to restore it.

tests/sched/sched_test.go
```diff
@@ -0,0 +1,91 @@
+package tests
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo"
```
harsh-px: you also need to update https://github.com/portworx/torpedo/blob/master/deployments/deploy-ssh.sh#L111 with the .test file for this new test.

harsh-px:
> This needs to be run with a stateful set since that behaves differently in this scenario than deployments.


@thiguetta tested this with all the apps. Failed with PX 1.5.1 and passed with master as expected.

PR #197, opened 2018-10-04
Configurable timeout for kubernetes driver

ram-infrac:
This PR adds following ,
- vendor update for stork (govendor)
- vendor update for sched-ops(govendor)
- Update torpedo scheduler driver to use updated interfaces in sched-ops

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -628,11 +628,11 @@ func (k *k8s) substituteNamespaceInVolumes(volumes []v1.Volume, ns string) []v1.
 	return updatedVolumes
 }
 
-func (k *k8s) WaitForRunning(ctx *scheduler.Context) error {
+func (k *k8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time.Duration) error {
 	k8sOps := k8s_ops.Instance()
 	for _, spec := range ctx.App.SpecList {
 		if obj, ok := spec.(*apps_api.Deployment); ok {
-			if err := k8sOps.ValidateDeployment(obj); err != nil {
+			if err := k8sOps.ValidateDeployment(obj, timeout, retryInterval); err != nil {
 				return &scheduler.ErrFailedToValidateApp{
 					App:   ctx.App,
 					Cause: fmt.Sprintf("Failed to validate Deployment: %v. Err: %v", obj.Name, err),
@@ -670,7 +670,7 @@ func (k *k8s) WaitForRunning(ctx *scheduler.Context) error {
 
 			logrus.Infof("[%v] Validated Rule: %v", ctx.App.Key, svc.Name)
 		} else if obj, ok := spec.(*v1.Pod); ok {
-			if err := k8sOps.ValidatePod(obj); err != nil {
+			if err := k8sOps.ValidatePod(obj, timeout, retryInterval); err != nil {
 				return &scheduler.ErrFailedToValidatePod{
 					App:   ctx.App,
 					Cause: fmt.Sprintf("Failed to validate Pod: [%s] %s. Err: Pod is not ready %v", obj.Namespace, obj.Name, obj.Status),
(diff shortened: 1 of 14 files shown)


ram-infrac:
Does hardcoed timeout in driver/k8s also needs to be configurable , those are used with `task.DoRetryWithTimeout` ?

tests/common.go
```diff
@@ -128,7 +133,7 @@ func ValidateVolumes(ctx *scheduler.Context) {
 	context("For validation of an app's volumes", func() {
 		var err error
 		Step(fmt.Sprintf("inspect %s app's volumes", ctx.App.Key), func() {
-			err = Inst().S.InspectVolumes(ctx)
+			err = Inst().S.InspectVolumes(ctx, 10*time.Second, 2*time.Second)
```
harsh-px: Make these constants

harsh-px: Also `ValidatePersistentVolumeClaim()` in sched-ops previously had a `validatePVCTimeout` of 5 minutes. This is now reduced to 10 seconds. 

As part of this change, ensure that the timeouts and intervals remaining the same. At some point, we will reduce the timeouts to more stricter limits.

ram-infrac: @harsh-px  I'm keeping max of timeout for all internal operation running, for eg in inspectVolume `ValidatePVCsForStatefulSet` has max timeout of 15 minutes so I will pass 15min timeout to inspectVolumes from tests that way we will not keep any hardcore values at `drivers/sched/k8s` side

harsh-px: 15 minutes is too much to allow a PVC to be bound. Just 5 minutes. In scheduler driver `InspectVolumes` implementation, use `Statefulset Replicas x timeout`, as the timeout for `ValidatePVCsForStatefulSet` since a statefulset has replica number of PVCs. For e.g If the timeout is given as 5 minutes and we have a statefulset with replica 3, all PVCs and Snapshot will be tested with timeout of 5 minutes. Statefulset PVCs will be tested with timeout of 5 * 3 minutes.

ram-infrac: sounds good, I have also made similar changes for ValidatePod in WaitForRunning which requires 10 Minute timeout

drivers/scheduler/k8s/k8s.go
```diff
@@ -990,7 +990,7 @@ func (k *k8s) InspectVolumes(ctx *scheduler.Context) error {
 				}
 			}
 
-			if err := k8sOps.ValidatePVCsForStatefulSet(ss); err != nil {
+			if err := k8sOps.ValidatePVCsForStatefulSet(ss, timeout*3, retryInterval); err != nil {
```
harsh-px: You cannot hardcoded 3 here. You need to lookup in the `ss` for it's replica count. Ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#statefulsetspec-v1-apps

drivers/scheduler/k8s/k8s.go
```diff
@@ -628,11 +628,11 @@ func (k *k8s) substituteNamespaceInVolumes(volumes []v1.Volume, ns string) []v1.
 	return updatedVolumes
 }
 
-func (k *k8s) WaitForRunning(ctx *scheduler.Context) error {
+func (k *k8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time.Duration) error {
 	k8sOps := k8s_ops.Instance()
 	for _, spec := range ctx.App.SpecList {
 		if obj, ok := spec.(*apps_api.Deployment); ok {
-			if err := k8sOps.ValidateDeployment(obj); err != nil {
+			if err := k8sOps.ValidateDeployment(obj, timeout, retryInterval); err != nil {
```
harsh-px: need to use timeout instead of `statefulSetValidateTimeout` below too. All timeouts in `WaitForRunning` should be proportional to the given `timeout`.

drivers/scheduler/k8s/k8s.go
```diff
@@ -670,7 +670,7 @@ func (k *k8s) WaitForRunning(ctx *scheduler.Context) error {
 
 			logrus.Infof("[%v] Validated Rule: %v", ctx.App.Key, svc.Name)
 		} else if obj, ok := spec.(*v1.Pod); ok {
-			if err := k8sOps.ValidatePod(obj); err != nil {
+			if err := k8sOps.ValidatePod(obj, timeout*2, retryInterval); err != nil {
```
harsh-px: why `*2` here?

drivers/scheduler/k8s/k8s.go
harsh-px: Yes as it takes that much time for the entire ss to come up and all of it's PVCs to be bound.

drivers/scheduler/k8s/k8s.go
harsh-px: Just use the given input timeout to WaitForRunning. 10 minutes was too much and was used to workaround an issue with the wordpress app.

PR #201, opened 2018-10-12
Fix import path of golint

rodrigc:
Same fix as here:
https://github.com/libopenstorage/stork/pull/157#issue-222576593


rodrigc:
Already fixed

PR #202, opened 2018-10-15
Add test to resize PVC

thiguetta:


harsh-px:
@thiguetta the travis build is failing with
`./drivers/scheduler/k8s/k8s.go:1129:1: receiver name d should be consistent with previous receiver name k for k8s
`

PR #203, opened 2018-10-17
fix volume replication test to increase/decrease repl for each app

thiguetta:


drivers/scheduler/k8s/k8s.go
```diff
@@ -1169,13 +1171,17 @@ func (k *k8s) ResizeVolume(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
-	extraAmount, _ := resource.ParseQuantity("1Gi")
-	storageSize.Add(extraAmount)
-	pvc.Spec.Resources.Requests[v1.ResourceStorage] = storageSize
-	if _, err := k8sOps.UpdatePersistentVolumeClaim(pvc); err != nil {
-		return nil, &scheduler.ErrFailedToResizeStorage{
-			App:   ctx.App,
-			Cause: err.Error(),
+
+	// TODO this test is required since stork snapshot doesn't support resizing, remove when feature is added
+	if resizeSupported, _ := strconv.ParseBool(pvc.Annotations["torpedo/resize-supported"]); resizeSupported{
```
harsh-px: Won't this skip resize for PVCs that don't have that annotation?

harsh-px: Use const for `torpedo/resize-supported`.

drivers/scheduler/k8s/k8s.go
```diff
@@ -47,6 +47,8 @@ const (
 	// SystemdSchedServiceName is the name of the system service resposible for scheduling
 	// TODO Change this when running on openshift for the proper service name
 	SystemdSchedServiceName = "kubelet"
+	// ResizeSupported is the parameter key for the annotation when a pvc is resizeble or not
+	ResizeSupported = "torpedo/resize-supported"
```
harsh-px: `ResizeSupported` doesn't need to be exported. Call it `resizeSupportedAnnotationKey`

drivers/scheduler/k8s/k8s.go
```diff
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"strconv"
```
harsh-px: This import needs to be after line 13 so it gets sorted by goimports.

PR #205, opened 2018-10-18
Add check to verify whether orphaned pod dir has a portworx volume or…

thiguetta:


drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -328,33 +334,32 @@ func (k *k8sSchedOps) ValidateVolumeCleanup(d node.Driver) error {
...
-			findFileOpts := node.FindOpts{
-				ConnectionOpts: connOpts,
-				MinDepth:       2,
-				MaxDepth:       2,
-			}
-			files, _ := d.FindFiles(path, n, findFileOpts)
-			if len(strings.TrimSpace(files)) > 0 {
-				dirtyVolPods = append(dirtyVolPods, podUID)
+			// Check if /var/lib/kubelet/pods/{podUID}/volumes/kubernetes.io~portworx-volume is empty
+			if isDirEmpty(path, n, d) {
```
piyush-nimbalkar: shouldn't this be `!isDirEmpty()`?

thiguetta: Hey @piyush-nimbalkar  as per my understanding of the discussion we had on the channel, if this dir is empty it is orphan otherwise it is dirty, does it make sense?

piyush-nimbalkar: That's correct. If the portworx-volume dir is empty, then why would we check the contents inside it.
If the dir is not empty, then we go and check if there is a pvc directory or not to decide it is dirty or just orphan, right?

harsh-px:
@piyush-nimbalkar  is correct. I have reverted this PR.

drivers/volume/portworx/schedops/k8s-schedops.go
thiguetta: Sorry, now I’ve got what you mean, fixing

PR #209, opened 2018-10-24
Add support to apply and validate CRD objects

ram-infrac:
This PR add support for applying & validating custom specs for clusterpair and migration object. Also does following -
 -  get remote PX node from destination cluster using kubeconfig file provided.
 -  Add clusterpairinfo() api which will return cluster pair information by querying remote node on k8s cluster

*Note* :  I have manually added sched-ops vendor file changes required for this PR. PR is already opened and yet to be reviewed in https://github.com/portworx/sched-ops/pull/83.

Way I tested this changes ->  vendored this changes into stork/integration_test and created clusterpair and migration objects.


drivers/volume/portworx/portworx.go
```diff
@@ -1075,6 +1115,18 @@ func (d *portworx) getClusterManagerByAddress(addr string) (cluster.Cluster, err
 	return clusterclient.ClusterManager(cClient), nil
 }
 
+func (d *portworx) getClusterToken(n node.Node, op string) (*client.Response, error) {
+	url := d.constructURL(n.Addresses[0])
```
disrani-px: There is an openstorage API for this

ram-infrac: @disrani-px openstorage build in current torpedo points to **v3.0.0-beta**, which is way behind current master branch. Since cluster pair api is added recently to openstorage we need to vendor in latest commit . Please let me know if it's ok, I have local commit already created for this changes

drivers/volume/portworx/portworx.go
```diff
@@ -1060,6 +1057,44 @@ func (d *portworx) UpgradeDriver(version string) error {
...
+	if err != nil {
+		return err
+	}
+
+	name := "objVolume"
+	req := &api.VolumeCreateRequest{
+		Locator: &api.VolumeLocator{Name: name},
+		Source:  &api.Source{},
+		Spec: &api.VolumeSpec{
+			Size:    50,
```
ram-infrac: what size should we keep here?

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1465,49 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
 	return nil
 }
 
+func (k *k8s) CreateCRDObjects(pathCRDSpec string) error {
```
harsh-px: This is an anti-pattern with how other things are done currently. Can we do the parsing at the same place other spec files are parsed and this thing just takes in a context?

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -583,6 +609,32 @@ func extractPodUID(volDirPath string) string {
...
+	if err != nil {
+		logrus.Errorf("Unable to get k8s instance: %v", err)
+		return workerNode, err
+	}
+	nodes, err := destClient.GetNodes()
+	if err != nil {
+		return workerNode, err
+	}
+
+	// TODO(ram-infrac) :find px node, right now it's assumed that px is installed
```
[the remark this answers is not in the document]
ram-infrac: yes , will update

drivers/scheduler/k8s/k8s.go
harsh-px: Can you share where in stork integration test are we creating the file?

As as when possible, I want to keep the same pattern for all interfaces. Don't want one interface to take a file and other take pre-loaded contexts.

ram-infrac:
I agree, but don't know where this should go in case of cloud Migration. @disrani-px what do you think?

drivers/scheduler/k8s/k8s.go
ram-infrac: That is not part of stork integration test yet, because creation of cluster pair depend on this PR for getting cluster token. 
This is my tryout branch if you want to have look -  https://github.com/libopenstorage/stork/compare/rs-ClusterMigration?expand=1#diff-6e48c877f1215e1e1f64e38abe60a903R33

drivers/volume/portworx/portworx.go
disrani-px: This should be an API in the volume interface. Shouldn't get called automatically. The test should call it if required.

disrani-px: Actually, this is very PX specific (and will go away envtually), so shouldn't be called from the integration test itself. So this should be done before the test starts.

ram-infrac: removed it from torpedo

drivers/scheduler/k8s/k8s.go
ram-infrac: Corrected to take context and use those for applying spec file. Since we have ```storkctl ``` command line to generate spec file we can use that as prerequisite to create and dump clusterpair spec file

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1465,46 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
...
+					Name:  obj.Name,
+					Cause: fmt.Sprintf("Failed to validate cluster Pair: %v. Err: %v", obj.Name, err),
+				}
+			}
+
+		} else if obj, ok := specObj.(*stork_api.Migration); ok {
+			logrus.Info("Applying Migration Spec")
+			err = k8sOps.CreateMigration(obj)
+			if err != nil {
+				logrus.Info("Error Creating Migration", err)
```
disrani-px: Infof() and need to add %v

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1465,46 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
...
+// CreateCRDObjects and Validate their deployment
+func (k *k8s) CreateCRDObjects(ctx *scheduler.Context, timeout, retryInterval time.Duration) error {
+	var err error
+	k8sOps := k8s_ops.Instance()
+	for _, specObj := range ctx.App.SpecList {
+		if obj, ok := specObj.(*stork_api.ClusterPair); ok {
+			logrus.Info("Applying clusterpair specs")
+			err = k8sOps.CreateClusterPair(obj)
+			if err != nil {
+				logrus.Info("Error Creating Cluster Pair", err)
```
disrani-px: Infof() and need to add %v

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1465,46 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
 	return nil
 }
 
+// CreateCRDObjects and Validate their deployment
+func (k *k8s) CreateCRDObjects(ctx *scheduler.Context, timeout, retryInterval time.Duration) error {
+	var err error
+	k8sOps := k8s_ops.Instance()
+	for _, specObj := range ctx.App.SpecList {
+		if obj, ok := specObj.(*stork_api.ClusterPair); ok {
+			logrus.Info("Applying clusterpair specs")
```
disrani-px: spec

drivers/scheduler/k8s/k8s.go
disrani-px: Need to return err here instead of continuing forward.

drivers/scheduler/errors.go
```diff
@@ -285,4 +285,16 @@ type ErrFailedToStartSchedOnNode struct {
 
 func (e *ErrFailedToStartSchedOnNode) Error() string {
 	return fmt.Sprintf("Failed to start scheduler service %v on node: %v due to err: %v", e.SystemService, e.Node, e.Cause)
-}
\ No newline at end of file
+}
+
+// ErrFailedToApplyCustomSpec error type when CRD objects does not applied successfully
+type ErrFailedToApplyCustomSpec struct {
```
disrani-px: Should also add type here.

drivers/volume/portworx/portworx.go
```diff
@@ -1050,6 +1057,30 @@ func (d *portworx) UpgradeDriver(version string) error {
 	return nil
 }
 
+// GetClusterPairingInfo return underlying storage details
+func (d *portworx) GetClusterPairingInfo() (map[string]string, error) {
+	pairInfo := make(map[string]string)
+	pxNodes, err := d.schedOps.GetRemotePXNodes(remoteKubeConfigPath)
+	if err != nil {
+		logrus.Errorf("err retriving remote px nodes: %v", err)
+		return pairInfo, err
```
disrani-px: return nil, err

drivers/volume/portworx/portworx.go
```diff
@@ -1050,6 +1057,30 @@ func (d *portworx) UpgradeDriver(version string) error {
...
+	}
+
+	clusterMgr, err := d.getClusterManagerByAddress(pxNodes[0].Addresses[0])
+	resp, err := clusterMgr.GetPairToken(true)
+	if err != nil {
+		return pairInfo, err
```
disrani-px: return nil, err

drivers/volume/portworx/portworx.go
```diff
@@ -1050,6 +1057,30 @@ func (d *portworx) UpgradeDriver(version string) error {
...
+// GetClusterPairingInfo return underlying storage details
+func (d *portworx) GetClusterPairingInfo() (map[string]string, error) {
+	pairInfo := make(map[string]string)
+	pxNodes, err := d.schedOps.GetRemotePXNodes(remoteKubeConfigPath)
+	if err != nil {
+		logrus.Errorf("err retriving remote px nodes: %v", err)
+		return pairInfo, err
+	}
+
+	clusterMgr, err := d.getClusterManagerByAddress(pxNodes[0].Addresses[0])
```
disrani-px: Check for err

drivers/volume/portworx/portworx.go
```diff
@@ -1050,6 +1057,30 @@ func (d *portworx) UpgradeDriver(version string) error {
...
+	resp, err := clusterMgr.GetPairToken(true)
```
disrani-px: true will reset the token every time. You shouldn't do that.

drivers/volume/portworx/portworx.go
```diff
@@ -1050,6 +1057,30 @@ func (d *portworx) UpgradeDriver(version string) error {
...
+	if err != nil {
+		return pairInfo, err
+	}
+	logrus.Info("Response for token:", resp.Token)
```
disrani-px: Use Infof() with %v

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -583,6 +609,31 @@ func extractPodUID(volDirPath string) string {
...
 }
 
+// return PX node on k8s cluster provided by kubeconfig file
+func getPXNode(destKubeConfig string) (corev1.Node, error) {
+	var workerNode corev1.Node
+	// get schd-ops/k8s instance of destination cluster
+	destClient, err := k8s.NewInstance(destKubeConfig)
+	if err != nil {
+		logrus.Errorf("Unable to get k8s instance: %v", err)
+		return workerNode, err
```
disrani-px: return nil, err

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -583,6 +609,31 @@ func extractPodUID(volDirPath string) string {
...
+	}
+	nodes, err := destClient.GetNodes()
+	if err != nil {
+		return workerNode, err
```
disrani-px: return nil, err

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -583,6 +609,31 @@ func extractPodUID(volDirPath string) string {
...
+	}
+
+	// get label on node where PX is Enabled
+	for _, node := range nodes.Items {
+		if node.Labels[pxEnabled] == "true" {
+			workerNode = node
```
disrani-px: You can do `return node, nil` here. And at the end of the function return an error since that will be reached if no node was found.

drivers/volume/portworx/portworx.go
```diff
@@ -34,6 +35,12 @@ const (
 	exitMaintenancePath     = "/exitmaintenance"
 	pxSystemdServiceName    = "portworx.service"
 	storageStatusUp         = "Up"
+	tokenKey                = "token"
+	clusterIP               = "clusterip"
+	clusterPort             = "clusterport"
+	remoteKubeConfigPath    = "/tmp/kubeconfig"
+	tokenPath               = "/v1/cluster/pairtoken"
```
disrani-px: tokenPath and objectstorPath aren't used

drivers/volume/portworx/portworx.go
```diff
@@ -1050,6 +1057,30 @@ func (d *portworx) UpgradeDriver(version string) error {
 	return nil
 }
 
+// GetClusterPairingInfo return underlying storage details
+func (d *portworx) GetClusterPairingInfo() (map[string]string, error) {
+	pairInfo := make(map[string]string)
+	pxNodes, err := d.schedOps.GetRemotePXNodes(remoteKubeConfigPath)
+	if err != nil {
+		logrus.Errorf("err retriving remote px nodes: %v", err)
```
disrani-px: retrieving

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -534,6 +534,32 @@ func (k *k8sSchedOps) IsPXEnabled(n node.Node) (bool, error) {
...
 
+// GetStorageInfo returns cluster pair info from destination cluster refereced by kubeconfig
+func (k *k8sSchedOps) GetRemotePXNodes(destKubeConfig string) ([]node.Node, error) {
+	var addrs []string
+	var pxNodes []node.Node
+
+	pxNode, err := getPXNode(destKubeConfig)
+	if err != nil {
+		logrus.Errorf("Error getting Node %v : %v", pxNode, err)
+		return pxNodes, err
```
disrani-px: return nil, err

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -534,6 +534,32 @@ func (k *k8sSchedOps) IsPXEnabled(n node.Node) (bool, error) {
 	return true, nil
 }
 
+// GetStorageInfo returns cluster pair info from destination cluster refereced by kubeconfig
+func (k *k8sSchedOps) GetRemotePXNodes(destKubeConfig string) ([]node.Node, error) {
```
disrani-px: Wrong comment for function. Also, this is always returning one node. So can be called GetRemotePXNode and can return just node.Node instead of an array.
Or, return all the nodes from the remote cluster.

drivers/volume/portworx/portworx.go
ram-infrac: default value should be false then?

disrani-px: Yes

drivers/volume/portworx/schedops/k8s-schedops.go
ram-infrac: that was earlier plan to get all px nodes for future node

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1465,47 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
...
+			logrus.Info("Applying clusterpair spec")
+			err = k8sOps.CreateClusterPair(obj)
+			if err != nil {
+				logrus.Errorf("Error Creating Cluster Pair %v", err)
+				return err
+			}
+			if err := k8sOps.ValidateClusterPair(obj.Name, defaultTimeout, defaultRetryInterval); err != nil {
+				return &scheduler.ErrFailedToApplyCustomSpec{
+					Name:  obj.Name,
+					Cause: fmt.Sprintf("Failed to validate cluster Pair: %v. Err: %v", obj.Name, err),
```
disrani-px: Add type to the error. No need to print clusterPair in the message. Same for migration below.

drivers/volume/portworx/portworx.go
```diff
@@ -1050,6 +1055,33 @@ func (d *portworx) UpgradeDriver(version string) error {
...
+		return nil, err
+	}
+	resp, err := clusterMgr.GetPairToken(false)
+	if err != nil {
+		return nil, err
+	}
+	logrus.Infof("Response for token: %v", resp.Token)
+
+	// file up cluster pair info
+	pairInfo[clusterIP] = pxNodes[0].Addresses[0]
```
disrani-px: pxNodes can be an empty array, so the length should be checked before accessing it.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1465,47 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
...
+func (k *k8s) CreateCRDObjects(ctx *scheduler.Context, timeout, retryInterval time.Duration) error {
+	var err error
+	k8sOps := k8s_ops.Instance()
+	for _, specObj := range ctx.App.SpecList {
+		if obj, ok := specObj.(*stork_api.ClusterPair); ok {
+			logrus.Info("Applying clusterpair spec")
+			err = k8sOps.CreateClusterPair(obj)
+			if err != nil {
+				logrus.Errorf("Error Creating Cluster Pair %v", err)
+				return err
```
disrani-px: Return ErrFailedToApplyCustomSpec. Same for Migration below

PR #213, opened 2018-11-01
Add check whether a volume is mounted on a host or not

thiguetta:


drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -179,10 +180,32 @@ func (k *k8sSchedOps) ValidateVolumeSetup(vol *volume.Volume) error {
...
+
+		// ignore error when a command not exactly fail, like grep when empty return exit 1
+		connOpts := node.ConnectionOpts{
+			TimeBeforeRetry: defaultRetryInterval,
+			Timeout: defaultTimeout,
+			IgnoreError: true,
+		}
+
+		out, _ := d.RunCommand(currentNode,
+			fmt.Sprintf("mount | grep %s", pvName), connOpts)
```
harsh-px: Just check the pv name is not sufficient since the path could be mounted on a non-px device and we need to catch that.

So do the same checks (pxd, pxfs etc) that we do for checking the mounts inside the container.

PR #214, opened 2018-11-01
Enable shared encrypted volume

thiguetta:


drivers/scheduler/k8s/specs/mysql/px-mysql-storage.yaml
```diff
@@ -53,6 +53,16 @@ parameters:
...
 ---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: px-mysql-sc-shared
+provisioner: kubernetes.io/portworx-volume
+parameters:
+  repl: "3"
+  shared: "true"
+allowVolumeExpansion: true
```
harsh-px: Add this to nginx and a new pvc and new mount point. We want to keep all shared volumes to nginx and wordpress.

PR #215, opened 2018-11-01
Get k8sOps instance using kubeconfig file

ram-infrac:
This PR allow user to call WaitForRunning() with kubeConfig file of k8s server, so that we can check application status running on given server.

If no config file path passed then k8s instance will be set to default inclusterConfig() instance. 

**Note** : This is proposed change, please let me know what do you think and if there is better alternative for achieving this.

### DIFF
--- a/drivers/scheduler/scheduler.go
+++ b/drivers/scheduler/scheduler.go
@@ -20,8 +20,9 @@ const (
 
 // Context holds the execution context of a test task.
 type Context struct {
-	UID string
-	App *spec.AppSpec
+	UID        string
+	App        *spec.AppSpec
+	KubeConfig string
 }
 
 // ScheduleOptions are options that callers to pass to influence the apps that get schduled
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -1251,7 +1251,7 @@ func (k *k8s) GetNodesForApp(ctx *scheduler.Context) ([]node.Node, error) {
 			}
 		}
 
-		if len(result) > 0{
+		if len(result) > 0 {
 			return result, false, nil
 		}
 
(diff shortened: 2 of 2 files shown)


drivers/scheduler/scheduler.go
```diff
@@ -20,8 +20,9 @@ const (
 
 // Context holds the execution context of a test task.
 type Context struct {
-	UID string
-	App *spec.AppSpec
+	UID        string
+	App        *spec.AppSpec
+	KubeConfig string
```
piyush-nimbalkar: Context object represents an application running in the system like mysql, elasticsearch, etc.
Why would we pass a different kubeconfig for every app?

piyush-nimbalkar:
Did not completely understand the purpose of the change. Are you trying to point to a different k8s cluster from the one running torpedo? If yes, you can always set the KUBECONFIG env variable and pass in the torpedo pod.

Else you could also do this -
```
import "k8s.io/client-go/tools/clientcmd"
config, _ := clientcmd.BuildConfigFromFlags("", "path/to/kubeconfig")
```
Dinesh recently introduced the SetConfig() method which you could use as well.

drivers/scheduler/scheduler.go
disrani-px: We want to verify the same app is running on a remote cluster after migration

disrani-px:
> Did not completely understand the purpose of the change. Are you trying to point to a different k8s cluster from the one running torpedo? If yes, you can always set the KUBECONFIG env variable and pass in the torpedo pod.
> 
> Else you could also do this -
> 
> ```
> import "k8s.io/client-go/tools/clientcmd"
> config, _ := clientcmd.BuildConfigFromFlags("", "path/to/kubeconfig")
> ```
> Dinesh recently introduced the SetConfig() method which you could use as well.

We want to make calls to both the local and remote kubernetes clusters. I think this change is error prone though, all callers will need to make sure to call the new API.

@ram-infrac Could you not just use SetConfig(remoteConfig) when you want to make remote calls and then SetConfig(nil) when you want to make local calls from the test itself?

ram-infrac:
No @disrani-px that won't do. For eg. from stork integration test lets say I have scheduled mysql pod on cluster and then Waiting to come it in running state. After migration done , I have to check application status running on cluster 2 using WaitForRunning() torpedo call and at end destroy mysql app on cluster 2 and then cluster 1.

SetConfig()/NewInstance() we have to add those in torpedo API's  instead of current **k8s.Instance()** or else need to rewrite everywhere in tests

PR #217, opened 2018-11-03
Add missing GetClusterPairInfo() Interface

ram-infrac:
This PR adds Missing GetClusterPairingInfo() API.
   - also add additional check detecting PX Nodes
   - check for empty pxNodes
- adds RescanSpecs() from given specDir
- add customCRDobjects under Schedule() call

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -710,6 +710,18 @@ func (k *k8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time
 			}
 
 			logrus.Infof("[%v] Validated pod: %v", ctx.App.Key, obj.Name)
+		} else if err := k8sOps.ValidateClusterPair(obj.Name, defaultTimeout, defaultRetryInterval); err != nil {
+			return &scheduler.ErrFailedToApplyCustomSpec{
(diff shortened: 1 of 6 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -1482,51 +1503,45 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
...
+				Cause: fmt.Sprintf("Failed to apply spec: %v. Err: %v", obj.Name, err),
+				Type:  obj,
 			}
-			if err := k8sOps.ValidateClusterPair(obj.Name, defaultTimeout, defaultRetryInterval); err != nil {
-				return &scheduler.ErrFailedToApplyCustomSpec{
-					Name:  obj.Name,
-					Cause: fmt.Sprintf("Failed to validate cluster Pair: %v. Err: %v", obj.Name, err),
-				}
+		}
+		if err := k8sOps.ValidateClusterPair(obj.Name, defaultTimeout, defaultRetryInterval); err != nil {
```
disrani-px: This function should only create the object, the validation should be done in WaitForRunning()

ram-infrac: Moved

PR #225, opened 2018-12-11
Add support to use ssh key in torpedo deploy

thiguetta:


drivers/node/ssh/ssh.go
```diff
@@ -43,11 +44,13 @@ func getKeyFile(keypath string) (ssh_pkg.Signer, error) {
 	file := keypath
 	buf, err := ioutil.ReadFile(file)
 	if err != nil {
+		logrus.Errorf("err: %s", err.Error())
```
harsh-px: print detail in the log .. something like Failed to read key file due to...

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
```diff
@@ -71,6 +71,8 @@ spec:
           value: "false"
         - name: "ES_JAVA_OPTS"
           value: "-Xms256m -Xmx256m"
+        - name: "NETWORK_HOST"
+          value: "_eth0_"
```
harsh-px: do you have a reference online that suggests to use this?

thiguetta: I found it here and worked for our case https://github.com/pires/kubernetes-elasticsearch-cluster/issues/64

drivers/node/ssh/ssh.go
harsh-px: @thiguetta ^^

drivers/node/ssh/ssh.go
```diff
@@ -43,11 +44,13 @@ func getKeyFile(keypath string) (ssh_pkg.Signer, error) {
 	file := keypath
 	buf, err := ioutil.ReadFile(file)
 	if err != nil {
+		logrus.Errorf("failed to reade ssh key file. Cause: %s", err.Error())
```
harsh-px: typo for `reade`.

PR #227, opened 2018-12-17
Fix Stopping volume driver made px cluster to go out of quorum

thiguetta:


tests/basic/basic_test.go
```diff
@@ -191,9 +192,19 @@ var _ = Describe("{VolumeDriverAppDown}", func() {
 					Expect(appNodes).NotTo(BeEmpty())
 				})
 
+				//get the nodes to be down randomly
+				appNodesLen := len(appNodes)
+				amountAllowedNodesDown := len(node.GetWorkerNodes()) % appNodesLen
```
harsh-px: You need to handle below scenario too

- 5 node cluster
- 2 appNodes, node1 and node2
- The volumes is repl=2 and node1 and node2 are both the replicas of the volume

The current code will allow node1 and node1 to do down and hence all nodes for that volume will be down. We don't want that now. We will add a out of quorum test later on.

Also below logic where you are populating nodesToBeDown can be better. You are doing

```
nodesToBeDown = nodesToBeDown[:amountAllowedNodesDown]
```

But amountAllowedNodesDown can be greater than actual appNodes. So you only need to copy upto the min(appNodes,  amountAllowedNodesDown)

tests/basic/basic_test.go
```diff
@@ -191,9 +192,19 @@ var _ = Describe("{VolumeDriverAppDown}", func() {
...
+				idxs := rand.Perm(appNodesLen)
+				nodesToBeDown := make([]node.Node, appNodesLen)
+				for i := 0; i < i; i++ {
```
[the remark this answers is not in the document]
thiguetta: you're right, it should be `appNodesLen`

tests/basic/basic_test.go
thiguetta: you're right, it's not taking into account replication factor and if all the nodes with replicas are down

drivers/volume/portworx/portworx.go
```diff
@@ -1166,6 +1166,40 @@ func (d *portworx) getStorageStatus(n node.Node) string {
...
+		for _, n := range rs.Nodes {
+			pxNode, err := d.clusterManager.Inspect(n)
+			if err != nil {
+				return nil, &ErrFailedToInspectVolume{
+					ID:    torpedovol.Name,
+					Cause: fmt.Sprintf("Failed to inspect replica set node: %s err: %v", n, err),
+				}
+			}
+
+			pxNodes = append(pxNodes, pxNode.Hostname)
```
harsh-px: If pxNode.SchedulerNodeName is available, use that instead of Hostname.

tests/basic/basic_test.go
```diff
@@ -191,9 +192,36 @@ var _ = Describe("{VolumeDriverAppDown}", func() {
...
+					for _, vol := range appVolumes {
+						replicas, err := Inst().V.GetReplicaSetNodes(vol)
+						Expect(err).NotTo(HaveOccurred())
+						Expect(replicas).NotTo(BeEmpty())
+						// at least 1 node with replica needs to be up
+						nodesThatCantBeDown[replicas[0]] = true
+					}
+
+					for _, node := range node.GetWorkerNodes() {
+						if _, exists := nodesThatCantBeDown[node.Name]; !exists {
```
harsh-px: 3 node PX cluster. repl=3 volumes.

Above code will keep 1 PX node up and bring down other 2 nodes. This will bring PX cluster out of quorum and hence all volumes are out of quorum.

thiguetta:
@harsh-px http://70.0.0.41:8080/view/Torpedo%20Dashboard/job/tp-2.0-vol-driver-thiago/15/console

PR #231, opened 2019-02-04
Adding Nomad driver

nikolaypopov:
Adding Nomad driver and vendor


drivers/scheduler/n0mad/n0mad.go
```diff
@@ -0,0 +1,229 @@
+package n0mad
+
```
harsh-px: does the package name have a zero instead of an o?

drivers/scheduler/n0mad/n0mad.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/portworx/torpedo/drivers/scheduler"
+	"github.com/portworx/torpedo/drivers/scheduler/spec"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	// SchedName is the name of the kubernetes scheduler driver implementation
+	SchedName = "n0mad"
+	// N0madMaster is a nodemad api concact (subject to change)
+	N0madMaster = "http://70.0.57.114:4646"
```
harsh-px: why is a 70.X IP address hardcoded in a driver imp?

harsh-px:
I'm going to close this PR here as this is a work in progress.

PR #235, opened 2019-03-06
add rancher deploy script

thiguetta:
Add a new environment variable to define whether it is rancher or generic kubernetes and change deploy node: master when generic kubernetes, controlplane when rancher

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -79,6 +79,21 @@ if [ -n "${TORPEDO_SSH_KEY_MOUNT}" ]; then
     VOLUME_MOUNTS="${VOLUME_MOUNTS},${TORPEDO_SSH_KEY_MOUNT}"
 fi
 
+K8S_VENDOR_KEY=""
+if [ -n "${K8S_VENDOR}" ]; then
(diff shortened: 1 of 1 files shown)


deployments/deploy-rancher.sh
```diff
@@ -0,0 +1,184 @@
...
+metadata:
+  name: torpedo
+  labels:
+    app: torpedo
+spec:
+  tolerations:
+  - key: node-role.kubernetes.io/controlplane
+    operator: Equal
+    value: "true"
+  - key: node-role.kubernetes.io/etcd
```
harsh-px: @thiguetta as discussed yesterday on slack, parameterize the keys in tolerations and the affinity sections and merge into single deploy script. 

We will have many distros of k8s being added to torpedo in near future so we need to avoid adding new scripts wherever some parametrization is possible.

thiguetta:
tests passing for both 
kubernetes: https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/1/console
rancher: https://jenkins.portworx.co/job/Torpedo/job/tp-rancher-basic/34/console (all tests passing, but the job failed on trying to save testresults)

PR #237, opened 2019-03-13
Migration schedules and dep for vendoring

harsh-px:
1. adds handlers for migration schedules
2. Switches vendoring from glide to dep
3. Add vendor rule in travis so it catches broken vendoring rules


drivers/scheduler/scheduler.go
```diff
@@ -35,6 +35,11 @@ func (in *Context) DeepCopy() *Context {
 	return out
 }
 
+// GetNamespace returns the namespace for the context
+func (in *Context) GetNamespace() string {
```
disrani-px: As I had mentioned offline, this actually breaks the scheduler abstraction, since there is no concept of "namespace" in DC/OS

harsh-px: Do you prefer renaming this to GetID(). So this will be the context's unique ID. For k8s users, this ID is implicitly the namespace.

PR #240, opened 2019-03-20
Add APIs to add stop/start labels to PX service.

Rohit-PX:
Why is this PR needed?
* https://portworx.atlassian.net/browse/PTX-1247

Impacts:
* Torpedo tests that stop PX driver
* Stork integration tests

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -568,7 +568,7 @@ func (d *portworx) StopDriver(nodes []node.Node, force bool) error {
 				return err
 			}
 		} else {
-			err = d.schedOps.DisableOnNode(n)
+			err = d.schedOps.StopPxOnNode(n)
 			if err != nil {
 				return err
 			}
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -1020,6 +1020,10 @@ func (d *portworx) testAndSetEndpoint(endpoint string) error {
 }
 
 func (d *portworx) StartDriver(n node.Node) error {
+	err := d.schedOps.StartPxOnNode(n)
+	if err != nil {
+		return err
+	}
 	return d.nodeDriver.Systemctl(n, pxSystemdServiceName, node.SystemctlOpts{
 		Action: "start",
 		ConnectionOpts: node.ConnectionOpts{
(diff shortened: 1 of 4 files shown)


drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -78,6 +84,14 @@ func (e *errLabelAbsent) Error() string {
 
 type k8sSchedOps struct{}
 
+func (k *k8sSchedOps) EnableOnNode(n node.Node) error {
```
harsh-px: instead of enable and disable, let's name them appropriately as start and stop as that's what the APIs do.

In the past, they might have been named incorrectly.

drivers/volume/portworx/portworx.go
```diff
@@ -568,6 +568,10 @@ func (d *portworx) StopDriver(nodes []node.Node, force bool) error {
 				return err
 			}
 		} else {
+			err = d.schedOps.DisableOnNode(n)
```
harsh-px: In the start driver impl, you need to ensure the label is reset. Just call the start sched-ops API and that will add px/service=start. It will be harmless.

drivers/volume/portworx/schedops/k8s-schedops.go
Rohit-PX: Fixed. 
`EnableOnNode ==> StartPxOnNode`
`DisableOnNode  ==> StopPxOnNode`

drivers/volume/portworx/portworx.go

Rohit-PX:
Testing it here:
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20Dashboard/job/tp-2-1-0-basic/7/console

harsh-px:
Merged once volume driver down test passes

Rohit-PX:
Passed here: https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20Dashboard/job/tp-fp-basic/327/console

drivers/volume/portworx/schedops/schedops.go
```diff
@@ -11,6 +11,10 @@ import (
 
 // Driver is the interface for portworx operations under various schedulers
 type Driver interface {
+	// EnableOnNode enable portworx on given node
```
rodrigc: @Rohit-PX comments should be fixed to reflect the real function name

PR #241, opened 2019-03-21
Add APIs to add stop/start labels to PX service. (#240)

Rohit-PX:
* Add APIs to add stop/start labels to PX service.


drivers/volume/portworx/schedops/schedops.go
```diff
@@ -11,6 +11,10 @@ import (
 
 // Driver is the interface for portworx operations under various schedulers
 type Driver interface {
+	// EnableOnNode enable portworx on given node
```
harsh-px: can you fix the function comments for start and stop? They are still called enable and disable

PR #244, opened 2019-03-24
PTX-1200 fix torpedo to support the new api response

thiguetta:


drivers/volume/portworx/portworx.go
```diff
@@ -519,11 +521,16 @@ func (d *portworx) ValidateUpdateVolume(vol *torpedovolume.Volume) error {
 	return nil
 }
 
+func errIsNotFound(err error) bool {
+	statusErr, _ := status.FromError(err)
+	return statusErr.Code() == codes.NotFound || strings.Contains(err.Error(), "code = NotFound")
```
harsh-px: Just check the code here. Skip the message.

thiguetta: we can't, it's a REST API the answer is string, so it can't convert to grpc status, so I added both methods as temporary solution as discussed with @lpabon

lpabon: Correct. status.FromError() expects that the error is of a certain type which is handled when using the grpc client.

harsh-px: Ah okay. Did you test this on a 2.1 job?

PR #247, opened 2019-03-29
Update torpedo to use openstorage release-6.0

adityadani:


harsh-px:
@thiguetta Just FYI this change will merge soon. We don't think this should break anything since torpedo is compatible with PX 2.1 now.

PR #251, opened 2019-04-03
PTX-1304 fix a torpedo bug which makes internal kvdb cluster go out of quorum causing cores

thiguetta:


harsh-px:
The PR still doesn't fix the internal kvdb issue.

It just fixes the torpedo logic to select nodes that can be down. Internal kvdb will require another fix. Please rename the PR title and commit messages to reduce confusion.

drivers/volume/portworx/portworx.go
```diff
@@ -49,7 +49,7 @@ const (
 	maintenanceWaitTimeout           = 2 * time.Minute
 	inspectVolumeTimeout             = 10 * time.Second
 	inspectVolumeRetryInterval       = 2 * time.Second
-	validateDeleteVolumeTimeout      = 3 * time.Minute
+	validateDeleteVolumeTimeout      = 5 * time.Minute
```
harsh-px: why is the timeout bumped? A PVC should get deleted in 3 minutes.

thiguetta: sometime it's taking longer than that, actually no error for that

thiguetta: actually this piece of code belong to another ticket I only put in here to test three problems at once

harsh-px: revert this timeout. If deletes are timing out, we need to root cause the actual issue in PX.

thiguetta: ok, I might need some help on that since I haven't found any error in the logs

harsh-px: Sure, I can help take a look if you have a cluster where PVC delete took more than 3 minutes. First place to look is the controller manager logs which will have the errors reported on the PVC delete call to Portworx

tests/basic/basic_test.go
```diff
@@ -209,19 +209,20 @@ func getNodesThatCanBeDown(ctx *scheduler.Context) []node.Node {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(replicas).NotTo(BeEmpty())
 			// at least n-1 nodes with replica need to be up
-			for i := 0; i <= len(replicas)-1; i++ {
-				nodesThatCantBeDown[replicas[i]] = true
+			maxNodesToBeDown := len(replicas)/2 + 1
```
harsh-px: for a repl 2 volume, this logic will shutdown both replica nodes.

PR #252, opened 2019-04-10
PTX-1323 add test to torpedo to reproduce pods getting stuck in termi…

thiguetta:
…nating state


tests/basic/basic_test.go
```diff
@@ -316,6 +316,53 @@ var _ = Describe("{AppScaleUpAndDown}", func() {
...
+				err := Inst().S.WaitForDestroy(ctx)
+				Expect(err).NotTo(HaveOccurred())
+			}
+		})
+		Step("teardown all apps", func() {
+			for _, ctx := range contexts {
+				TearDownContext(ctx, nil)
+			}
+		})
+		Step("Cordon all nodes but one", func() {
```
harsh-px: The step comment should be uncordon

tests/basic/basic_test.go
```diff
@@ -316,6 +316,53 @@ var _ = Describe("{AppScaleUpAndDown}", func() {
...
+			for _, ctx := range contexts {
+				err := Inst().S.Destroy(ctx, opts)
+				Expect(err).NotTo(HaveOccurred())
+			}
+		})
+		Step("Validate destroy", func() {
+			opts := make(map[string]bool)
+			opts[scheduler.OptionsWaitForResourceLeakCleanup] = true
+			for _, ctx := range contexts {
+				err := Inst().S.WaitForDestroy(ctx)
```
harsh-px: what is the current timeout for waitfordestory? And is that per context?

harsh-px: As discussed offline, let's fix the Destroy and WaitForDestroy timeouts.

harsh-px:
@thiguetta CI is failing

drivers/scheduler/dcos/dcos.go
```diff
@@ -475,6 +475,16 @@ func (d *dcos) RescanSpecs(specDir string) error {
 	return nil
 }
 
+func (d *dcos) EnableSchedulingOnNode(n node.Node) error {
+	// TODO implement this method
+	return nil
```
harsh-px: instead of nil, can you change this to return a not supported error? Let's start doing that for all new interface methods that aren't supported by these drivers.

The pattern to use is something like this: https://github.com/libopenstorage/cloudops/blob/master/unsupported/unsupported.go

Have all drivers inherit an unsupported driver.

Gopkg.toml
```diff
@@ -44,8 +44,9 @@
   branch = "release-4.0"
 
 [[constraint]]
-  branch = "master"
+  branch = "PTX-1620"
```
harsh-px: first merge https://github.com/portworx/sched-ops/pull/134#pullrequestreview-259120232 and then revert this back to master

drivers/scheduler/dcos/dcos.go
harsh-px: This is not done

tests/basic/basic_test.go
harsh-px: I was expecting WaitForDestroy to now take in a timeout.

drivers/scheduler/dcos/dcos.go
thiguetta: done

tests/basic/basic_test.go
thiguetta: this is inside WaitForDestroy https://github.com/portworx/torpedo/pull/252/commits/dd929e3c9517bf5dbc6809a5217c3141b6a3277a

harsh-px: The `defaultTimeout` inside WaitForDestroy won't work. The reason we added custom timeouts to the API was so that callers can use different timeouts based on their use case.

The timeout for WaitForDestroy needs to be proportional to number of apps inside ctx. We had seen that this test was timing out as we had pods at a larger scale.

PR #253, opened 2019-04-11
PTX-1001 fix upgrade when custom px image

thiguetta:


thiguetta:
tests passing
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20Upgrade%20Dashboard/job/tp-upgrade-1.6.1.1-2.1/5/
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20Upgrade%20Dashboard/job/tp-upgrade-1.7.7-2.1/5/

drivers/volume/portworx/portworx.go
```diff
@@ -1039,27 +1039,45 @@ func (d *portworx) StartDriver(n node.Node) error {
 }
 
 func (d *portworx) UpgradeDriver(version string) error {
+	customImageRegex := regexp.MustCompile(`oci=(?P<oci>([\w][\w._-]+\/)?[\w][\w._-]*\:[\w][\w.-]*),px=(?P<px>([\w][\w._-]+\/)?[\w][\w._-]*\:[\w][\w.-]*)`)
```
harsh-px: can u give examples of what this regex matches to?

thiguetta: any valid docker image in the following 
oci=<any valid docker image>,px=<any valid docker image>

i.e
oci=repo/image:tag,px=repo/image:tag
oci=repo/image,px=repo/image
oci=image:tag,px=image:tag
oci=image,px=image

harsh-px: Make version a proper input type (struct) instead of a special string pattern that the implementation must parse.

The struct can have the oci and px image as items.

For any interface, the function definition must be clear and the arguments must be self-documenting so that users can know how to invoke them.

thiguetta: correct me if I'm wrong, but volume interface is driver agnostic, so if we add a different structure we might tie it to portworx driver

harsh-px: Add a generic input which has list of images. Each item has an image type and image version. Both of these are strings.

tests/upgrade/upgrade_test.go
```diff
@@ -60,7 +77,8 @@ var _ = PDescribe("{UpgradeDowngradeVolumeDriver}", func() {
 		}
 
 		Step("start the upgrade of volume driver", func() {
-			err := Inst().V.UpgradeDriver(Inst().StorageDriverUpgradeVersion)
+			images := getImages(Inst().StorageDriverUpgradeVersion)
```
harsh-px: Update documentation for `StorageDriverUpgradeVersion` on the expected CSV format

harsh-px: @thiguetta let's handle the above comment and then we should be good to merge.

PR #254, opened 2019-04-11
PTX-1325 Add IsScalable interface and check whether an application can scale or not

thiguetta:


drivers/scheduler/k8s/k8s.go
```diff
@@ -1659,6 +1662,28 @@ func (k *k8s) StartSchedOnNode(n node.Node) error {
...
 
+func (k *k8s) IsScalable(spec interface{}) bool {
+	if obj, ok := spec.(*apps_api.Deployment); ok {
+		dep, err := k8s_ops.Instance().GetDeployment(obj.Name, obj.Namespace)
+		if err != nil {
+		}
+		for _, vol := range dep.Spec.Template.Spec.Volumes {
+			pvc, err := k8s_ops.Instance().GetPersistentVolumeClaim(vol.PersistentVolumeClaim.ClaimName, dep.Namespace)
+			if err != nil {
+
```
harsh-px: Err return missing you here. Same above.

PR #255, opened 2019-04-12
PTX-1241 fix a bug when pod is mounted in other host

thiguetta:
this PR fixes a problem in a specific case when two pods with the same non-shared volume reach Running/Ready state but one of them has the volume in read only, it means the volume in read only is mounted in the pod but not in the host itself, so we skip host mount check

### DIFF
--- a/drivers/volume/portworx/schedops/k8s-schedops.go
+++ b/drivers/volume/portworx/schedops/k8s-schedops.go
@@ -270,6 +270,7 @@ func (k *k8sSchedOps) validateMountsInPods(
 		}
 
 		containerPaths := getContainerPVCMountMap(*pod)
+		skipHostMountCheck := false
 		for containerName, path := range containerPaths {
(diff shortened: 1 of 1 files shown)


thiguetta:
test running https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20Dashboard/job/tp-2.1-vol-driver/12/

thiguetta:
test passed

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -287,6 +288,8 @@ func (k *k8sSchedOps) validateMountsInPods(
 				if len(pxMounts) > 0 {
 					logrus.Debugf("pod: [%s] %s has PX mount: %v", pod.Namespace, pod.Name, pxMounts)
 					pxMountFound = true
+					// in case there are two pods running with non shared volume, one of them will be in read-only
+					skipHostMountCheck = isMountReadOnly(line)
```
harsh-px: @thiguetta how does this prevent from ignoring a read only mount for an actual bug?

So a single (repl 1 deployment) pod mounted a non-shared volume and the mount became read only.

thiguetta: I think this answer your question https://jenkins.portworx.co/job/Torpedo/job/tp-master-kops/48/display/redirect

this `if` won't mark pod as validated, so it will keep retrying and skipping in a loop generating a timeout

PR #256, opened 2019-04-24
[PTX-1330] Run torpedo pod on GKE worker node where px is not installed

vinayakshnd:
1. Run torpedo pod on GKE worker node where px is not installed. 
2. Pass `TORPEDO_SSH_KEY` as ENV var to torpedo

Contract between `Spawn` and `Torpedo` is:

1. One of the GKE worker node will have label `px/enabled=false`. This node will be used to run torpedo pod. PX will not be installed on this node.
2. Same node will also have `apps=false:NoSchedule` taints so that test application (MySql, nginx etc) pods will not get scheduled on it.
3. Torpedo pod will have toleration for `apps=false:NoSchedule` taint so that it can run on that node.

Jenkins Job run with above changes:
https://jenkins.portworx.co/job/tp-2-1-gke/20/console

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -88,6 +88,10 @@ if [ -n "${K8S_VENDOR}" ]; then
         rancher)
             K8S_VENDOR_KEY=node-role.kubernetes.io/controlplane
             ;;
+        gke)
+            # Run torpedo on worker node, where px installation is disabled. 
+            K8S_VENDOR_KEY=px/enabled
+            ;;
     esac
 else
     K8S_VENDOR_KEY=node-role.kubernetes.io/master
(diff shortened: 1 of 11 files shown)


deployments/deploy-ssh.sh
```diff
@@ -88,6 +88,10 @@ if [ -n "${K8S_VENDOR}" ]; then
         rancher)
             K8S_VENDOR_KEY=node-role.kubernetes.io/controlplane
             ;;
+        gke)
+            # Run torpedo on worker node, where px installation is disabled. 
+            K8S_VENDOR_KEY=px/enabled
```
harsh-px: actually `px/enabled` needs to be equal to `false` instead of just existing.

Add `K8S_VENDOR_VALUE` and keep it empty for others. Also update nodeSelectorTerms below

vinayakshnd: Hi @harsh-px,  Updated PR to use In operator instead of Exists.

harsh-px:
@vinayakshnd you can merge this. Once merged, please monitor that atleast one official torpedo job from the dashboard passes on this.

PR #259, opened 2019-05-08
PTX-1352 update storageNodeId after reboot

thiguetta:


harsh-px:
@thiguetta did u test this on a jenkins job?

drivers/volume/portworx/portworx.go
```diff
@@ -1228,6 +1228,17 @@ func (d *portworx) GetReplicaSetNodes(torpedovol *torpedovolume.Volume) ([]strin
 	return pxNodes, nil
 }
 
+func (d *portworx) UpdateNodeID(n node.Node) (node.Node, error) {
```
harsh-px: @thiguetta instead of an explicit UpdateNodeID, it would be safer to always looking up the node ID every time the portworx driver needs to perform an operation based on ID.

thiguetta:
@harsh-px test passed https://jenkins.portworx.co/job/Torpedo/view/Torpedo%202.1/job/tp-2-1-kops/90/console ( I had to use ubuntu image since default kops is hanging to reboot)

PR #261, opened 2019-05-10
PTX-1327 adding mysql IO sidecar

thiguetta:


scripts/sidecar-mysql/entry-point.sh
```diff
@@ -0,0 +1,32 @@
...
+trap '' SIGTSTP
+
+cd /employees_db
+until mysql -h mysql -u root -p${MYSQL_ROOT_PASSWORD} -t < employees.sql
+do 
+	echo "adding data"
+	sleep 2
+done
+
+while :
```
harsh-px: looks like after this, it will only do reads (SELECT).

Let's do a mix of reads and writes in steady state.

Makefile
```diff
@@ -100,3 +100,7 @@ sidecar: sidecar-wp-cli
 sidecar-wp-cli:
 	docker build -t $(SIDECAR_WP_CLI_IMG) -f $(SIDECAR_DIR)/wp-cli.dockerfile $(SIDECAR_DIR)
 	docker push $(SIDECAR_WP_CLI_IMG)
+
+sidecar-mysql:
+	docker build -t $(SIDECAR_MYSQL_CLI_IMG) -f scripts/sidecar-mysql/Dockerfile scripts/sidecar-mysql/
```
nareshdesh: SIDECAR_MYSQL_CLI_IMG is not declared in the make file
"sidecar-mysql" is not added to the sidecar target list

scripts/sidecar-mysql/entry-point.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+set -xe
+trap '' SIGINT
+trap ''  SIGQUIT
+trap '' SIGTSTP
+
+cd /employees_db
+until mysql -h mysql -u root -p${MYSQL_ROOT_PASSWORD} -t < employees.sql
```
nareshdesh: MySQL server hostname should be configurable or passed as environment variable, otherwise one will have to create a service 'mysql' in the same namespace

scripts/sidecar-mysql/entry-point.sh
harsh-px: @thiguetta ^^

harsh-px:
@thiguetta let's resolve the comments on this one soon.

PR #264, opened 2019-05-19
PTX-1457: Add longevity support in torpedo

vilasd:


vilasd:
https://jenkins.portworx.co/job/Torpedo/job/torpedo-vilas/9/

PR #265, opened 2019-05-20
CSI support in torpedo drivers.

Rohit-PX:

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -36,6 +36,10 @@ if [ -n "${STORAGE_BASE_VERSION}" ]; then
     UPGRADE_BASE_VERSION_ARG="--storage-driver-base-version=$STORAGE_BASE_VERSION"
 fi
 
+if [ -n "${PROVISIONER}" ]; then
+	PROVISIONER="$PROVISIONER"
+fi
+
 if [ -z "${TORPEDO_IMG}" ]; then
     TORPEDO_IMG="portworx/torpedo:latest"
     echo "Using default torpedo image: ${TORPEDO_IMG}"
@@ -174,6 +178,7 @@ spec:
             "--app-list", "$APP_LIST",
             "--node-driver", "ssh",
             "--scale-factor", "$SCALE_FACTOR",
+            "--provisioner", "$PROVISIONER",
             "$UPGRADE_VERSION_ARG",
             "$UPGRADE_BASE_VERSION_ARG" ]
     tty: true
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -62,6 +62,16 @@ const (
 	resizeSupportedAnnotationKey = "torpedo/resize-supported"
 )
 
+const (
+	portworxStorage = "portworx"
+	csiStorage      = "csi"
+)
+
+var provisioners = map[string]string{
+	portworxStorage: "kubernetes.io/portworx-volume",
+	csiStorage:      "com.openstorage.pxd",
+}
+
 var (
 	namespaceRegex = regexp.MustCompile("{{NAMESPACE}}")
 )
@@ -281,7 +291,7 @@ func (k *k8s) Schedule(instanceID string, options scheduler.ScheduleOptions) ([]
 	for _, app := range apps {
 
 		appNamespace := app.GetID(instanceID)
-		specObjects, err := k.createSpecObjects(app, appNamespace)
+		specObjects, err := k.createSpecObjects(app, appNamespace, options.StorageProvisioner)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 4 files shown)


deployments/deploy-ssh.sh
```diff
@@ -189,6 +198,8 @@ spec:
             "--app-list", "$APP_LIST",
             "--node-driver", "ssh",
             "--scale-factor", "$SCALE_FACTOR",
+            "--config-map", "$CONFIG_MAP",
```
harsh-px: what is the use of -config-map in this PR?

tests/common.go
```diff
@@ -410,6 +418,8 @@ func ParseFlags() {
 				StorageDriverUpgradeVersion: volUpgradeVersion,
 				StorageDriverBaseVersion:    volBaseVersion,
 				AppList:                     appList,
+				ConfigMap:                   configMapName,
```
harsh-px: looks like your config map changes for security are showing up here? Let's keep this PR just for CSI changes.

drivers/scheduler/k8s/k8s.go
```diff
@@ -429,10 +439,21 @@ func (k *k8s) createNamespace(app *spec.AppSpec, namespace string) (*v1.Namespac
...
+
 	if obj, ok := spec.(*storage_api.StorageClass); ok {
+		logrus.Infof("[%v] Requested provisioner: %s", app.Key, storageprovisioner)
 		obj.Namespace = ns.Name
+		if storageProvisioner, ok := provisioners[storageprovisioner]; ok {
+			obj.Provisioner = storageProvisioner
+		} else {
+			return nil, &scheduler.ErrFailedToGetStorage{
+				App:   app,
+				Cause: fmt.Sprintf("Failed to get storage: %v. Err: %v", storageprovisioner, fmt.Errorf("Invalid storage provider")),
```
harsh-px: you don't need to fmt.Errorf here since you are just printing the string and not returning an err.

deployments/deploy-ssh.sh
Rohit-PX: it's not needed in this PR, will delete.

tests/common.go
Rohit-PX: it's not needed in this PR, will delete.

drivers/scheduler/k8s/k8s.go
Rohit-PX: Fixed.

deployments/deploy-ssh.sh
harsh-px: This is still present.

drivers/scheduler/scheduler.go
```diff
@@ -47,6 +47,8 @@ type ScheduleOptions struct {
 	AppKeys []string
 	// Nodes restricts the applications to get scheduled only on these nodes (Optional)
 	Nodes []node.Node
+	// StorageProvider identifies what storage provider should be used
```
harsh-px: the comment is incorrect.

drivers/scheduler/k8s/k8s.go
```diff
@@ -429,10 +439,21 @@ func (k *k8s) createNamespace(app *spec.AppSpec, namespace string) (*v1.Namespac
...
 }
 
-func (k *k8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.AppSpec) (interface{}, error) {
+func (k *k8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.AppSpec, storageprovisioner string) (interface{}, error) {
 	k8sOps := k8s_ops.Instance()
+
 	if obj, ok := spec.(*storage_api.StorageClass); ok {
+		logrus.Infof("[%v] Requested provisioner: %s", app.Key, storageprovisioner)
 		obj.Namespace = ns.Name
+		if storageProvisioner, ok := provisioners[storageprovisioner]; ok {
```
harsh-px: default to native driver if not provided.

drivers/scheduler/scheduler.go
Rohit-PX: changed comment to `StorageProvisioner`

drivers/scheduler/k8s/k8s.go
Rohit-PX: Fixed.

deployments/deploy-ssh.sh
Rohit-PX: Deleted.

drivers/scheduler/k8s/k8s.go
```diff
@@ -429,10 +439,18 @@ func (k *k8s) createNamespace(app *spec.AppSpec, namespace string) (*v1.Namespac
 	return nsObj.(*v1.Namespace), nil
 }
 
-func (k *k8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.AppSpec) (interface{}, error) {
+func (k *k8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.AppSpec, storageprovisioner string) (interface{}, error) {
 	k8sOps := k8s_ops.Instance()
+
 	if obj, ok := spec.(*storage_api.StorageClass); ok {
+		logrus.Infof("[%v] Requested provisioner: %s", app.Key, storageprovisioner)
```
harsh-px: print this only if it is non-empty. Else this will add an extra log line for default jobs too.

Rohit-PX: Done.

PR #275, opened 2019-06-14
PTX-1562 fix WaitDriverDownOnNode in torpedo

thiguetta:


drivers/volume/portworx/portworx.go
```diff
@@ -212,55 +217,38 @@ func (d *portworx) CleanupVolume(name string) error {
...
 }
 
-func (d *portworx) getPxNode(n node.Node, cManager cluster.Cluster) (api.Node, error) {
-	if cManager == nil {
-		cManager = d.getClusterManager()
-	}
-	pxNode, err := cManager.Inspect(n.VolDriverNodeID)
-	if (err == nil && pxNode.Status == api.Status_STATUS_OFFLINE) || (err != nil && pxNode.Status == api.Status_STATUS_NONE) {
+func (d *portworx) getPxNode(n node.Node /*, cManager cluster.Cluster*/) (api.StorageNode, error) {
+	//if cManager == nil {
```
harsh-px: remove commented code. same at other places in the PR.

harsh-px:
Is a torpedo job passing on this?

drivers/volume/portworx/portworx.go
```diff
@@ -937,35 +937,28 @@ func (d *portworx) WaitDriverUpOnNode(n node.Node, timeout time.Duration) error
...
-			}
-
-			pxNode, err := d.getPxNode(n, cManager)
-			if err != nil {
-				if regexp.MustCompile(`.+timeout|connection refused.*`).MatchString(err.Error()) {
-					logrus.Infof("px on node %s addr %s is down as inspect returned: %v",
-						n.Name, addr, err.Error())
-					continue
-				}
+	cManager, err := d.pickAlternateClusterManager(n)
```
harsh-px: you can easily get cManager from another node which is also down (if the test brings 2 nodes down).

In that case, you will get the connection refused or timed out from the 2nd node and the task will keep failing below.

In this case, I think it's best to use the portworx-service k8s service IP and use that IP address to inspect this node.

This service endpoints will always converge to only healthy PX nodes. If the service has 0 endpoints, all PX nodes are down.

You can add a get service endpoints API to the portworx sched-ops interface here that will inspect the portworx-service k8s object and return it's endpoints.

harsh-px: synced up offline with thiago on this. `pickAlternateClusterManager` internally checks if the picked node is good by during an enumerate. So we are good here.

PR #279, opened 2019-06-25
Changed one of mysql StorageClass to use xfs instead of ext4

mborodin-px:
Jenkins job: http://jenkins.portworx.co:8081/job/Torpedo/job/tp-dev-maxb/

### DIFF
--- a/drivers/scheduler/k8s/specs/mysql/px-mysql-storage.yaml
+++ b/drivers/scheduler/k8s/specs/mysql/px-mysql-storage.yaml
@@ -42,7 +42,7 @@ metadata:
     name: px-mysql-sc-seq
 provisioner: kubernetes.io/portworx-volume
 parameters:
-   fs: "ext4"
+   fs: "xfs"
    repl: "3"
    aggregation_level: "1"
    io_profile: "sequential"


harsh-px:
We can't merge this yet as torpedo jobs will start failing on Portworx versions that don't support this.

drivers/scheduler/k8s/specs/mysql/px-mysql-storage.yaml
```diff
@@ -42,7 +42,7 @@ metadata:
     name: px-mysql-sc-seq
 provisioner: kubernetes.io/portworx-volume
 parameters:
-   fs: "ext4"
+   fs: "xfs"
```
harsh-px: @mborodin-px for now, can you also add `force_unsupported_fs_type: "true"` to the StorageClass? This will ensure xfs is exercise on existing PX versions.

vilasd: force_unsupported_fs_type should only be done for 2.0.x jobs for 2.1.x jobs

mborodin-px:
I tested this change on:
* all basic tests: https://jenkins.portworx.co/job/maxb/job/maxb-temp/15/console
* upgrade job with 1.7.7,2.1.5,2.2.0,2.3.0,wipe https://jenkins.portworx.co/job/maxb/job/maxb-temp2/13/console

PR #280, opened 2019-06-28
PTX-1563 fix logic to avoid getting more nodes to be down than available

thiguetta:


tests/basic/basic_test.go
```diff
@@ -295,9 +298,6 @@ func getMaxNodesToBeDown(replicas int) int {
 	if replicas == 1 {
 		return 0
 	}
-	if replicas%2 != 0 {
-		return replicas/2 + 1
-	}
 	return replicas / 2
```
harsh-px: if number of replicas is 3, it is okay to shutdown 2 PX nodes in a 5 node cluster. With this change, it will only shutdown 1 node in a 5 node cluster.

tests/basic/basic_test.go
```diff
@@ -291,13 +294,14 @@ var _ = Describe("{AppTasksDown}", func() {
 	})
 })
 
-func getMaxNodesToBeDown(replicas int) int {
+func getMaxNodesToBeDown(nodes, replicas int) int {
 	if replicas == 1 {
 		return 0
 	}
-	if replicas%2 != 0 {
+	if nodes > 3 && replicas%2 != 0 {
```
harsh-px: this will shutdown 2 nodes in a 4 node cluster for a repl 3 volume and hence will bring PX out of quorum.

Also add a comment on `getMaxNodesToBeDown` on the purpose of the function.

PR #282, opened 2019-07-02
PTX-1564 fix problem when torpedo failed to determine valid postgres pod status

thiguetta:


thiguetta:
tests passing https://jenkins.portworx.co/job/Torpedo/view/Torpedo%202.1/job/tp-2-1-reboot/70/

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -335,6 +335,10 @@ func isMountReadOnly(mount string) bool {
 	return re.MatchString(mount)
 }
 
+func isErrPodNotFound(err error) bool {
```
harsh-px: The `container not found` check is not the right thing to do based on the method name. `isErrPodNotFound` should return only if the pod is actually not present. 

`container not found` means that a particular container in a pod is not found. The pod itself is present.

thiguetta: yeah container not found happens when a pod is in a not ready state

harsh-px: Exactly. `Completed` is not the same as `isErrPodNotFound`. The helper function name can mislead future users of the function.

thiguetta:
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%202.1/job/tp-2-1-reboot/86 passed

PR #286, opened 2019-07-11
PTX-1637 Add retry for getting attached volumes

nikolaypopov:
Add retry for getting attached volumes

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -622,7 +622,7 @@ func (d *portworx) GetNodeForVolume(vol *torpedovolume.Volume) (*node.Node, erro
 	t := func() (interface{}, bool, error) {
 		vols, err := d.getVolDriver().Inspect([]string{name})
 		if err != nil {
-			logrus.Warnf("failed to inspect volume: %s due to: %v", name, err)
+			logrus.Warnf("Failed to inspect volume: %s due to: %v", name, err)
 			return nil, true, err
 		}
 		if len(vols) != 1 {
@@ -630,7 +630,6 @@ func (d *portworx) GetNodeForVolume(vol *torpedovolume.Volume) (*node.Node, erro
 			logrus.Warnf(err.Error())
 			return nil, true, err
 		}
-
 		return vols[0], false, nil
 	}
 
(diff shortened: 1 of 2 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -642,22 +641,40 @@ func (d *portworx) GetNodeForVolume(vol *torpedovolume.Volume) (*node.Node, erro
...
 
-	return nil, &ErrFailedToInspectVolume{
-		ID:    name,
-		Cause: "Volume is not attached on any node",
+	if n != nil {
+		node := n.(*node.Node)
+		return node, nil
 	}
+
+	return nil, nil
```
harsh-px: the default return should return an error just like it was doing before. An API should not be returning a nil err and also a nil return object. The user of the function wouldn't know what to do with it.

drivers/volume/portworx/portworx.go
```diff
@@ -642,22 +641,40 @@ func (d *portworx) GetNodeForVolume(vol *torpedovolume.Volume) (*node.Node, erro
...
+		pxVol := v.(*api.Volume)
+		for _, n := range node.GetStorageDriverNodes() {
+			if n.VolDriverNodeID == pxVol.AttachedOn {
+				return &n, false, nil
+			}
+		}
+
+		// Snapshots may not be attached to a node
+		if pxVol.Source.Parent != "" {
+			logrus.Debugf("Volume with ID: %s is a snapshot", pxVol.Id)
```
harsh-px: remove the debugf. Not needed.

drivers/volume/portworx/portworx.go
```diff
@@ -642,22 +641,40 @@ func (d *portworx) GetNodeForVolume(vol *torpedovolume.Volume) (*node.Node, erro
...
+			return nil, false, nil
+		}
+
+		return nil, true, &ErrFailedToInspectVolume{
```
Rohit-PX: need to define a new error called `ErrFailedToAttachVolume` might be confusing for new users if we print the failure to be for inspect.

drivers/volume/portworx/portworx.go
```diff
@@ -642,22 +641,40 @@ func (d *portworx) GetNodeForVolume(vol *torpedovolume.Volume) (*node.Node, erro
...
+			ID:    name,
+			Cause: "Volume is not attached on any node",
 		}
 	}
 
-	// Snapshots may not be attached to a node
-	if pxVol.Source.Parent != "" {
-		return nil, nil
+	n, err := task.DoRetryWithTimeout(r, defaultTimeout, defaultRetryInterval)
```
harsh-px: Use a 30 second timeout instead of the default 2 minutes.

drivers/volume/portworx/portworx.go
harsh-px: discussed offline with nikolay. The current implementation is correct.

drivers/volume/portworx/portworx.go
nikolaypopov: Removed.

drivers/volume/portworx/portworx.go
nikolaypopov: Made custom timeout vars for validating volume attached

drivers/volume/portworx/portworx.go
nikolaypopov: Made custom err for ErrFailedToValidateAttachment

drivers/volume/portworx/portworx.go
```diff
@@ -642,22 +643,38 @@ func (d *portworx) GetNodeForVolume(vol *torpedovolume.Volume) (*node.Node, erro
...
+			}
+		}
+
+		// Snapshots may not be attached to a node
+		if pxVol.Source.Parent != "" {
+			return nil, false, nil
 		}
+
+		err := fmt.Errorf("Volume: %s is not attached on any node", name)
+		logrus.Warnf(err.Error())
```
harsh-px: The `logrus.Warnf(err.Error())` is not needed. The task.DoRetryWithTimeout should print the error.

nikolaypopov: Removed logrus.Warnf() print and just returning fmt.Errorf()

PR #297, opened 2019-07-25
Fix restore validation api

ram-infrac:
- corrected alert msg for cloudsnap restore 
  - use correct name for cloud restore objects

Signed-off-by: Ram <ram.suradkar@portworx.com>

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2169,12 +2169,8 @@ func (k *K8s) ValidateVolumeSnapshotRestore(ctx *scheduler.Context, timeStart ti
 		if err != nil {
 			return fmt.Errorf("snapshot: %s is not complete. %v", snapshotData.Metadata.Name, err)
 		}
-		snapID := snapshotData.Spec.PortworxSnapshot.SnapshotID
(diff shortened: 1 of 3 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -2169,8 +2169,12 @@ func (k *K8s) ValidateVolumeSnapshotRestore(ctx *scheduler.Context, timeStart ti
 		if err != nil {
 			return fmt.Errorf("snapshot: %s is not complete. %v", snapshotData.Metadata.Name, err)
 		}
+		snapID := snapshotData.Spec.PortworxSnapshot.SnapshotID
+		if snapshotData.Spec.PortworxSnapshot.SnapshotType == snap_v1.PortworxSnapshotTypeCloud {
+			snapID = "in-place-restore-" + vol
+		}
 		// validate each snap restore
-		if err := driver.ValidateVolumeSnapshotRestore(vol, snapshotData.Spec.PortworxSnapshot.SnapshotID, timeStart); err != nil {
+		if err := driver.ValidateVolumeSnapshotRestore(vol, snapID, timeStart); err != nil {
```
disrani-px: Pass in `snapshotData` instead of `snapID` here. This file shouldn't know if it was a Portworx volume

ram-infrac: moved to `volume/portworx`

PR #298, opened 2019-07-26
PTX-1675 Fix testresults folder permission issues in OpenShift cluster

nikolaypopov:
On OpenShift cluster nodes you cannot mount /testresults and write to it. So to have work around we are mounting /mnt/testresults and giving it permissions to be able to write to it.


harsh-px:
@vinayakshnd and @thiguetta please review this.

deployments/deploy-ssh.sh
```diff
@@ -201,6 +201,13 @@ spec:
           - key: ${K8S_VENDOR_KEY}
             operator: ${K8S_VENDOR_OPERATOR}
             ${K8S_VENDOR_VALUE}
+  initContainers:
+  - name: init-sysctl
+    image: busybox
+    imagePullPolicy: IfNotPresent
+    securityContext:
+      privileged: true
+    command: ["sh", "-c", "mkdir -p /mnt/testresults && chmod 777 /mnt/testresults/"]
```
thiguetta: @nikolaypopov did you test it in tp-dev-test job? normally to change a permission on a folder you need to mount the volume in initContainer as well

nikolaypopov: I ran this change in https://jenkins.portworx.co/job/Torpedo/job/tp-openshift4-op-2.1.2 seems to work.

PR #300, opened 2019-07-30
PTX-1714 add nginx sharedv4 longevity test

thiguetta:


drivers/scheduler/k8s/specs/nginx-sharedv4-longevity/px-nginx-app.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - protocol: TCP
+      port: 80
+      targetPort: 80
+---
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: nginx
+spec:
+  replicas: 24
```
harsh-px: instead of hardcoding, let's find a way to parameterize this. Call the parameter as application_replicas_per_namespace. In the helm of the param, mention that this is honored only by statefulsets and apps using read write many PVCs.

Once it's parameterized, remove this new spec and the existing spec itself should be enough.

drivers/scheduler/scheduler.go
```diff
@@ -49,6 +49,8 @@ type ScheduleOptions struct {
 	Nodes []node.Node
 	// StorageProvisioner identifies what storage provider should be used
 	StorageProvisioner string
+	//DeploymentReplicas specific parameter when we need change statefulset replicas
```
nikolaypopov: specific parameter when we need **to** change statefulset replicas

thiguetta: renamed to AppReplicas since it can be used to deployments or statefulsets

thiguetta:
closing this PR since it's being redesigned

PR #302, opened 2019-07-31
Implement modifying security context constraints when scheduling apps for OC driver

nikolaypopov:
Adding Schedule func() to OC driver to be able to use security context constraints APIs to overwrite permissions to allow privileged container creation.

### DIFF
--- a/drivers/scheduler/openshift/openshift.go
+++ b/drivers/scheduler/openshift/openshift.go
@@ -1,9 +1,11 @@
 package openshift
 
 import (
+	k8s_ops "github.com/portworx/sched-ops/k8s"
 	"github.com/portworx/torpedo/drivers/node"
 	"github.com/portworx/torpedo/drivers/scheduler"
 	kube "github.com/portworx/torpedo/drivers/scheduler/k8s"
+	"github.com/portworx/torpedo/drivers/scheduler/spec"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


drivers/scheduler/openshift/openshift.go
```diff
@@ -57,6 +59,70 @@ func (k *openshift) StartSchedOnNode(n node.Node) error {
...
+		}
+
+		contexts = append(contexts, ctx)
+	}
+
+	return contexts, nil
+}
+
+func (k *openshift) updateSecurityContextConstraints(k8sOps k8s_ops.Ops, namespace string) error {
+	// Get priviledged context
```
piyush-nimbalkar: nit: *privileged

drivers/scheduler/openshift/openshift.go
```diff
@@ -57,6 +59,70 @@ func (k *openshift) StartSchedOnNode(n node.Node) error {
...
+			},
+		}
+
+		contexts = append(contexts, ctx)
+	}
+
+	return contexts, nil
+}
+
+func (k *openshift) updateSecurityContextConstraints(k8sOps k8s_ops.Ops, namespace string) error {
```
piyush-nimbalkar: No need to pass `k8sOps`. We could just do `k8s_ops.Instance().GetSCC()` as this is a singleton and we are not creating an object everytime.

drivers/scheduler/openshift/openshift.go
```diff
@@ -57,6 +59,70 @@ func (k *openshift) StartSchedOnNode(n node.Node) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Add user and namespace to context
+	context.Users = append(context.Users, "system:serviceaccount:"+namespace+":default")
+
+	// Update context
+	_, err = k8sOps.UpdateSecurityContextConstraints(context)
+	if err != nil {
```
piyush-nimbalkar: nit: We can simplify this by just 
`return err`
If it's nil then we will return nil else error.

nikolaypopov: I'm going to leave this be

drivers/scheduler/openshift/openshift.go
nikolaypopov: Fixed, thanks

drivers/scheduler/openshift/openshift.go
nikolaypopov: Done.

PR #305, opened 2019-08-02
Removing echo t > /proc/sysrq-trigger command from log collection

nikolaypopov:
Removing echo t > /proc/sysrq-trigger command from log collection, because it freezes instances and causes timeouts.


vilasd:
Put a comment on top the line to add the echo t later wherever necessary.

tests/common.go
```diff
@@ -317,7 +317,7 @@ func CollectSupport() {
 			expect(nodes).NotTo(beEmpty())
 
 			journalCmd := fmt.Sprintf(
-				"echo t > /proc/sysrq-trigger && journalctl -l > ~/all_journal_%v",
+				"journalctl -l > ~/all_journal_%v",
```
harsh-px: remove the journal too.. not needed.

nikolaypopov: @harsh-px If we remove journalctl command as well there is no need for this function at all then, should we just disable this func?

nikolaypopov: Removed CollectSupport()

PR #306, opened 2019-08-05
 Made wait time for auto recovery of storage nodes configurable

vinayakshnd:


vinayakshnd:
@harsh-px  in general, its good to have auto-storage nodes recovery configurable. Will also be helpful till we fix PWX-9599

vinayakshnd:
@harsh-px addressed review comments

PR #308, opened 2019-08-08
PTX-1755 check whether AttachedOn is a node-id or an ip

thiguetta:


thiguetta:
tests succeeds https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/55/console

PR #312, opened 2019-08-09
PTX-1644 add support to get kvdb members and mark nodes as metadata

thiguetta:


drivers/volume/portworx/portworx.go
```diff
@@ -181,6 +193,25 @@ func (d *portworx) updateNode(n node.Node, pxNodes []api.Node) error {
 	return fmt.Errorf("failed to find px node for node: %v PX nodes: %v", n, pxNodes)
 }
 
+func (d *portworx) isMetadataNode(node node.Node, address string) bool {
+	members, err := d.getKvdbMembers(node)
+	ipRegex := regexp.MustCompile(`http:\/\/(?P<address>.*)\:\d+`)
+	if err != nil {
+		logrus.Errorf("Failed to get metadata nodes. Cause: %v", err)
+		return false
```
Rohit-PX: This means we are essentially saying it's not a metadata node, could this be a false negative?
The actual error is what line 200 says which is `logrus.Errorf("Failed to get metadata nodes. Cause: %v", err)`

harsh-px: we should not be ignoring the error here. failure to determine it's a kvdb node does not mean the node is not a metadata node.

drivers/volume/portworx/portworx.go
```diff
@@ -171,7 +181,9 @@ func (d *portworx) updateNode(n node.Node, pxNodes []api.Node) error {
 			if address == pxNode.DataIp || address == pxNode.MgmtIp || n.Name == pxNode.Hostname {
 				n.VolDriverNodeID = pxNode.Id
 				n.IsStorageDriverInstalled = isPX
+				n.IsMetadataNode = d.isMetadataNode(n, address)
 				node.UpdateNode(n)
+				logrus.Debugf("Node updated: %v", n)
```
harsh-px: remove the debug log

PR #316, opened 2019-08-12
PTX-1781 Adding ability to set log level, default is debug

nikolaypopov:
Adding ability to set log level, default is debug

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -438,14 +438,12 @@ func ParseFlags() {
 		})
 	}
 
-	// Set logging
+	// Set log level
 	logLvl, err := logrus.ParseLevel(instance.LogLevel)
 	if err != nil {
 		logrus.Fatalf("Failed to set log level due to Err: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


tests/common.go
```diff
@@ -432,6 +437,15 @@ func ParseFlags() {
...
 		})
 	}
+
+	// Set logging
+	logLvl, err := logrus.ParseLevel(instance.LogLevel)
+	if err != nil {
+		logrus.Fatalf("Failed to set log level due to Err: %v", err)
+	}
+	logrus.SetLevel(logLvl)
+	logrus.StandardLogger().Hooks.Add(log.NewHook())
```
harsh-px: you don't need to do this again here. It's already being done in init(). Just set the log level here.

harsh-px: Also did you test if this actually works? i.e invoke with different log levels in a private job.

nikolaypopov: Yes I have tested this with a private image, it works.

PR #319, opened 2019-08-14
PTX-1312 fix Panic due to type assertion after k8s API timeout

thiguetta:


drivers/scheduler/k8s/k8s.go
```diff
@@ -1091,10 +1089,8 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
 			}
 			return nil, false, nil
 		}
-		pods, err = task.DoRetryWithTimeout(t, k8sDestroyTimeout, DefaultRetryInterval)
-		if err != nil {
-			podList = append(podList, pods.(v1.Pod))
-		}
+
+		task.DoRetryWithTimeout(t, k8sDestroyTimeout, DefaultRetryInterval)
```
harsh-px: we cannot ignore the err here.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1104,10 +1100,8 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
...
```
harsh-px: we cannot ignore the err here.

harsh-px: same for other places in PR.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1064,9 +1062,17 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
...
-			podList = append(podList, pods.(v1.Pod))
+			// in case we're not waiting for resource cleanup
+			if value, ok := opts[scheduler.OptionsWaitForResourceLeakCleanup]; !ok || !value {
+				return err
+			}
+			if pods != nil {
+				podList = append(podList, pods.([]v1.Pod)...)
+			}
+			// we're ignoring this error since we want to verify cleanup down below, so simply logging it
+			logrus.Errorf("Failed to destroy core objects. Cause: %v", err)
```
harsh-px: if this is an error we are ignoring, use `logrus.Warnf`.

thiguetta: fixed

thiguetta:
jobs passing https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/111/

PR #320, opened 2019-08-15
 Update rule namespace only if empty

disrani-px:
Don't need to validate destroy of rule


drivers/scheduler/k8s/k8s.go
```diff
@@ -1214,23 +1216,6 @@ func (k *K8s) WaitForDestroy(ctx *scheduler.Context, timeout time.Duration) erro
 			}
 
 			logrus.Infof("[%v] Validated destroy of Service: %v", ctx.App.Key, obj.Name)
-		} else if obj, ok := spec.(*stork_api.Rule); ok {
```
Rohit-PX: We would still need this to verify if rule exists?

disrani-px: This is for destroying, it is only helpful for objects like deployments and stateful sets which need pods to be deleted

Rohit-PX: ok that works then as long as we are not dependent on the rule getting deleted.

PR #321, opened 2019-08-15
Vendor update for external-storage

disrani-px:


disrani-px:
Yes, some tests uses the same rule multiple times in different namespaces, which was causing an issue since it updates it to the first namespace and then doesn't update for other namespaces

PR #324, opened 2019-08-19
PTX-1790 add support to install flush cache service

thiguetta:


thiguetta:
tests passing https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/90/console

drivers/node/ssh/ssh.go
```diff
@@ -581,13 +579,11 @@ EOF`
 		logrus.Debugf("flush_cache already started on node %s", n.Name)
 	}
 
-	out, err = s.Systemctl(n, "flush_cache.service", node.SystemctlOpts{
+	//in this case we are ignoring error cause is-enabled can return unknown exited 3 in case service is not found
```
thiguetta: just in case you're wondering what happens if we ignore, there are proper debug logs inside `Systemctl ` if the error is somehow different than expected

PR #328, opened 2019-08-20
PTX-1808 Imlementing collecting diags when failing to wait for PX to come up

nikolaypopov:
Imlementing collecting diags when failing to wait for PX to come up

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -33,6 +33,8 @@ const (
 	DriverName              = "pxd"
 	pxdClientSchedUserAgent = "pxd-sched"
 	pxdRestPort             = 9001
+	pxDiagPort              = 9014
+	pxDiagPath              = "/remotediags"
 	pxVersionLabel          = "PX Version"
 	maintenanceOpRetries    = 3
 	enterMaintenancePath    = "/entermaintenance"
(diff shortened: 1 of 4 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1517,6 +1550,40 @@ func (d *portworx) getKvdbMembers(n node.Node) (map[string]metadataNode, error)
...
 }
 
+func (d *portworx) CollectDiags(n node.Node) error {
+	var err error
+
+	pxNode, err := d.getPxNode(n, nil)
+	if err != nil {
+		return err
+	}
+	if pxNode.Status == api.Status_STATUS_OFFLINE {
```
harsh-px: @vilasd when PX is offline, shouldn't we still take diags? We won't be able to take live cores but we can atleast take the offline diags.

vilasd: Yes! put else condition to collect regular diags

nikolaypopov: As per conversation with @jrivera-px, if the container is not running 9014 won't be there so you won't be able to issue the request. They have same check in Porx for collection diags.

jrivera-px: The diags server runs inside the PX container.   So if the PX container is down/not running you won't be able to issue a diags via an http request.   In that case you should take the diags from the host which px is running on.

nikolaypopov: Ok I will use pxctl sv diags to collect it incase of PX is offline.

drivers/volume/portworx/portworx.go
```diff
@@ -1517,6 +1550,60 @@ func (d *portworx) getKvdbMembers(n node.Node) (map[string]metadataNode, error)
...
+		// Only way to collect diags when PX is offline is using pxctl
+		out, err := d.nodeDriver.RunCommand(n, "/opt/pwx/bin/pxctl sv diags -a -f", opts)
+		if err != nil {
+			return fmt.Errorf("Failed to collect diags on node %v, Err: %v %v", pxNode.Hostname, err, out)
+		}
+		logrus.Debugf("Successfully collected diags on node %v", pxNode.Hostname)
+		return nil
+	}
+
+	url := fmt.Sprintf("http://%s:9014", n.Addresses[0])
```
mborodin-px: it should be in the else statement, otherwise if pxNode.Status == api.Status_STATUS_OFFLINE then it will try to collect diags from http request.

tests/common.go
```diff
@@ -249,6 +249,11 @@ func StartVolDriverAndWait(appNodes []node.Node) {
 		Step(fmt.Sprintf("wait for volume driver to start on nodes: %v", appNodes), func() {
 			for _, n := range appNodes {
 				err := Inst().V.WaitDriverUpOnNode(n, Inst().DriverStartTimeout)
+				if err != nil {
```
mborodin-px: This is the candidate to have a function.

drivers/volume/portworx/portworx.go
nikolaypopov: It will return after successfully collecting diags and will not try continue down to use http to collect diags as well

tests/common.go
nikolaypopov: Removed unnecessary debug message, no need to make a func() out of it

PR #335, opened 2019-09-03
[PTX-1840] Send kill -SIGUSR1 <px_process_id> while collecting diags

vinayakshnd:


drivers/volume/portworx/portworx.go
```diff
@@ -1572,18 +1572,29 @@ func (d *portworx) CollectDiags(n node.Node) error {
 		return err
 	}
 
+	opts := node.ConnectionOpts{
```
nikolaypopov: Why did you remove Sudo: true from here, it was required for collecting diags on some of the systems. It was added in PTX-1825

vinayakshnd: @nikolaypopov , Added "Sudo: true" back.

vinayakshnd:
@nikolaypopov  addressed review comments

PR #338, opened 2019-09-16
Check if config map is empty before retrieving secrets.

Rohit-PX:
Why do we need this PR?
If config map is not being created in the job, retrieving secrets should skipped.

Failed here:
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%202.2/job/tp-2-2-sysbench/68/console


Rohit-PX:
Testing here: https://jenkins.portworx.co/job/Torpedo/view/Torpedo%202.2/job/tp-2-2-sysbench/69/console

PR #341, opened 2019-09-18
[DO-NOT-MERGE][PTX-1854] Fix kill -SIGUSER1 command during diags collection

vinayakshnd:


sangleganesh:
if you are going to collect diags with live == true then you don't need to send the signal again.

vinayakshnd:
Closing this PR in favor of #345

PR #345, opened 2019-09-20
Generate live diags

vinayakshnd:


drivers/volume/portworx/portworx.go
```diff
@@ -1642,26 +1642,13 @@ func (d *portworx) CollectDiags(n node.Node) error {
...
-	logrus.Debugf("Successfully sent SIGUSR1 signal to px process [%s] on [%v]", pxPid, pxNode.Hostname)
-
 	logrus.Debugf("Collecting diags on node %v, because there was an error", pxNode.Hostname)
 
 	if pxNode.Status == api.Status_STATUS_OFFLINE {
 		logrus.Debugf("Node %v is offline, collecting diags using pxctl", pxNode.Hostname)
 
 		// Only way to collect diags when PX is offline is using pxctl
-		out, err := d.nodeDriver.RunCommand(n, "pxctl sv diags -a -f", opts)
+		out, err := d.nodeDriver.RunCommand(n, "pxctl sv diags -a -f -l", opts)
```
nikolaypopov: I don't think there is a point to collect live diags when PX is down.

nikolaypopov:
Would probably a good idea to rename this issue to something like "removing SIGUSR signal code", was confusing at first. Just a thought :)

harsh-px:
@thiguetta  to include live diags to collect as part of #378

PR #346, opened 2019-09-20
PTX-1564 fix validation when deployment is using non-shared volumes

thiguetta:
this PR fixes the problem when validating mounts in pods when we have postgres which is expected to have one container in Creating/CrashLoopBackoff/Completed state


thiguetta:
do not merge it yet I'm running the tests in dev-test as discussed with @harsh-px as it is a recurring bug, I'm running the tests over and over at least 20 times

nikolaypopov:
Looks good to me, more simple logic with less code. Have you tested it?

thiguetta:
as mentioned I ran 20 times the tests https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/ if you check builds from 139 to 167 you'll see 20 successful runs, you can ignore the failed jobs which are not related to the same problem, although feel free to check those logs as well

PR #347, opened 2019-09-20
AUT-16 add autopilot test for pool expansion

thiguetta:


drivers/scheduler/k8s/k8s.go
```diff
@@ -200,7 +206,28 @@ func (k *K8s) ParseSpecs(specDir string) ([]interface{}, error) {
...
+		}
+
+		// in case we're not able to execute the file, warn the user and fallback to file reading
+		if err = specTemplate.Execute(&buff, k.ExternalOpts); err != nil {
+			logrus.Warnf("failed to execute template. Cause: %v", err)
+			ioReader = file
+		}
+
+		reader := bufio.NewReader(ioReader)
 		specReader := yaml.NewYAMLReader(reader)
```
mborodin-px: I would prefer to have a file with all variables related to specs and then parse them and substitute to appropriate values rather than add them to structs.

deployments/deploy-ssh.sh
```diff
@@ -157,6 +157,13 @@ else
     K8S_VENDOR_KEY=node-role.kubernetes.io/master
 fi
 
+if [ -z "${PGBENCH_STORAGE_SIZE}" ]; then
+	PGBENCH_STORAGE_SIZE="--pgbench-storage-size=$PGBENCH_STORAGE_SIZE"
+fi
+
+if [ -z "${PGBENCH_DATA_SIZE}" ]; then
+	PGBENCH_DATA_SIZE="--pgbench-data-size=$PGBENCH_DATA_SIZE"
+fi
```
mborodin-px: see below comment regarding to ave a file with variables

drivers/volume/portworx/portworx.go
```diff
@@ -1684,6 +1722,94 @@ func (d *portworx) CollectDiags(n node.Node) error {
...
+	if errors.IsAlreadyExists(err) {
+		logrus.Infof("Found existing AutopilotRule: %v", autopilotRule.Name)
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to create AutopilotRule: %s. Err: %v", autopilotRule.Name, err)
+	}
+
+	return nil
+}
+
```
mborodin-px: move this logic to driver/scheduler/k8s/k8s.go, the same as I did for for VolumeResize. Looks like you didn't apply my changes from AUT-13 as I already have create createAutopilotRule function there.

drivers/volume/portworx/portworx.go
```diff
@@ -1684,6 +1722,94 @@ func (d *portworx) CollectDiags(n node.Node) error {
...
+func (d *portworx) createAutopilotPoolExpandObject(apParams *torpedovolume.AutopilotRuleParams) (*ap_api.AutopilotRule, error) {
+	obj := &ap_api.AutopilotRule{}
+	obj.Name = apParams.Name
+	obj.Spec.Enforcement = ap_api.EnforcementRequired
+
+	actions := &ap_api.RuleAction{
+		Name: "penstorage.io.action.storagepool/expand",
```
mborodin-px: I think it should be openstorage instead of penstorage

drivers/volume/portworx/portworx.go
thiguetta: in this case this is not related to scheduler but the volume driver, @harsh-px found better put under volume driver

mborodin-px: In line 1776 fix typo penstorage, should be openstorage

tests/autopilot_pool_expand/autopilot_pool_expand_test.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		Step("validating pool and verifying size of it", func() {
+			err = Inst().V.ValidateStorage(timeout, retryInterval)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		opts := make(map[string]bool)
+		opts[scheduler.OptionsWaitForResourceLeakCleanup] = true
+		ValidateAndDestroy(contexts, opts)
```
nikolaypopov: Why are we not checking for errors here?

harsh-px:
This can be closed as @mborodin-px is working on this in a different branch now

PR #349, opened 2019-09-25
PTX-1907  [torpedo] Torpedo failing to validate host on openshift 4.2…

nareshdesh:
modified code to use 'date'


nareshdesh:
Following torpedo job is using the code changes made in this request
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%202.2/job/tp-openshift4.2-op-2.2-csi/

drivers/node/ssh/ssh.go
```diff
@@ -195,7 +195,7 @@ func (s *SSH) TestConnection(n node.Node, options node.ConnectionOpts) error {
 	if useSSH() {
 		cmd = "hostname"
 	} else {
-		cmd = "cat /etc/hostname"
+		cmd = "date"
```
harsh-px: this is okay for now but we could just use the same hostname command for both. Once we are done with 2.2, you can send another PR to use `cmd = "hostname"` both both. (i.e remove the if condition)

thiguetta:
@nareshdesh could you please run a test on openshift 3.11 before merging, I remember it has some limited command set

harsh-px:
@nareshdesh as thiago requested, let's run this PR on a openshift 4.2 job an merge after it passes.

PR #352, opened 2019-09-30
Changed timeout values for pgbench workload and small fix in Autopilot test

mborodin-px:
Added pgbench workload timeout, as it fails when you run with higher SCALE_FACTOR number.

### DIFF
--- a/drivers/scheduler/k8s/specs/pgbench/px-pgbench-storage.yaml
+++ b/drivers/scheduler/k8s/specs/pgbench/px-pgbench-storage.yaml
@@ -4,9 +4,8 @@ metadata:
   name: postgres-pgbench-sc
 provisioner: kubernetes.io/portworx-volume
 parameters:
-  repl: "3"
-  priority_io: "high"
-  io_profile: "db"
+  repl: "1"
+  priority_io: "low"
 allowVolumeExpansion: true
 ---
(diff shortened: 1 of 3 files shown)


drivers/scheduler/k8s/specs/pgbench/px-pgbench-storage.yaml
```diff
@@ -4,9 +4,8 @@ metadata:
   name: postgres-pgbench-sc
 provisioner: kubernetes.io/portworx-volume
 parameters:
-  repl: "3"
-  priority_io: "high"
-  io_profile: "db"
+  repl: "1"
```
thiguetta: why using profile low? since pgbench is database, and why so little replicas?

mborodin-px: If you try to create a PVC with repl: 1 and priority_io: high you will get "Failed to provision volume with StorageClass "postgres-pgbench-sc": rpc error: code = Internal desc = Failed to create volume: This IO profile requires a minimum replication factor of 3 nodes". That is why I changed it to low.

thiguetta: we can assume we'll have always 4 workers nodes in our tests

PR #354, opened 2019-10-01
For auth clusters pass tokens to Update HA and Resize Vol.

Rohit-PX:
* We need to pass the security tokens when updating HA for volumes and also for resizing volumes.
* The resizing volumes won't work yet because of:
   https://portworx.atlassian.net/browse/PWX-10100


drivers/scheduler/k8s/k8s.go
```diff
@@ -1805,10 +1805,26 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
 
 //ResizeVolume  Resize the volume
-func (k *K8s) ResizeVolume(ctx *scheduler.Context) ([]*volume.Volume, error) {
+func (k *K8s) ResizeVolume(ctx *scheduler.Context, configMapName string) ([]*volume.Volume, error) {
 	k8sOps := k8s_ops.Instance()
 	var vols []*volume.Volume
 	for _, spec := range ctx.App.SpecList {
+		// Add security annotations if running with auth-enabled
+		if configMapName != "" {
+			configMap, err := k8sOps.GetConfigMap(configMapName, "default")
```
thiguetta: are you sure namespace will be always `default`?

Rohit-PX: Yes we create the config maps in the jenkins job and don't want the users to need to specify the namespace.

drivers/scheduler/dcos/dcos.go
```diff
@@ -414,7 +414,7 @@ func (d *dcos) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 	}
 }
 
-func (d *dcos) ResizeVolume(cxt *scheduler.Context) ([]*volume.Volume, error) {
```
stgleb: Is it configMap content or configMap name?

PR #357, opened 2019-10-11
Update latest sched-ops and use autopilot-api package instead of autopilot

harsh-px:


piyush-nimbalkar:
If not needed try removing the override block for azure assuming it's not a direct dependency and due to sched-ops dependency the right version may get pulled anyway. 
Does the same apply for k8s packages?
This will probably make `dep ensure` better in stork which depends on torpedo.

harsh-px:
Tested image on torpedo dev job. merging now. https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/172/console

PR #358, opened 2019-10-16
PTX-1940 Add timeout parameter for upgrade script

mborodin-px:

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -83,6 +83,7 @@ const (
 	startDriverTimeout               = 2 * time.Minute
 	upgradeTimeout                   = 10 * time.Minute
 	upgradeRetryInterval             = 30 * time.Second
+	upgradePerNodeTimeout            = 15 * time.Minute
 	waitVolDriverToCrash             = 1 * time.Minute
 )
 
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -83,6 +83,7 @@ const (
 	startDriverTimeout               = 2 * time.Minute
 	upgradeTimeout                   = 10 * time.Minute
 	upgradeRetryInterval             = 30 * time.Second
+	upgradePerNodeTimeout            = "900"
```
thiguetta: a small cosmetic suggestion, can we use the `time.Duration` notation as we do above and then use `duration.Seconds()` in your command?

drivers/volume/portworx/portworx.go
```diff
@@ -356,6 +357,29 @@ func (d *portworx) getPxNode(n node.Node, cManager cluster.Cluster) (api.Node, e
...
 }
 
+func (d *portworx) getPxVersionOnNode(n node.Node, cManager cluster.Cluster) (string, error) {
+
+	pxNode, err := d.getPxNode(n, cManager)
+	if err != nil {
+		return "", err
+	}
+
+	t := func() (interface{}, bool, error) {
```
harsh-px: this timeout look doesn't make sense.

In each retry, you are using the same pxNode that you got from a single attempt above. So nothing is going to change between subsequent retries.

drivers/volume/portworx/portworx.go
mborodin-px: fixed

drivers/volume/portworx/portworx.go
mborodin-px: fixed

drivers/volume/portworx/portworx.go
```diff
@@ -1280,9 +1294,21 @@ func (d *portworx) UpgradeDriver(endpointURL string, endpointVersion string) err
 	}
 	logrus.Infof("permission changed on file %s", upgradeFileName)
 
+	nodeList := node.GetStorageDriverNodes()
+	pxNode := nodeList[0]
+	pxVersion, err := d.getPxVersionOnNode(pxNode, nil)
```
harsh-px: you are ignoring the `err` here

mborodin-px: fixed

drivers/volume/portworx/portworx.go
```diff
@@ -356,6 +357,28 @@ func (d *portworx) getPxNode(n node.Node, cManager cluster.Cluster) (api.Node, e
 	return pxnode.(api.Node), nil
 }
 
+func (d *portworx) getPxVersionOnNode(n node.Node, cManager cluster.Cluster) (string, error) {
+
+	t := func() (interface{}, bool, error) {
+		logrus.Debugf("Getting PX Version on node [%s]", n.Name)
+		pxNode, err := d.getPxNode(n, cManager)
+		if err != nil {
+			return "", err
```
harsh-px: don't think this compiles

PR #360, opened 2019-10-16
PTX-1563 fix nodes available to be down

thiguetta:


tests/basic/basic_test.go
```diff
@@ -52,19 +52,21 @@ var _ = Describe("{VolumeDriverDown}", func() {
...
-				})
+				Step(
+					fmt.Sprintf("starting volume %s driver on node %s",
+						Inst().V.String(), appNode.Name),
+					func() {
+						StartVolDriverAndWait([]node.Node{appNode})
+					})
 
 				Step("Giving few seconds for volume driver to stabilize", func() {
 					time.Sleep(20 * time.Second)
```
harsh-px: add validate context after this..

after bouncing PX on each node, we want to ensure all apps are fine.

tests/basic/basic_test.go
```diff
@@ -92,21 +94,29 @@ var _ = Describe("{VolumeDriverDownAttachedNode}", func() {
...
+							StartVolDriverAndWait([]node.Node{appNode})
+						})
 
-				Step("Giving few seconds for volume driver to stabilize", func() {
-					time.Sleep(20 * time.Second)
-				})
+					Step("Giving few seconds for volume driver to stabilize", func() {
+						time.Sleep(20 * time.Second)
+					})
+				}
```
harsh-px: validate `ctx` here. Similar as above test. After bouncing PX on each `appNode`, validate the current ctx.

harsh-px: The test seems be using `ValidateAndDestroy` at below. Instead move the validate up here and just do destroy in the end.

This should help us catch errors much earlier.

we need to run this in a private job as the test behavior has changed.

PR #366, opened 2019-10-28
Add IO load to elastic-search application

saheienko:
Use the [elastic-stress-test](https://github.com/logzio/elasticsearch-stress-test) tool to generate a load for the es cluster.

### DIFF
--- a/drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
+++ b/drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
@@ -14,15 +14,14 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
-  name: elasticsearch-loadbalancer
+  name: elasticsearch-api
 spec:
   selector:
     app: es-cluster
   ports:
   - name: http
-    port: 80
+    port: 9200
     targetPort: 9200
-  type: LoadBalancer
 ---
 apiVersion: v1
 kind: ConfigMap
@@ -44,7 +43,7 @@ kind: StatefulSet
 metadata:
   name: esnode
 spec:
-  serviceName: elasticsearch
+  serviceName: elasticsearch-api
   replicas: 3
   selector:
     matchLabels:
@@ -119,3 +118,35 @@ spec:
       resources:
         requests:
           storage: 5Gi
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: es-load
+  namespace: default
+  labels:
+    app: es-load
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: es-load
+  template:
+    metadata:
+      labels:
+        app: es-load
+    spec:
+      containers:
+      - name: es-load
+        image: portworx/torpedo-esload
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        command:
+          - /bin/bash
+          - -c
+          - "while true; do echo '=========='; python elasticsearch-stress-test.py --es_address esnode-0.elasticsearch-api.$POD_NAMESPACE.svc.cluster.local:9200 --indices 4 --documents 5 --seconds 120 --not-green --clients 1; sleep 5; done"
+      restartPolicy: Always
+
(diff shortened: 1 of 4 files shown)


drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
```diff
@@ -119,3 +131,29 @@ spec:
...
+      labels:
+        app: esrally
+    spec:
+      containers:
+      - name: esrally
+        image: elastic/rally:1.3.0
+        command:
+          - /bin/sh
+          - -c
+          - "while true; do esrally --pipeline=benchmark-only --target-hosts=esnode-0.elasticsearch-cluster.default.svc.cluster.local:9200,esnode-1.elasticsearch-cluster.default.svc.cluster.local:9200,esnode-2.elasticsearch-cluster.default.svc.cluster.local:9200; sleep 5; done"
```
vilasd: We need to parameterize the load and time that it runs for . Can you see how can we do that ?

harsh-px: From torpedo standpoint, it might be okay if it keeps running through the course of the test.

Only concern would be the rate of IO this generates. Do we know the throughput or IOPS the above command would generate?

saheienko: I haven't found any time parameters for track/tests. Try to find an alternative tool for this.

drivers/scheduler/sidecars/es-stress-test/elasticsearch-stress-test.py
```diff
@@ -0,0 +1,462 @@
...
+try:
+    main()
+
+except Exception as e:
+    print("Got unexpected exception. probably a bug, please report it.")
+    print("")
+    print(e.message)
+    print("")
+
+    sys.exit(1)
```
thiguetta: instead of adding this to our repo, can we simple clone this https://github.com/logzio/elasticsearch-stress-test repo inside docker image and run it from there?

drivers/scheduler/sidecars/es-stress-test/requirements.txt
```diff
@@ -0,0 +1 @@
+elasticsearch
```
thiguetta: I guess this is better to put this inside dockerfile

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
```diff
@@ -119,3 +119,30 @@ spec:
...
+    matchLabels:
+      app: es-load
+  template:
+    metadata:
+      labels:
+        app: es-load
+    spec:
+      containers:
+      - name: es-load
+        image: serhiiqbox/px-enterprise:es-stress-test
```
thiguetta: change this to not use your personal image, ask @shailvipx to create a repository like portworx/es-sidecar

drivers/scheduler/sidecars/es-stress-test/elasticsearch-stress-test.py
saheienko: It could be done in that way. The reason for copying these files is to make it persistent (it will be available if the upstream repo is deleted).

thiguetta: all in all the data will be safe inside docker image since we won't generate a new image often

drivers/scheduler/sidecars/es-stress-test/requirements.txt
saheienko: This file is passed as docker build context and a base image use it to download dependencies.

thiguetta: this is part of https://github.com/logzio/elasticsearch-stress-test/blob/master/requirements.txt as well, let's keep it simple

drivers/scheduler/sidecars/es-stress-test/elasticsearch-stress-test.py
saheienko: done

drivers/scheduler/sidecars/es-stress-test/requirements.txt
saheienko: done

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
saheienko: That was used for testing purposes. Updated to `portworx/torpedo-esload` but I can't push the image (have no access permissions).

PR #367, opened 2019-10-29
PTX-2005 fix GetWorkerNodes for rancher

thiguetta:


drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -717,7 +718,7 @@ func getPXNodes(destKubeConfig string) ([]corev1.Node, error) {
 	// get label on node where PX is Enabled
 	for _, node := range nodes.Items {
 		// worker node and px is not disabled
-		if !destClient.IsNodeMaster(node) && node.Labels[PXEnabledLabelKey] != "false" {
+		if !destClient.IsNodeMaster(node) && node.Labels[PXEnabledLabelKey] != "false" && node.Labels[rancherControlPlaneLabelKey] != "true" {
```
harsh-px: instead of doing it like this, create a map of labels above for which PX is not expected to run. And iterate over that map here and check if the node has a label from any entry in that map.

harsh-px: The end effect is the same. But the code will be easier to read and maintain.

PR #368, opened 2019-10-29
Set ConfigMap name when initializing volume driver.

Rohit-PX:
Why do we need this PR?
* Currently, we are passing the config map name as part of `options` params to the `Schedule` method. 
* Instead we will now set the config map name when initializing the `Scheduler` driver.
* That way when using the Schedule method from torpedo, config map name need not be passed and will be retrieved from the Scheduler driver.

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -85,9 +85,10 @@ var (
 
 //K8s  The kubernetes structure
 type K8s struct {
-	SpecFactory    *spec.Factory
-	NodeDriverName string
-	VolDriverName  string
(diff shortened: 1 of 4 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -81,6 +81,8 @@ const (
 
 var (
 	namespaceRegex = regexp.MustCompile("{{NAMESPACE}}")
+	// SecretConfigMap is name of the config map that will store px secrets with auth tokens
+	SecretConfigMap string
```
disrani-px: Doesn't need to be exposed from the package and move it to the K8s struct below.

PR #371, opened 2019-11-06
PTX-1975 Added autopilot test for max resize rule

mborodin-px:


tests/autopilot-capacity/aut_test.go
```diff
@@ -74,8 +75,31 @@ var ruleResizeBy50IfPvcCapacityLessThan10 = scheduler.AutopilotRuleParameters{
 	ExpectedPVCSize: 12079595520,
 }
 
+var ruleResizeBy50UntilPvcMaxSize20 = scheduler.AutopilotRuleParameters{
```
harsh-px: Don't add a new rule test case. Add maxsize in the existing basic test case itself.

harsh-px: Discussed offline. Would be good to have a separate test case.

PR #374, opened 2019-11-07
Increased initialDelaySeconds for sysbench

mborodin-px:
It should help to fix the issue when mysql database initialization is not completed and pod restarts which causing **Access denied for user 'root'@'localhost' (using password: YES).**


thiguetta:
@mborodin-px was it tested anywhere?

mborodin-px:
Yes, all upgrade jobs are green now.

PR #376, opened 2019-11-27
PTX-2076 expose app destroy timeout in deploy-ssh.sh

thiguetta:


deployments/deploy-ssh.sh
```diff
@@ -75,6 +75,11 @@ if [ -z "$DRIVER_START_TIMEOUT" ]; then
     echo "Using default timeout of ${DRIVER_START_TIMEOUT}"
 fi
 
+APP_DESTROY_TIMEOUT_ARG=""
```
vilasd: Whats the default value for APP_DESTROY_TIMEOUT_ARG ?

thiguetta: empty, it will be empty to take the default which is 5m

https://github.com/portworx/torpedo/blob/3fd89a2d1edebad1f8cc81fdc6a2b0cac00461f3/tests/common.go#L419

https://github.com/portworx/torpedo/blob/3fd89a2d1edebad1f8cc81fdc6a2b0cac00461f3/tests/common.go#L75

deployments/deploy-ssh.sh
```diff
@@ -271,7 +276,8 @@ spec:
             "--provisioner", "$PROVISIONER",
             "--config-map", "$CONFIGMAP",
             "$UPGRADE_ENDPOINT_URL_ARG",
-            "$UPGRADE_ENDPOINT_VERSION_ARG" ]
+            "$UPGRADE_ENDPOINT_VERSION_ARG",
+            "$APP_DESTROY_TIMEOUT_ARG" ]
```
mborodin-px: don't use empty string as an argument here, use --destroy-app-timeout=$APP_DESTROY_TIMEOUT

thiguetta:
tested passing here https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/216/console

deployments/deploy-ssh.sh
thiguetta: IMHO we already have a default set in torpedo side, there is no purpose to pass `--destroy-app-timeout=$APP_DESTROY_TIMEOUT` when `APP_DESTROY_TIMEOUT` is not provided

PR #234, opened 2019-02-28
PTX-841 Describe objects in a namespace in case of a failure

hrishipx:
What does this PR do?
- Updates to the ginkgo framework version. 
- Adds functions to each test to describe objects within the namespace. 

Validated it locally on my cluster with runs. 

Need to soak it in via Jenkins on the Torpedo infra.

DO NOT MERGE until the tests pass

PR #378, opened 2019-12-05
PTX-2099 collect support bundle when job fails

thiguetta:


tests/common.go
```diff
@@ -329,6 +329,57 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
...
+			nodes := node.GetWorkerNodes()
+			expect(nodes).NotTo(beEmpty())
+
+			for _, n := range nodes {
+				if !n.IsStorageDriverInstalled {
+					continue
+				}
+
+				// this task may hang so we wrapped in a timeout cmd
+				dumpTasksCmd := "timeout 10s bash -c \"echo t > /proc/sysrq-trigger\""
```
vilasd: There is a possibly that it might hang on cloud instances. Lets make it optional or take it out for now.

tests/common.go
```diff
@@ -329,6 +329,57 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
 	})
 }
 
+// CollectSupport creates a support bundle
+func CollectSupport() {
+	context(fmt.Sprintf("generating support bundle..."), func() {
+		Step(fmt.Sprintf("save all useful logs on each node"), func() {
```
vilasd: We need to collect all the output of the describe on pvc/sc / app

thiguetta:
added the content from #234 to this PR, and closed that

tests/common.go
```diff
@@ -329,6 +329,68 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
...
+				dumpTasksCmd := "timeout 10s bash -c \"echo t > /proc/sysrq-trigger\""
+				logrus.Infof("dump the list of current tasks and their information on %s", n.Name)
+				runCmd(dumpTasksCmd, n)
+
+				journalCmd := fmt.Sprintf("journalctl -l > ~/all_journal_%v.log", time.Now().Format(time.RFC3339))
+				logrus.Infof("saving journal output on %s", n.Name)
+				runCmd(journalCmd, n)
+
+				logrus.Infof("saving portworx journal output on %s", n.Name)
+				runCmd("journalctl -lu portworx* > portworx.log", n)
```
nikolaypopov: Where is this executed on? You cannot write anything in / dir on OpenShift clusters, its gotta go to like /tmp or other available for write dir.

nikolaypopov:
Test this by running it on OpenShift deployment, it might have issues as I mentioned in my comment.

thiguetta:
in this case the test is expected to fail because we need to collect the logs and print namespaces objects as you can see in the console https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/238/console

tests/common.go
thiguetta: @nikolaypopov as we agreed we're saving to /var/cores, which all systems have access to

tests/autopilot-capacity/aut_test.go
```diff
@@ -124,12 +124,13 @@ var _ = BeforeSuite(func() {
...
 // PVC sizes of the volumes
 var _ = Describe(fmt.Sprintf("{%sWaitForWorkload}", testNameSuite), func() {
+	var contexts []*scheduler.Context
+
 	It("has to fill up the volume completely, resize the volume, validate and teardown apps", func() {
 		var err error
 		testName := strings.ToLower(fmt.Sprintf("%sWaitForWorkload", testNameSuite))
 
 		for _, apRule := range autopilotruleBasicTestCases {
-			var contexts []*scheduler.Context
```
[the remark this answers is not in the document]
thiguetta: it will, all in all we need them in case of failure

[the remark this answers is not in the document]
thiguetta: https://github.com/portworx/torpedo/blob/e3e81190a846d5c037389127302befeb92b4e3ca/tests/autopilot-capacity/aut_test.go#L257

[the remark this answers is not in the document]
mborodin-px: that test was designed to loop thru the autopilot rules, schedule, validate and delete context. If you move contexts out of autopilot rule loop, the test will fail on next iteration, as it contains deleted context from previous iteration.

tests/common.go
```diff
@@ -356,6 +348,66 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
...
+				runCmd("journalctl -lu kubelet* > /var/cores/kubelet.log", n)
+
+				logrus.Infof("saving dmesg output on %s", n.Name)
+				runCmd("dmesg -T > /var/cores/dmesg.log", n)
+
+				logrus.Infof("saving disk list on %s", n.Name)
+				runCmd("lsblk > /var/cores/lsblk.log", n)
+
+				logrus.Infof("saving mount list on %s", n.Name)
+				runCmd("cat /proc/mounts > /var/cores/mounts.log", n)
```
mborodin-px: since you are using /var/cores folder name in many places I would create a variable for that

tests/autopilot-capacity/aut_test.go
```diff
@@ -184,18 +185,27 @@ var _ = Describe(fmt.Sprintf("{%sWaitForWorkload}", testNameSuite), func() {
 			})
 		}
 	})
+	JustAfterEach(func() {
+		if CurrentGinkgoTestDescription().Failed {
+			CollectSupport()
+			DescribeNamespace(contexts)
+		}
+	})
 })
```
harsh-px: is it possible to put these 6 lines into a afterEachTest function in common.go and just invoke that from all tests. Looks like all 6 lines are duplicated in each test

drivers/scheduler/k8s/k8s.go
```diff
@@ -2088,44 +2097,46 @@ func (k *K8s) getPodsForApp(ctx *scheduler.Context) ([]v1.Pod, error) {
...
+	k8sOps := k8sops.Instance()
 	var buf bytes.Buffer
 	var err error
-	for _, spec := range ctx.App.SpecList {
-		if obj, ok := spec.(*apps_api.Deployment); ok {
-			buf.WriteString(insertLineBreak(obj.Name))
-			var depStatus *apps_api.DeploymentStatus
+	for _, specObj := range ctx.App.SpecList {
+		if obj, ok := specObj.(*appsapi.Deployment); ok {
+			buf.WriteString(insertLineBreak(fmt.Sprintf("Deployment: %s", obj.Name)))
```
harsh-px: add the namespace here. Do the same at other places.

Something like

`fmt.Sprintf("Deployment: [%s] %s", obj.Namespace, obj.Name))`

PR #381, opened 2019-12-13
PTX-2117 [torpedo] openshift-3.11 job failing during `rebootonenode` …

nareshdesh:
Modified to reboot nodes only once, instead of for each app

### DIFF
--- a/tests/reboot/reboot_test.go
+++ b/tests/reboot/reboot_test.go
@@ -29,8 +29,6 @@ var _ = BeforeSuite(func() {
 var _ = Describe("{RebootOneNode}", func() {
 	It("has to schedule apps and reboot node(s) with volumes", func() {
 		var err error
-		timeout := 60 * time.Second
-		retryInterval := 5 * time.Second
 		var contexts []*scheduler.Context
 		for i := 0; i < Inst().ScaleFactor; i++ {
 			contexts = append(contexts, ScheduleAndValidate(fmt.Sprintf("rebootonenode-%d", i))...)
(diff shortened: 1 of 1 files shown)


tests/reboot/reboot_test.go
```diff
@@ -60,7 +60,10 @@ var _ = Describe("{RebootOneNode}", func() {
 					}
 				})
 
-				Step(fmt.Sprintf("reboot app %s's node(s): %v", ctx.App.Key, nodesToReboot), func() {
+			}
+
+			// Reboot node and check driver status
+				Step(fmt.Sprintf("reboot node one at a time from the node(s): %v",  nodesToReboot), func() {
 					for _, n := range nodesToReboot {
```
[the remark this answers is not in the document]
nareshdesh: In the loop before this section, nodes for volumes of all apps is being collected. The reason for not using `node.GetWorkerNodes()`  in the reboot section  is because there could be nodes which do not have any volumes, like storage-less nodes.

tests/reboot/reboot_test.go
```diff
@@ -37,8 +37,8 @@ var _ = Describe("{RebootOneNode}", func() {
 		}
 
 		Step("get nodes for all apps in test and reboot their nodes", func() {
+			var nodesToReboot []node.Node
 			for _, ctx := range contexts {
-				var nodesToReboot []node.Node
 
 				Step(fmt.Sprintf("get nodes for %s app to reboot, where volumes are attached", ctx.App.Key), func() {
 					volumes, err := Inst().S.GetVolumes(ctx)
```
thiguetta: as mentioned before, as we're stopping  a single node at a time, this logic can go away, and replace for `node.GetWorkerNodes()`

nareshdesh: incorporated review comment

tests/reboot/reboot_test.go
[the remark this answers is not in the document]
nareshdesh: Changed the code to use node.GetWorkerNodes()

tests/reboot/reboot_test.go
```diff
@@ -29,38 +29,19 @@ var _ = BeforeSuite(func() {
...
-						}
-					}
-				})
-
-				Step(fmt.Sprintf("reboot app %s's node(s): %v", ctx.App.Key, nodesToReboot), func() {
+			// Reboot node and check driver status
+				Step(fmt.Sprintf("reboot node one at a time from the node(s): %v",  nodesToReboot), func() {
 					for _, n := range nodesToReboot {
 						err = Inst().N.RebootNode(n, node.RebootNodeOpts{
 							Force: true,
```
mborodin-px: you don't need to create another array of nodes, just loop thru node.GetWorkerNodes() and reboot one by one.

nareshdesh: Just created for now, so that if any change in logic for node selection is required/suggested that will not affect this section.
Will change it now

tests/reboot/reboot_test.go
```diff
@@ -29,38 +29,16 @@ var _ = BeforeSuite(func() {
...
-
-				Step(fmt.Sprintf("get nodes for %s app to reboot, where volumes are attached", ctx.App.Key), func() {
-					volumes, err := Inst().S.GetVolumes(ctx)
-					Expect(err).NotTo(HaveOccurred())
-
-					nodeMap := make(map[string]struct{})
-					for _, v := range volumes {
-						n, err := Inst().V.GetNodeForVolume(v, timeout, retryInterval)
-						Expect(err).NotTo(HaveOccurred())
+				nodesToReboot:=node.GetWorkerNodes()
```
nikolaypopov: Can we follow the style of the rest of the code and separate this with spaces, like:
nodesToReboot := node.GetWorkerNodes()

harsh-px:
@thiguetta just checkout this PR branch, fix formatting and push it here. The branch will be from naresh's fork.

thiguetta:
sorry @harsh-px merged before I had seen your comment, opened #386

PR #382, opened 2019-12-14
Use cloud provider storageclasses if present.

Rohit-PX:
Why do we need this PR?
* Add new volume drivers in torpedo for GKE
   https://portworx.atlassian.net/browse/STOR-126
* This also changes the way the specs are parsed in torpedo, we now look for directories specific to the requested provisioners and if present parse those specs.


drivers/scheduler/k8s/k8s.go
```diff
@@ -262,6 +271,15 @@ func (k *K8s) ParseSpecs(specDir string) ([]interface{}, error) {
 	return specs, nil
 }
 
+func isValidProvider(specPath string) bool {
+	for provider := range spec.SupportedStorageProviders {
+		if strings.Contains(specPath, provider) {
```
disrani-px: Should check for the driver name being the suffix, this will match the driver anywhere in the path

drivers/volume/common.go
```diff
@@ -0,0 +1,212 @@
...
+
+// RandomizeVolumeName randomizes the volume name from the given name
+func (d *DefaultDriver) RandomizeVolumeName(params string) string {
+	re := regexp.MustCompile("(" + api.Name + "=)([0-9A-Za-z_-]+)(,)?")
+	return re.ReplaceAllString(params, "${1}${2}_"+uuid.New()+"${3}")
+}
+
+// WaitDriverUpOnNode must wait till the volume driver becomes usable on a given node
+func (d *DefaultDriver) WaitDriverUpOnNode(n node.Node, timeout time.Duration) error {
+	return nil
```
disrani-px: These should return NotImplented/NotSupported error

drivers/volume/portworx/portworx.go
```diff
@@ -60,13 +60,12 @@ const (
 	RuleScaleTypeResizeDisk = "resize-disk"
 	// RuleMaxSize is name for rule max size
 	RuleMaxSize = "maxsize"
+	// CsiStorage CSI storage driver name
+	CsiStorage torpedovolume.StorageProvisionerType = "csi"
+	// GkeStorage GKE storage driver name
+	GkeStorage torpedovolume.StorageProvisionerType = "gke"
```
disrani-px: These should not be here?

drivers/volume/volume.go
```diff
@@ -156,6 +156,22 @@ type Driver interface {
...
+	PortworxStorage StorageProvisionerType = "portworx"
+	//CsiStorage csi storage name
+	CsiStorage StorageProvisionerType = "csi"
+	//GkeStorage  gce storage name
+	GkeStorage StorageProvisionerType = "gke"
+)
+
+// Provisioners types of supported provisioners
+var Provisioners = map[StorageProvisionerType]StorageProvisionerType{
+	PortworxStorage: "kubernetes.io/portworx-volume",
```
disrani-px: These provisioners should be in the driver implementation

drivers/volume/gke/gke.go
```diff
@@ -0,0 +1,19 @@
+package gke
```
disrani-px: Let's call the driver gce, since this can work on k8s on GCE too

drivers/volume/volume.go
Rohit-PX: moved.

drivers/volume/portworx/portworx.go
Rohit-PX: moved to their individual driver implementations.

drivers/volume/gke/gke.go
Rohit-PX: Renamed driver and variables to gce.

drivers/volume/common.go
Rohit-PX: Returning NotSupported type error objects.

drivers/scheduler/spec/factory.go
```diff
@@ -15,8 +15,23 @@ type Factory struct {
 	specParser Parser
 }
 
+const (
```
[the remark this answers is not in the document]
Rohit-PX: Implemented GetVolumeDrivers method.

drivers/scheduler/k8s/k8s.go
```diff
@@ -193,11 +193,17 @@ func (k *K8s) RefreshNodeRegistry() error {
...
-func (k *K8s) ParseSpecs(specDir string) ([]interface{}, error) {
+func (k *K8s) ParseSpecs(specDir, storageProvisioner string) ([]interface{}, error) {
 	fileList := make([]string, 0)
 	if err := filepath.Walk(specDir, func(path string, f os.FileInfo, err error) error {
 		if !f.IsDir() {
-			fileList = append(fileList, path)
+			if !isValidProvider(path) {
+				fileList = append(fileList, path)
+			} else { //specs from cloud provider directory
+				if strings.Contains(path, storageProvisioner) {
```
disrani-px: Needs to be similar to the check in `isValidProvider`

drivers/volume/gce/gce.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (d *gce) Init(sched string, nodeDriver string, token string, storageProvisioner string) error {
+	logrus.Infof("Using the GCE volume driver with provisioner %s under scheduler: %v", storageProvisioner, sched)
+	// Set provisioner for torpedo
+	if storageProvisioner != "" {
+		if p, ok := provisioners[torpedovolume.StorageProvisionerType(storageProvisioner)]; ok {
+			torpedovolume.StorageProvisioner = p
+		}
+	} else {
+		torpedovolume.StorageProvisioner = torpedovolume.DefaultStorageProvisioner
```
disrani-px: This will end up using `portworx` as the provisioner. We should just return error instead

drivers/scheduler/k8s/k8s.go
Rohit-PX: fixed.

drivers/volume/gce/gce.go
Rohit-PX: Returning error here.

PR #388, opened 2019-12-20
PTX-2093 Added GetReplicaSets implementation

mborodin-px:


thiguetta:
is it already tested in dev-test job?

mborodin-px:
@thiguetta there is nothing to test in dev-test job as that method is not used anywhere. I just ran amke build to make sure there is no issues with the build. That method is needed for my Autopilot storage pool expansion and for @nareshdesh in VPS automation.

PR #390, opened 2020-01-03
PTX-2097 Implement calculation of Storage Pool sizes

mborodin-px:
Implementation of size calculation for Volumes and Storage Pools based on app workload and autopilot rule

Volume resize tests passed https://jenkins.portworx.co/job/Autopilot/job/tp-aut-resize-pvc/99/
Storage pool tests passed https://jenkins.portworx.co/job/Autopilot/job/tp-aut-str-pool/73/


drivers/node/node.go
```diff
@@ -27,6 +27,15 @@ const (
 	Directory FindType = "d"
 )
 
+// StoragePool is the storage pool structire on the node
+type StoragePool struct {
+	UUID                string
+	TotalSize           uint64
```
harsh-px: embed the existing StoragePool structure returned by SDK API and add other fields that test needs.

This will be useful when we need to start looking at other fields in the pools (like labels) in future. We want to avoid redeclaring fields that are already present in api.StoragePool.

drivers/scheduler/k8s/k8s.go
```diff
@@ -18,6 +18,7 @@ import (
 	stork_api "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	k8s_ops "github.com/portworx/sched-ops/k8s"
 	"github.com/portworx/sched-ops/task"
+	torpedo_api "github.com/portworx/torpedo/drivers/api"
```
harsh-px: rename alias as `torpedo_driver_api`

drivers/scheduler/k8s/k8s.go
```diff
@@ -67,9 +68,12 @@ const (
 	//DefaultRetryInterval  Default retry interval
 	DefaultRetryInterval = 10 * time.Second
 	//DefaultTimeout default timeout
-	DefaultTimeout               = 2 * time.Minute
-	resizeSupportedAnnotationKey = "torpedo/resize-supported"
-	autopilotResizeAnnotationKey = "torpedo.io/autopilot-enabled"
+	DefaultTimeout                                 = 2 * time.Minute
+	resizeSupportedAnnotationKey                   = "torpedo/resize-supported"
+	scaleStorageAnnotationKey                      = "torpedo.io/scale-storage-enabled"
```
harsh-px: use `torpedo.io` as const and have other keys reference that

drivers/scheduler/k8s/specs/aut-postgres/aut-postgres-storage.yaml
```diff
@@ -0,0 +1,26 @@
+##### Portworx storage class
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+    name: px-postgres-sc
+provisioner: kubernetes.io/portworx-volume
+parameters:
+   repl: "3"
+   priority_io: "high"
+   shared: "true"
```
harsh-px: postgres should not use a shared volume. If both pods run on the same node (due to the pod affinity), PX will allow mounting a regular volume on multiple pods

drivers/scheduler/k8s/specs/aut-postgres/aut-postgres-storage.yaml
```diff
@@ -0,0 +1,26 @@
...
+allowVolumeExpansion: true
+---
+##### Portworx persistent volume claim
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+   name: postgres-data
+   annotations:
+     volume.beta.kubernetes.io/storage-class: px-postgres-sc
```
harsh-px: instead of the annotation, use storageClassName in the spec.

drivers/volume/portworx/portworx.go
```diff
@@ -977,6 +989,100 @@ func (d *portworx) WaitDriverDownOnNode(n node.Node) error {
...
+								n.Name, nodeStoragePool.UUID, nStoragepool.CalculatedTotalSize, nodeStoragePool.TotalSize)
+						}
+					}
+				}
+			}
+		}
+		if allDone {
+			return "", false, nil
+		}
+		return "", true, fmt.Errorf("some pools size are not as expected")
```
harsh-px: nit: reword as `some pools sizes are not as expected`

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,215 @@
...
+			Values:   []string{"80"},
+		},
+	},
+	Actions: []torpedo_api.AutopilotRuleAction{
+		{
+			Name: torpedo_api.StorageSpecAction,
+			Params: map[string]string{
+				torpedo_api.RuleActionsScalePercentage: "50",
+				torpedo_api.RuleScaleType:              torpedo_api.RuleScaleTypeAddDisk,
+				// torpedo_api.RuleMaxSize:                "322122547200",
```
harsh-px: don't think we support max size for pool resize. comment can be removed. same for other places

drivers/scheduler/k8s/k8s.go
```diff
@@ -1123,35 +1146,46 @@ func (k *K8s) destroyCoreObject(spec interface{}, opts map[string]bool, app *spe
...
+			Value: value,
+		})
 	}
 
-	logrus.Infof("Using Autopilot Object: %+v\n", obj)
+	return envVars
+}
 
-	return obj, nil
+func (k *K8s) isPvcAnnotationAutopilotEnabled(spec interface{}) bool {
```
harsh-px: just called this `isAutopilotEnabledForPVC`. No need to have annotations in the function name. Same for other functions

drivers/scheduler/k8s/k8s.go
```diff
@@ -1654,24 +1688,20 @@ func (k *K8s) InspectVolumes(ctx *scheduler.Context, timeout, retryInterval time
...
-					pvcAnnotationSupported, _ = strconv.ParseBool(pvcAnnotation)
-				}
-				if pvcAnnotationSupported {
-					expectedPVCSize := apParams.AutopilotRuleParameters.ExpectedPVCSize
-
+			apRule := ctx.Options.AutopilotRules
+			if apRule.Name != "" && k.isPvcAnnotationAutopilotEnabled(obj) {
+				if k.isPvcResizeAnnotationAutopilotEnabled(obj) {
+					// expectedPVCSize := apParams.AutopilotRuleParameters.ExpectedPVCSize
+					expectedPVCSize := int64(27179089920)
```
harsh-px: why is `27179089920` hardcoded here?

drivers/volume/portworx/portworx.go
```diff
@@ -1538,6 +1644,32 @@ func (d *portworx) getStorageStatus(n node.Node) string {
...
+	return nodeStoragePools, nil
+}
+
+func (d *portworx) EnableStorageExpansion(apRule torpedo_api.AutopilotRule) error {
+	_, err := d.schedOps.CreateAutopilotRule(apRule)
+	if err != nil {
+		return err
+	}
+	logrus.Infof("Created Autopilot rule: %v", apRule.Name)
+	d.autopilotRule = apRule
```
harsh-px: this means we can only have a single rule for storage pool expansion.

drivers/volume/portworx/schedops/schedops.go
```diff
@@ -40,6 +42,8 @@ type Driver interface {
 	// GetRemotePXNodes returns list of PX node found on destination k8s cluster
 	// referenced by kubeconfig
 	GetRemotePXNodes(destKubeConfig string) ([]node.Node, error)
+
+	CreateAutopilotRule(apParams torpedo_api.AutopilotRule) (*ap_api.AutopilotRule, error)
```
harsh-px: public function so would need a function doc above this

drivers/volume/volume.go
```diff
@@ -135,6 +137,12 @@ type Driver interface {
 
 	// Collect live diags on a node
 	CollectDiags(n node.Node) error
+
+	// ValidateStorage validates all the storage
```
harsh-px: function comment is incomplete

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,215 @@
...
+	unscheduledResizeTimeout = 10 * time.Minute
+)
+
+var ruleResizeBy50IfPoolUsageMoreThan50AddDisk = torpedo_api.AutopilotRule{
+	Name: "storage-pool-add-disk",
+	ConditionExpressions: []torpedo_api.AutopilotRuleCondition{
+		{
+			Key:      torpedo_api.PxPoolAvailableCapacityMetric,
+			Operator: torpedo_api.LabelSelectorOpLt,
+			Values:   []string{"80"},
```
harsh-px: the value `80` here does not match with rule name `ruleResizeBy50IfPoolUsageMoreThan50AddDisk`

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,215 @@
...
+	},
+}
+
+var ruleResizeBy50IfPoolUsageMoreThan50ResizeDisk = torpedo_api.AutopilotRule{
+	Name: "storage-pool-resize-disk",
+	ConditionExpressions: []torpedo_api.AutopilotRuleCondition{
+		{
+			Key:      torpedo_api.PxPoolAvailableCapacityMetric,
+			Operator: torpedo_api.LabelSelectorOpLt,
+			Values:   []string{"85"},
```
harsh-px: value `85` does not match with rule name `ruleResizeBy50IfPoolUsageMoreThan50ResizeDisk`

tests/common.go
```diff
@@ -52,28 +56,30 @@ const (
...
+	defaultLogLocation                   = "/mnt/torpedo_support_dir"
+	defaultLogLevel                      = "debug"
+	defaultAppScaleFactor                = 1
+	defaultMinRunTimeMins                = 0
+	defaultChaosLevel                    = 5
+	defaultStorageUpgradeEndpointURL     = "https://install.portworx.com/upgrade"
+	defaultStorageUpgradeEndpointVersion = "2.1.1"
+	defaultStorageProvisioner            = "portworx"
+	defaultStorageNodesPerAZ             = 2
+	storagePoolResizeAnnotationKey       = "torpedo.io/autopilot-storage-pool-resize-enabled"
```
harsh-px: you have the `torpedo.io/autopilot-storage-pool-resize-enabled` key declared at 2 places

```
drivers/scheduler/k8s/k8s.go:   autopilotStoragePoolResizeEnabledAnnotationKey = "torpedo.io/autopilot-storage-pool-resize-enabled"
tests/common.go:        storagePoolResizeAnnotationKey       = "torpedo.io/autopilot-storage-pool-resize-enabled"
```

tests/common.go
```diff
@@ -356,6 +362,48 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
 	})
 }
 
+// AddLabelsOnNode adds labels on the node
+func AddLabelsOnNode(n node.Node, labels map[string]string) error {
+	for labelKey, labelValue := range labels {
+		if err := k8s.Instance().AddLabelOnNode(n.Name, labelKey, labelValue); err != nil {
```
harsh-px: the tests directly should not assume k8s. This breaks the scheduler abstraction. Check for other places in PR too. 

Tests should invoke the scheduler driver interface APIs.

tests/common.go
harsh-px: Actually `storagePoolResizeAnnotationKey` doesn't look like it's being referenced anywhere. This can be removed.

drivers/node/node.go
```diff
@@ -27,6 +28,13 @@ const (
 	Directory FindType = "d"
 )
 
+// StoragePool is the storage pool structure on the node
+type StoragePool struct {
+	*api.StoragePool
+	InitialSize  uint64
```
harsh-px: just add comments above the variables to indicate what they are.

For e.g

// WorkloadSize is the size in GiB of the workload that will be launched by test on this storage pool

drivers/scheduler/k8s/k8s.go
```diff
@@ -578,6 +581,7 @@ func (k *K8s) createNamespace(app *spec.AppSpec, namespace string) (*v1.Namespac
 			map[string]string{
 				"creator": "torpedo",
 				"app":     app.Key,
+				"name":    namespace,
```
harsh-px: it's a little odd to put the name of the namespace in the labels.

labels are used to group objects and then access them using those group queries. Putting the name as a label has no additional value as you can already access the object directly with names.

Is your use test to tell autopilot to monitor all applications per test or just one application per test? If it's all applications per test, we can add a testname label.

drivers/scheduler/k8s/k8s.go
```diff
@@ -897,6 +890,12 @@ func (k *K8s) createCoreObject(spec interface{}, ns *v1.Namespace, app *spec.App
 	if obj, ok := spec.(*appsapi.Deployment); ok {
 		obj.Namespace = ns.Name
 		obj.Spec.Template.Spec.Volumes = k.substituteNamespaceInVolumes(obj.Spec.Template.Spec.Volumes, ns.Name)
+		if isDeploymentAppEnvEnabled(obj.Annotations) && len(options.AppEnvs) != 0 {
```
harsh-px: Do we need the `isDeploymentAppEnvEnabled`? Why not always append the env variables if they are defined in the options.

Trying to keep the test logic simple with lesser if checks. That would also be simpler for the user of Schedule. Just add the env options during the call and no need to make any annotation changes to the yamls.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1150,51 +1149,53 @@ func (k *K8s) destroyCoreObject(spec interface{}, opts map[string]bool, app *spe
...
-		}
-		obj.Spec.Actions = append(obj.Spec.Actions, actions)
+	for _, appEnvVar := range appEnvVars {
+		tmpEnvMapVars[appEnvVar.Name] = appEnvVar.Value
+	}
+	for key, value := range tmpEnvMapVars {
+		envVars = append(envVars, v1.EnvVar{
+			Name:  key,
+			Value: value,
+		})
```
harsh-px: you should be able to reduce one for loop here.  Skip the tmpEnvMapVars and directly append to the envVars in the first 2 loops.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1150,51 +1149,53 @@ func (k *K8s) destroyCoreObject(spec interface{}, opts map[string]bool, app *spe
...
 	k8sOps := k8sops.Instance()
-	apRule, err := k8sOps.CreateAutopilotRule(autopilotRule)
-	if errors.IsAlreadyExists(err) {
-		if apRule, err = k8sOps.GetAutopilotRule(autopilotRule.Name); err == nil {
-			logrus.Infof("Found existing AutopilotRule: %v", autopilotRule.Name)
-			return apRule, nil
+
+	t := func() (interface{}, bool, error) {
+		autopilotRule.Labels = map[string]string{
+			"creator": "torpedo",
```
harsh-px: `"creator": "torpedo"` is now being used at 3 places in the code base. Use const.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1664,7 +1677,17 @@ func (k *K8s) GetVolumeParameters(ctx *scheduler.Context) (map[string]map[string
...
+	var err error
 	k8sOps := k8sops.Instance()
+	apRule := ctx.ScheduleOptions.AutopilotRule
+	calculatedVolumesMap := make(map[string]uint64)
+	if apRule.Name != "" {
+		calculatedVolumesMap, err = k.CalculateVolumesSizes(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed to calculate volume sizes. Err: %v", err)
+		}
+		logrus.Debugf("calculated volumes map: %+v", calculatedVolumesMap)
```
harsh-px: rename `calculated volumes size map`

drivers/scheduler/k8s/k8s.go
```diff
@@ -1999,11 +1992,11 @@ func (k *K8s) resizePVCBy1GB(ctx *scheduler.Context, pvc *v1.PersistentVolumeCla
...
 	vol := &volume.Volume{
-		ID:        string(pvc.UID),
-		Name:      pvc.Name,
-		Namespace: pvc.Namespace,
-		Size:      uint64(sizeInt64),
-		Shared:    k.isPVCShared(pvc),
+		ID:          string(pvc.UID),
+		Name:        pvc.Name,
+		Namespace:   pvc.Namespace,
+		ResizedSize: uint64(sizeInt64),
```
harsh-px: why add a new field called ResizedSize? I don't see you using it anywhere.

This adds confusion since there is already a field called Size. Which does does the user of the API use? Is Size implicitly outdated initial size and should I stop using it?

drivers/scheduler/k8s/k8s.go
```diff
@@ -2841,6 +2834,109 @@ func (k *K8s) destroyBackupObjects(
...
+
+	pvcMatchLabels := map[string]string{}
+	for _, volume := range volumes {
+		if k.IsAutopilotEnabledForPvc(volume.Annotations) {
+			for k, v := range volume.Labels {
+				pvcMatchLabels[k] = v
+			}
+		}
+	}
+	ns, err := k8sOps.GetNamespace(volumes[0].Namespace)
```
harsh-px: This (volumes[0]) will panic if the app doesn't have volumes.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2841,6 +2834,109 @@ func (k *K8s) destroyBackupObjects(
...
 }
 
+// EnableVolumeExpansion creates autopilot rules for given context
+func (k *K8s) EnableVolumeExpansion(ctx *scheduler.Context) error {
+	if ctx == nil {
+		return fmt.Errorf("no context provided")
+	}
+	k8sOps := k8sops.Instance()
+	apRule := ctx.ScheduleOptions.AutopilotRule
+	volumes, _ := k.GetVolumes(ctx)
```
harsh-px: any reason to ignore err here?

drivers/scheduler/k8s/k8s.go
```diff
@@ -2841,6 +2834,109 @@ func (k *K8s) destroyBackupObjects(
 	return nil
 }
 
+// EnableVolumeExpansion creates autopilot rules for given context
+func (k *K8s) EnableVolumeExpansion(ctx *scheduler.Context) error {
```
harsh-px: can we do this EnableVolumeExpansion logic at Schedule time itself? In other words, do we need a new Interface and an explicit call from test to enable this.

drivers/scheduler/scheduler.go
```diff
@@ -178,6 +147,9 @@ type Driver interface {
 	// ResizeVolume resizes all the volumes of a given context
 	ResizeVolume(*Context, string) ([]*volume.Volume, error)
 
+	// CalculateVolumesSize calculates volumes size based on autopilot rule
+	CalculateVolumesSizes(*Context) (map[string]uint64, error)
```
harsh-px: CalculateVolumesSizes doesn't need to a an interface. It's just an internal helper method that is being called from InspectVolumes.

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,247 @@
...
+	scaleTimeout             = 2 * time.Hour
+	workloadTimeout          = 5 * time.Hour
+	retryInterval            = 30 * time.Second
+	unscheduledResizeTimeout = 10 * time.Minute
+)
+
+// Resize storage pool by adding disks by 50% until storage available capacity is less than %70
+var ruleResizeBy50IfPoolAvailableUsageMoreThan70AddDisk = apapi.AutopilotRule{
+	ObjectMeta: meta_v1.ObjectMeta{
+		Name: "storage-pool-resize-add-disk-70",
```
harsh-px: for this test, your object name and the label in selector is identical.

But this is not the case for the test below (ruleResizeBy50IfPoolAvailableUsageMoreThan70ResizeDisk)

tests/common.go
```diff
@@ -358,6 +360,54 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
...
+func ValidateStoragePools(contexts []*scheduler.Context, apRules []ap_api.AutopilotRule) {
+	var err error
+	var expectedStoragePoolsWorkloadSize = make(map[string]uint64)
+	for _, ctx := range contexts {
+		Step(fmt.Sprintf("get replica sets %s app volumes", ctx.App.Key), func() {
+			appVolumes, err := Inst().S.GetVolumes(ctx)
+			expect(err).NotTo(haveOccurred())
+			expect(appVolumes).NotTo(beEmpty())
+			for _, volume := range appVolumes {
+				if Inst().S.IsAutopilotEnabledForPvc(volume.Annotations) {
```
harsh-px: The IsAutopilotEnabledForPvc() interface method should not be needed.

The logic below should upgrade the `.StoragePools[idx].WorkloadSize` regardless of whether autopilot is enabled or not. The fact that a workload runs on a storage pool doesn't have direct correlation to autopilot running in the cluster.

Now when we do `ValidateStoragePoolsSizes`, that's where the volume driver can detect if autopilot is enabled for that pool and then check if the pool expanded as per the workload size. If autopilot is not enabled, then the pool is expected to stay at it's original size and the validation should pass.

In general, I'm trying to reduce the many if conditionals we have here and the indirect/implicit expectations between different code pieces.

tests/common.go
```diff
@@ -358,6 +360,54 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
...
+	for _, n := range node.GetStorageDriverNodes() {
+		for idx, pool := range n.StoragePools {
+			// Adding 10% of initial storage size to workload as default storage usage comes with it
+			tenPercentOfStorage := n.StoragePools[idx].InitialSize / 10
+			n.StoragePools[idx].WorkloadSize = tenPercentOfStorage + expectedStoragePoolsWorkloadSize[pool.Uuid]
+		}
+		err = node.UpdateNode(n)
+		expect(err).NotTo(haveOccurred())
+	}
+	logrus.Debugf("storage pools workload sizes map: %v\n", expectedStoragePoolsWorkloadSize)
```
harsh-px: don't need the "\n" here

drivers/volume/portworx/portworx.go
```diff
@@ -952,6 +995,64 @@ func (d *portworx) WaitDriverDownOnNode(n node.Node) error {
 	return nil
 }
 
+// ValidateStoragePoolsSizes checks if sizes of storage pool are expanded based on autopilot rules
+func (d *portworx) ValidateStoragePoolsSizes(apRules []apapi.AutopilotRule) error {
```
harsh-px: No need to pass in apRules.

The portworx volume driver has an instance of k8s sched-ops (https://github.com/portworx/torpedo/tree/master/drivers/volume/portworx/schedops) and it can simply use that to enumerate all autopilot rules that affect the storage pools being validated.

This is a better check anyways since the volume driver validate does not depend on it being provided a correct list of rules but instead it will independently figure out the list of apRules that affect the pool.

drivers/volume/volume.go
```diff
@@ -135,6 +139,16 @@ type Driver interface {
...
+
+	// ValidateStoragePoolsSizes validates all the storage pools sizes basen on
+	// autopilot rules parameters
+	ValidateStoragePoolsSizes([]apapi.AutopilotRule) error
+
+	// EnableStorageExpansion enables storage expansion using autopilot
+	EnableStorageExpansion([]apapi.AutopilotRule) error
+
+	//CalculateTotalSize calculates expected size based on autopilot rule, initial and workload sizes
+	CalculateTotalSize(apRule apapi.AutopilotRule, initSize uint64, workloadSize uint64) uint64
```
harsh-px: `CalculateTotalSize` of what? Clarify either in the name or the function doc

drivers/volume/portworx/portworx.go
```diff
@@ -1775,6 +1888,65 @@ func (d *portworx) CollectDiags(n node.Node) error {
 	return nil
 }
 
+// CalculateTotalSize calculates expected size based on autopilot rule, initial and workload sizes
+func (d *portworx) CalculateTotalSize(apRule apapi.AutopilotRule, initSize, wklSize uint64) uint64 {
+	var maxSize, calculatedTotalSize uint64
+	calculatedTotalSize = initSize
+	applyRuleTrigger := true
+	for applyRuleTrigger {
```
harsh-px: the logic here is little difficult to understand. Can you add a summary of what this is doing?

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,247 @@
...
+				Params: map[string]string{
+					portworx.RuleActionsScalePercentage: "50",
+					portworx.RuleScaleType:              portworx.RuleScaleTypeResizeDisk,
+				},
+			},
+		},
+	},
+}
+
+var ruleResizeBy50IfPoolTotalSizeLessThan300ResizeDisk = apapi.AutopilotRule{
```
harsh-px: unused

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+var _ = BeforeSuite(func() {
+	InitInstance()
+})
+
+// This testsuite is used for performing basic scenarios with Autopilot rules where it
+// schedules apps and wait until workload is completed on the volumes and then validates
+// storage pool(s) size(s) on the nodes where volumes reside
+var _ = Describe(fmt.Sprintf("{%sBasicAddDisk}", testNameSuite), func() {
```
harsh-px: incorrect name "sBasicAddDisk". The `autopilotruleBasicTestCases` you use here also has a resize disk test case

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,247 @@
...
+				contexts = append(contexts, context...)
+			}
+		})
+
+		appEnvs := []scheduler.EnvVar{
+			scheduler.EnvVar{
+				Name:  "SIZE",
+				Value: strconv.FormatUint(appWorkloadSize/(1024*1024*1024), 10),
+			},
+		}
```
harsh-px: Can be written as 

```
		appEnvs := []scheduler.EnvVar{
			{
				Name:  "SIZE",
				Value: strconv.FormatUint(appWorkloadSize/(1024*1024*1024), 10),
			},
		}
```

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,247 @@
...
+				})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(context).NotTo(BeEmpty())
+				contexts = append(contexts, context...)
+			}
+		})
+
+		appEnvs := []scheduler.EnvVar{
+			scheduler.EnvVar{
+				Name:  "SIZE",
```
harsh-px: "SIZE" needs to be a `const` as it is a special variable and used at multiple places

tests/common.go
```diff
@@ -358,6 +360,54 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
...
+	for labelKey, labelValue := range labels {
+		if err := Inst().S.AddLabelOnNode(n, labelKey, labelValue); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// ValidateStoragePools validates sizes of storage pools based on autopilot rules
+func ValidateStoragePools(contexts []*scheduler.Context, apRules []ap_api.AutopilotRule) {
```
harsh-px: under the same package, create validate.go and move all Validate* methods there.

common.go is getting too big.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1362,6 +1363,18 @@ func (k *K8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time
...
 func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
 	var podList []v1.Pod
+	k8sOps := k8sops.Instance()
+	apRule := ctx.ScheduleOptions.AutopilotRule
+	if apRule.Name != "" {
+		if err := k8sOps.DeleteAutopilotRule(apRule.ObjectMeta.Name); err != nil {
+			if !errors.IsNotFound(err) {
+				return &scheduler.ErrFailedToDestroyAutopilotRule{
+					Name:  apRule.Name,
+					Cause: fmt.Sprintf("%v", err),
```
thiguetta: if not using any message, let's use `err.Error()` which already returns a string

drivers/scheduler/scheduler.go
```diff
@@ -220,6 +178,15 @@ type Driver interface {
 
 	// Get token for a volume
 	GetTokenFromConfigMap(string) (string, error)
+
+	// AddLabelOnNode adds key value labels on the node
+	AddLabelOnNode(node.Node, string, string) error
+
+	//IsAutopilotEnabledForPvc checks if autopilot enabled for a given PVC
+	IsAutopilotEnabledForPvc(map[string]string) bool
```
harsh-px: API should be

```
IsAutopilotEnabledForVolume(v *volume.Volume)
```

2 issues with current method

1. It calls out PVC which is just a k8s concept
2. It says ForPvc but provides a map of strings. There is an implicit assumption is the API that volumes have annotations or labels which is a map[string]string

drivers/volume/portworx/portworx.go
```diff
@@ -71,6 +94,8 @@ const (
 	validateClusterStartTimeout      = 2 * time.Minute
 	validatePXStartTimeout           = 5 * time.Minute
 	validateNodeStopTimeout          = 5 * time.Minute
+	validateStoragePoolSizeTimeout   = 3 * time.Hour
```
harsh-px: Create a PTX to make this a function of the workload size. A static 3 hour timeout is too large for small workloads.

tests/autopilot_pool_expand/aut_pool_expand_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/portworx/torpedo/drivers/scheduler"
+	portworx "github.com/portworx/torpedo/drivers/volume/portworx"
+	. "github.com/portworx/torpedo/tests"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	testNameSuite            = "AutPoolResize"
+	timeout                  = 30 * time.Minute
+	scaleTimeout             = 2 * time.Hour
```
harsh-px: both of these are unused

drivers/volume/portworx/portworx.go
```diff
@@ -952,6 +998,68 @@ func (d *portworx) WaitDriverDownOnNode(n node.Node) error {
 	return nil
 }
 
+func (d *portworx) ValidateStoragePools() error {
+	// check if the sizes of pools match the expected sizes after storage expand
+	if len(d.autopilotRules) > 0 {
```
harsh-px: I thought we had decided the query autopilotRules at runtime

https://github.com/portworx/torpedo/pull/390#discussion_r369187563

The portworx driver should not be needing to save `autopilotRules`

drivers/volume/portworx/portworx.go
```diff
@@ -980,6 +1088,25 @@ func (d *portworx) pickAlternateClusterManager(n node.Node) (api.OpenStorageNode
 	return nil, fmt.Errorf("failed to get an alternate cluster manager for %s", n.Name)
 }
 
+// EnableStorageExpansion creates autopilot rules
+func (d *portworx) EnableStorageExpansion(apRules []apapi.AutopilotRule) error {
```
harsh-px: The method says `EnableStorageExpansion` but in reality it isn't doing that. 

The implementation just takes in a bunch of apRules and applies them. There is an implicit assumption that the caller has written rules that perform storage expansion.

drivers/volume/portworx/portworx.go
```diff
@@ -980,6 +1088,25 @@ func (d *portworx) pickAlternateClusterManager(n node.Node) (api.OpenStorageNode
...
+		autopilotRule, err := d.schedOps.CreateAutopilotRule(apRule)
+		if err != nil {
+			return err
+		}
+		d.autopilotRules = append(d.autopilotRules, autopilotRule)
+	}
+	return nil
+}
+
+func (d *portworx) IsStorageExpansionEnabled() bool {
```
harsh-px: This is incorrect. 

Just because you have autopilotRules does not mean storage expansion is enabled.

What do we do when an autopilot rule has been written for a different use case?

harsh-px:
can you fetch and merge this branch with master?

PR #391, opened 2020-01-08
PTX-2144 Fix security annotations for rabbitmq app

nikolaypopov:
Fix security annotations for rabbitmq app.

### DIFF
--- a/drivers/scheduler/k8s/specs/rabbitmq/px-rabbitmq-app.yml
+++ b/drivers/scheduler/k8s/specs/rabbitmq/px-rabbitmq-app.yml
@@ -168,9 +168,8 @@ spec:
   volumeClaimTemplates:
     - metadata:
         name: rabbitmq-data
-        annotations:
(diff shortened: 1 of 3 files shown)


drivers/scheduler/k8s/specs/rabbitmq/px-rabbitmq-app.yml
```diff
@@ -168,8 +168,9 @@ spec:
   volumeClaimTemplates:
     - metadata:
         name: rabbitmq-data
+        annotations:
+          volume.beta.kubernetes.io/storage-class: px-rabbitmq-sc
       spec:
-        storageClassName: rabbitmq-sc
```
harsh-px: can we fix torpedo to also look for `storageClassName`? using storageClassName in the spec is the right way to specify it. Using annotations was the previous method.

nikolaypopov: Done, found an issue, no need to add annotations here.

PR #393, opened 2020-01-10
PTX-2183 Adding write check for px volume in postgres app

nikolaypopov:
Adding write check for px volume in postgres app

### DIFF
--- a/drivers/scheduler/k8s/specs/postgres/px-postgres-app.yaml
+++ b/drivers/scheduler/k8s/specs/postgres/px-postgres-app.yaml
@@ -37,12 +37,12 @@ spec:
           name: postgredb
         readinessProbe:
           exec:
-            command: ["touch", "/var/lib/postgresql/data/pgdata/PG_VERSION"]
+            command: ["psql", "-U", "pgbench", "-c", "CREATE TABLE IF NOT EXISTS test_table ( test_text TEXT ); INSERT INTO test_table VALUES ( 'test_message' ); SELECT * FROM test_table; SELECT pg_sleep(1); DELETE FROM test_table;"]
           initialDelaySeconds: 15
           timeoutSeconds: 2
         livenessProbe:
           exec:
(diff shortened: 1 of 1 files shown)


drivers/scheduler/k8s/specs/postgres/px-postgres-app.yaml
```diff
@@ -37,12 +37,12 @@ spec:
           name: postgredb
         readinessProbe:
           exec:
-            command: ["psql", "-U", "pgbench", "-d", "pgbench", "-c", "SELECT 1"]
+            command: ["touch", "/var/lib/postgresql/data/pgdata/PG_VERSION"]
```
harsh-px: @nikolaypopov can we try if we can do both?

i.e run a pgsql command first and then use `&&` and run the touch command.


This will ensure the the postgres DB is up and also ensure we can write to our mounted volume. Let's try this in your test cluster.

nikolaypopov: Done.

nikolaypopov:
> in general, it looks good to me, not so sure about `IF NOT EXIST` for postgres, post the link to the tests

This `IF NOT EXIST` is need here since the probes are constantly running this check and we only need to create the database once. Without this `IF NOT EXIST` the command will fail with `database already exists`.

thiguetta:
> > in general, it looks good to me, not so sure about `IF NOT EXIST` for postgres, post the link to the tests
> 
> This `IF NOT EXIST` is needed here since the probes are constantly running this check and we only need to create the database once. Without this `IF NOT EXIST` the command will fail with `database already exists`.

yeah, just wanted to make sure Postgres supports it, not all DBMS does

nikolaypopov:
I ran the test here with my private image, postgres has no issues, but test failed due to PX node being down
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20NextPX/job/tp-nextpx-ocp4.2-op-csi/60

PR #394, opened 2020-01-13
support to custom app parameters

thiguetta:
this PR fixes PTX-1878  and required by PTX-1714

it adds a new parameter `custom-config` which takes a custom yml config file. expected format
```
<appId>:
   replicas: <repliicas>
   volume_size: <volume_size>
   ... <any new custom parameter which might be added in the future>
```
i.e
```yaml
nginx-sharedv4:
   replicas: 24
```

in `drivers/scheduler/scheduler.go` we have the following structure
```go
// AppConfig custom settings
type AppConfig struct {
	Replicas   int    `yaml:"replicas"`
	VolumeSize string `yaml:"volume_size"`
}
```

in specs we use text/template notation
i.e
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  replicas: {{ .Replicas }}
```

with this PR we can also use conditionals, loops and other features available in go `text/template`

i.e for default value 
``` yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  {{ if .Replicas }}
  replicas: {{ .Replicas }}
  {{ else }}
  replicas: 3{{ end }}
```

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -92,7 +92,7 @@ type K8s struct {
 	NodeDriverName      string
 	VolDriverName       string
 	secretConfigMapName string
-	customConfig        map[string]*scheduler.AppConfig
+	customConfig        map[string]scheduler.AppConfig
 }
 
 //IsNodeReady  Check whether the cluster node is ready
@@ -122,6 +122,11 @@ func (k *K8s) String() string {
 
 //Init Initialize the driver
 func (k *K8s) Init(schedOpts scheduler.InitOptions) error {
+	k.NodeDriverName = schedOpts.NodeDriverName
+	k.VolDriverName = schedOpts.VolDriverName
+	k.secretConfigMapName = schedOpts.SecretConfigMapName
+	k.customConfig = schedOpts.CustomAppConfig
+
 	nodes, err := k8sops.Instance().GetNodes()
 	if err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


thiguetta:
these tests show that there is no regression https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/261/console, if you don't provide any custom parameter it will still work.

drivers/scheduler/k8s/k8s.go
```diff
@@ -199,14 +204,36 @@ func (k *K8s) ParseSpecs(specDir string) ([]interface{}, error) {
 
 	var specs []interface{}
 
+	splitPath := strings.Split(specDir, "/")
+	appName := ""
+	if len(splitPath) > 0 {
```
rodrigc: As long as `specDir` is a non-empty string, `len(splitPath)` will always be greater than 0.
If specDir has no slashes in it, `len(splitPath)` will be 1.

Will there ever be a case where `len(splitPath)` <= 0, and be valid for this function?

thiguetta: yeah, that's exactly the case, we want to specDir is non-empty implicitly

rodrigc: If there is no case where specDir can be empty, then I would suggest changing the logic something like:

```
if len(specDir) == 0 {
    return an err here
}
splitPath := strings.Split(specDir, "/")
appName = splitPath[len(splitPath)-1]
```

thiguetta: checked the code above, os.Walk will fail in case specDir 1 and 4, so I removed the check.
case 2 and 3 is expected since the app name will be always the last part of path or it will be specDir itself in case it's not inside another dir.

thiguetta:
tested here https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/278/console, using the same example as the description, as seen in the logs you see 24 nginx-sharedv4 replicas being created and validated

PR #395, opened 2020-01-14
PTX-2148 , PTX-2149, PTX-2150, PTX-2151, PTX-2152 , PTX-2153, PTX-2156   Validate volume replicas  placement as per the volume placement stratergy

nareshdesh:
This  pull request contains the following support
1. Validate VSP statting module - containing support for replica affinity case
2. Contains the solution for validating replica affinity cases
3. Contains the replica affinity matchexpression support

Refer PTX-2148 Validate Volume Replica Affinity,  PTX-2153 Volume Replica MatchExpression  and  PTX-2152 Validate VPS

Note that,  further support for validating ReplicaAntiAffinity, Volume Affinity, Volume Anti Affinity is in progress.
The VPS apply spec will undergo change, so that the spec can be applied to any application in the torpedo suite 

**Note: For review purpose, ignore these 2 files for now, since it is going to be changed soon
tests/provision/provisioning_rules.go
tests/provision/provisioning_test.go**


drivers/scheduler/k8s/k8s.go
```diff
@@ -776,6 +811,32 @@ func (k *K8s) substituteNamespaceInPVC(pvc *v1.PersistentVolumeClaim, ns string)
 	}
 }
 
+//Set or disable the placement_strategy in the storageclass
+func (k *K8s) substituteVpsNameInStorageClass(sc *storageapi.StorageClass, vps_map map[string]string) {
+	//	sc.Name = namespaceRegex.ReplaceAllString(sc.Name, ns)
+	for k, v := range sc.Parameters {
+		logrus.Infof("[%v] storaegeclass parameters: %v", k, v)
+		newstr := placement1Regex.ReplaceAllString(v, vps_map["placement-1"])
```
thiguetta: @naresh we already have support to templating in the torpedo, can we try to use that here? if you have any questions I can help you. check here for an example https://github.com/portworx/torpedo/pull/394

tests/common.go
```diff
@@ -127,16 +128,25 @@ func ValidateCleanup() {
 }
 
 // ValidateContext is the ginkgo spec for validating a scheduled context
-func ValidateContext(ctx *scheduler.Context) {
+func ValidateContext(ctx *scheduler.Context, vState int) {
 	ginkgo.Describe(fmt.Sprintf("For validation of %s app", ctx.App.Key), func() {
 		Step(fmt.Sprintf("validate %s app's volumes", ctx.App.Key), func() {
-			ValidateVolumes(ctx)
+			ValidateVolumes(ctx, vState)
```
thiguetta: are we changing vState somewhere? it seems it might be a problem if we're using copy instead of reference

tests/common.go
```diff
@@ -228,23 +254,28 @@ func DeleteVolumesAndWait(ctx *scheduler.Context) {
 }
 
 // ScheduleAndValidate schedules and validates applications
-func ScheduleAndValidate(testname string) []*scheduler.Context {
+func ScheduleAndValidate(testname string, vps_map map[string]string) []*scheduler.Context {
```
thiguetta: instead of using a map can we use ScheduleOptions as a parameter?

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,520 @@
...
+}
+
+var _ = BeforeSuite(func() {
+	InitInstance()
+})
+
+// This test performs VolumePlacementStrategy's replica affinity  of application
+// volume
+var _ = Describe("{ReplicaAffinity}", func() {
+	It("has to schedule app and verify the volume replica affinity", func() {
```
thiguetta: after `It` we need to collect bundle in case of failure, check the example here https://github.com/portworx/torpedo/blob/5644fcce6fb4d4f33db7cb767a2c6317cb26e3bc/tests/basic/basic_test.go#L104

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,520 @@
...
+				TearDownContext(ctx, opts)
+			}
+		}
+	})
+})
+
+// This test performs VolumePlacementStrategy's volume affinity  of application
+// volume
+var _ = Describe("{VolumeAffinity}", func() {
+	It("has to schedule app and verify the volume replica affinity", func() {
```
thiguetta: same as the comment above

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,520 @@
...
+				TearDownContext(ctx, opts)
+			}
+		}
+	})
+})
+
+// This test performs VolumePlacementStrategy's replica & volume affinity  of application
+// volume
+var _ = Describe("{ReplicaVolumeAffinity}", func() {
+	It("has to schedule app and verify the volume replica affinity", func() {
```
thiguetta: same as the comment above

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,520 @@
...
+				TearDownContext(ctx, opts)
+			}
+		}
+	})
+})
+
+// This test performs VolumePlacementStrategy's replica & volume affinity
+// with app scale Up  & Down of application
+var _ = Describe("{ReplicaVolumeAffinityScale}", func() {
+	It("has to schedule app and verify the volume replica affinity", func() {
```
thiguetta: same as the comment above

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,520 @@
...
+				TearDownContext(ctx, opts)
+			}
+		}
+	})
+})
+
+// This test performs VolumePlacementStrategy's replica & volume affinity  of application
+// with volumes pending state
+var _ = Describe("{ReplicaVolumeAffinityPending}", func() {
+	It("has to schedule app and verify the volume replica affinity", func() {
```
thiguetta: same as the comment above

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,520 @@
...
+
+			for _, ctx := range contexts {
+				TearDownContext(ctx, opts)
+			}
+		}
+	})
+})
+
+var _ = Describe("{DefaultRepVolAffinity}", func() {
+	It("has to schedule app and verify the volume replica affinity", func() {
```
thiguetta: same as the comment above

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+	Register(v.name, v,2)
+}
+
+func init() {
+	v := &vpscase14{"case14-T809549 Volume Anti-Affinity 'NotIn'", true}
+	Register(v.name, v,2)
+}
+*/
+
+///*
```
thiguetta: if not using it remove this code

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+	v := &vpscase22{"case22-T871040 Verify statefulset/deployment scale up/down w.r.t replica and volume affinity rules ", true}
+	Register(v.name, v, 4)
+}
+
+/*
+// Pool labeling is pending
+func init() {
+	v := &vpscase23{"case23-T871040 T955476 Replica & Volume Affinity & Anti Affinity ", true}
+	Register(v.name, v)
+}
```
thiguetta: remove if not needed

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+// Test case inits
+//
+
+/*
+ *
+ *     Replica  Affinity and Anti-Affinity related test cases init
+ *
+ */
+///*
+func init() {
```
thiguetta: is there any specific reason why we have init functions define so many times?

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+	//Case description
+	name string
+	// Enabled
+	enabled bool
+}
+
+func (v *vpscase21) GetLabels() ([]labelDict, int) {
+
+	lbldata := []labelDict{}
+	node1lbl := labelDict{"failure-domain.beta.kubernetes.io/px_zone": "east", "failure-domain.beta.kubernetes.io/px_region": "usa"}
```
thiguetta: we use `failure-domain.beta.kubernetes.io/px_zone` a lot, can we use constants for these labels?

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+
+//StorageClass placement_strategy mapping
+func (v *vpscase20) GetScStrategyMap() map[string]string {
+	return map[string]string{"placement-1": "placement-1", "placement-2": "placement-2", "placement-3": ""}
+}
+
+func (v *vpscase20) GetSpec() string {
+
+	var vpsSpec string
+	vpsSpec = `apiVersion: portworx.io/v1beta2
```
thiguetta: can we put it into a spec file like the apps?

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+
+//StorageClass placement_strategy mapping
+func (v *vpscase19) GetScStrategyMap() map[string]string {
+	return map[string]string{"placement-1": "placement-1", "placement-2": "placement-2", "placement-3": ""}
+}
+
+
+func (v *vpscase19) GetSpec() string {
+
+	var vpsSpec string
```
thiguetta: can we add this spec as a file?

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+	//Case description
+	name string
+	// Enabled
+	enabled bool
+}
+
+func (v *vpscase19) GetLabels() ([]labelDict, int) {
+
+	lbldata := []labelDict{}
+	node1lbl := labelDict{"failure-domain.beta.kubernetes.io/px_zone": "east", "failure-domain.beta.kubernetes.io/px_region": "usa"}
```
thiguetta: use constants for labels

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+	//Case description
+	name string
+	// Enabled
+	enabled bool
+}
+
+func (v *vpscase18) GetLabels() ([]labelDict, int) {
+
+	lbldata := []labelDict{}
+	node1lbl := labelDict{"failure-domain.beta.kubernetes.io/px_zone": "east", "failure-domain.beta.kubernetes.io/px_region": "usa"}
```
thiguetta: use constants for labels

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+	//Case description
+	name string
+	// Enabled
+	enabled bool
+}
+
+func (v *vpscase17) GetLabels() ([]labelDict, int) {
+
+	lbldata := []labelDict{}
+	node1lbl := labelDict{"failure-domain.beta.kubernetes.io/px_zone": "east", "failure-domain.beta.kubernetes.io/px_region": "usa"}
```
thiguetta: constant for labels

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1708 @@
...
+	//Case description
+	name string
+	// Enabled
+	enabled bool
+}
+
+func (v *vpscase4) GetLabels() ([]labelDict, int) {
+
+	lbldata := []labelDict{}
+	node1lbl := labelDict{"failure-domain.beta.kubernetes.io/px_zone": "east", "failure-domain.beta.kubernetes.io/px_region": "usa"}
```
thiguetta: I see this code is being repeated a lot, is there any way we can improve it?

tests/common.go
nareshdesh: vState stands for VolumeState, in a test a volume is expected to be in pending state. The test will pass this flag to inform the ValidateVolumes to pass the  test  even if  the volume is in pending state

tests/provision/provisioning_rules.go
nareshdesh: There are at present 24 VPS test cases and each has different spec. 
If we create all the 24 spec file, they will get activated at once instead of  during the  corresponding test case

tests/provision/provisioning_rules.go
nareshdesh: Sure, will make the changes

thiguetta:
@nareshdesh build-ci is failing, fix `https://travis-ci.org/portworx/torpedo/builds/640040713?utm_source=github_status&utm_medium=notification` and make sure `make lint` passes before committing the change

drivers/volume/volume.go
```diff
@@ -50,7 +50,7 @@ type Driver interface {
 
 	// ValidateCreateVolume validates whether a volume has been created properly.
 	// params are the custom volume options passed when creating the volume.
-	ValidateCreateVolume(name string, params map[string]string) error
+	ValidateCreateVolume(appVols map[string]map[string]string, vState int, appName string, appToken string) error
```
thiguetta: too much information here, can we use a wrapper struct to reduce the number of parameters passed?

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1546 @@
...
+	} else {
+		return fmt.Errorf("vps rule category: %d, is not valid", cat)
+	}
+
+	return nil
+}
+
+// GetVpsRules return the list of vps rules
+func GetVpsRules(cat int) map[string]VpsTemplate {
+	if cat == 1 {
```
thiguetta: instead of adding this put the rules in an indexed list like `[]map[string]VpsTemplate`, so you can use `return rules[cat]`

drivers/volume/volume.go
```diff
@@ -129,6 +129,10 @@ type Driver interface {
 	// GetReplicaSets returns the replica sets for a given volume
 	GetReplicaSets(vol *Volume) ([]*api.ReplicaSet, error)
 
+	//GetReplicatNodeSets returns the replica grouped per set for a given volume
```
rodrigc: Delete these commented out lines.  If you need them they should be uncommented.

tests/asg/asg_test.go
```diff
@@ -21,6 +21,7 @@ import (
 
 const (
 	scaleTimeout          = 10 * time.Minute
+	defaultVstate         = 1
```
rodrigc: `defaultVState` constant should not be added

tests/basic/basic_test.go
```diff
@@ -14,6 +14,10 @@ import (
 	. "github.com/portworx/torpedo/tests"
 )
 
+const (
+	defaultVstate = 1
```
rodrigc: `defaultVstate` constant should not be added

tests/common.go
rodrigc: I don't think that changing `ValidateContext()` like this should be done.
It is not obvious what the `vState` variable is, and what it's impact on the function is.

You are passing in defaultInterval which is a context from many places into this function.  You essentially have two functions:

One where defaultInterval is 0,
another where defaultInterval is 1.

I suggest that you create two functions:


 * `ValidateContextXXX1()`
 * `ValidateContextXXX2()`

replace `XXX1` and `XXX2` with obvious names as to how the functions are behaving.

Then replace ValidateContext() with ValidateConextXXX?() depending on what you need.

tests/common.go
```diff
@@ -155,39 +165,58 @@ func ValidateContext(ctx *scheduler.Context) {
...
+					}
 				})
 			}
 		})
 	})
 }
 
 // ValidateVolumes is the ginkgo spec for validating volumes of a context
-func ValidateVolumes(ctx *scheduler.Context) {
+func ValidateVolumes(ctx *scheduler.Context, vState int) {
```
rodrigc: Looks like we need two functions `ValidateVolumes()` functions instead of a single function with a `vState` flag.

tests/reboot/reboot_test.go
```diff
@@ -13,6 +13,10 @@ import (
 	. "github.com/portworx/torpedo/tests"
 )
 
+const (
```
rodrigc: `defaultVstate` should not be added

tests/sched/sched_test.go
```diff
@@ -14,6 +14,10 @@ import (
 	"math/rand"
 )
 
+const (
```
rodrigc: `defaultVstate` should not be added

tests/volume_ops/volume_ops_test.go
```diff
@@ -15,6 +15,10 @@ import (
 	"time"
 )
 
+const (
```
rodrigc: `defaultVstate` should not be added

tests/common.go
nareshdesh: @rodrigc , creating 2 functions might create problem in case there is a future development in validateVolume and the changes are not updated in both the functions. 
I can make the following change, instead of passing as argument to the function, can set it in `Options` of ctx and use it inside ValidateVolume directly, if the option flag of vState is set

nareshdesh: Moved the flag as part of ScheduleOptions and as a result  there is no need to pass it as parameter to this function

tests/common.go
nareshdesh: Used the ScheduleOption to pass the vps related data, as a result , passing of this parameter is not required

tests/provision/provisioning_rules.go
nareshdesh: Will be enabling this testcase soon

tests/provision/provisioning_rules.go
nareshdesh: The init function is adding  each test case to the list of test cases to be tested. 
This will changed, if the changes as per your comment below is incorporated
`instead of adding this put the rules in an indexed list like var rules []map[string]interface{}, so you can use return rules[cat]`

tests/asg/asg_test.go
nareshdesh: removed it

tests/basic/basic_test.go
nareshdesh: removed it

tests/common.go
nareshdesh: Changed the way of passing this flag , it is now sent as part of ScheduleOptions

tests/reboot/reboot_test.go
nareshdesh: removed it

tests/volume_ops/volume_ops_test.go
nareshdesh: removed it

tests/sched/sched_test.go
nareshdesh: removed it

tests/provision/provisioning_rules.go
nareshdesh: For each test case, the  kubernetes nodes are labelled differently  and hence you are seeing this code with different values or  different combination of entry

tests/provision/provisioning_test.go
nareshdesh: Incorporated the changes

drivers/scheduler/k8s/k8s.go
nareshdesh: @thiguetta  can you review it again

tests/provision/provisioning_rules.go
nareshdesh: Changed these to constants

tests/provision/provisioning_rules.go
nareshdesh: Changed to constants

tests/provision/provisioning_rules.go
nareshdesh: Changed to constants

drivers/volume/volume.go
nareshdesh: Modified slightly, can you please review  again

tests/provision/provisioning_rules.go
nareshdesh: Made the changes as per the suggestion

drivers/volume/volume.go
nareshdesh: Removed the commented lines

drivers/scheduler/k8s/k8s.go
nareshdesh: made changes around this code area, can you please review it again

tests/provision/provisioning_rules.go
nareshdesh: Modified the init function, please review it again

tests/provision/provisioning_rules.go
nareshdesh: changed to constants

tests/provision/provisioning_rules.go
nareshdesh: changed to constants

tests/provision/provisioning_rules.go
nareshdesh: changed to constants

tests/provision/provisioning_rules.go
nareshdesh: changed to constants

tests/provision/provisioning_rules.go
harsh-px: It's okay for each test case to have it's own labels

thiguetta:
@nareshdesh l suggest breaking this test into files to be more readable, like move all the rules to a file called `rules.go`, as they are under the same package you don't need much change, just cut & paste

tests/provision/provisioning_test.go
```diff
@@ -107,11 +107,12 @@ var _ = Describe("{ReplicaAffinity}", func() {
 			for _, ctx := range contexts {
 				TearDownContext(ctx, opts)
 			}
+
+			JustAfterEach(func() {
+				AfterEachTest(contexts)
+			})
 		}
 	})
-	JustAfterEach(func() {
```
thiguetta: don't move this code, this has to be outside `It()`

tests/provision/provisioning_test.go
```diff
@@ -180,11 +181,11 @@ var _ = Describe("{VolumeAffinity}", func() {
 			for _, ctx := range contexts {
 				TearDownContext(ctx, opts)
 			}
+			JustAfterEach(func() {
+				AfterEachTest(contexts)
+			})
 		}
 	})
-	JustAfterEach(func() {
```
thiguetta: same above

tests/provision/provisioning_test.go
```diff
@@ -253,11 +254,11 @@ var _ = Describe("{ReplicaVolumeAffinity}", func() {
...
```
thiguetta: same above

tests/provision/provisioning_test.go
```diff
@@ -363,11 +364,11 @@ var _ = Describe("{ReplicaVolumeAffinityScale}", func() {
...
```
thiguetta: same above

tests/provision/provisioning_test.go
nareshdesh: Made changes as suggested

tests/provision/provisioning_test.go
nareshdesh: Made changes as suggested

tests/provision/provisioning_test.go
nareshdesh: Made changes as suggested

tests/provision/provisioning_test.go
nareshdesh: Made changes as suggested

drivers/volume/volume.go
```diff
@@ -54,7 +54,7 @@ type Driver interface {
 
 	// ValidateCreateVolume validates whether a volume has been created properly.
 	// params are the custom volume options passed when creating the volume.
-	ValidateCreateVolume(name string, params map[string]string) error
```
rodrigc: This PR has improved a lot.  I'm mostly OK with what is in it, but I am a bit hesitant about
this change to `ValidateCreateVolume()`.  Before your PR, this function was fairly simple in that it took in a name and params, and then validated if a volume was up/proper.

After your change you internally changed it to loop over many volumes.  This is a major semantic change to this function.

Is it possible for you to:

1.  Keep ValidateCreateVolume() the way it was before
2.  Create a new function which takes an array or map of volumes, and in turn calls ValidateCreateVolume() in a loop?

This would ease my mind, and help me push this PR into the tree.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1514,19 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
...
 
+	for _, spec := range ctx.App.SpecList {
+		t := func() (interface{}, bool, error) {
+			err := k.destroyVpsObjects(spec, ctx.App)
+			if err != nil {
+				return nil, true, err
+			}
+			return nil, false, nil
+		}
+		pods, err := task.DoRetryWithTimeout(t, k8sDestroyTimeout, DefaultRetryInterval)
```
harsh-px: Why is a delete VPS task returning pods and appending to podList below?

drivers/scheduler/k8s/k8s.go
```diff
@@ -1461,6 +1514,19 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
 		}
 	}
 
+	for _, spec := range ctx.App.SpecList {
+		t := func() (interface{}, bool, error) {
+			err := k.destroyVpsObjects(spec, ctx.App)
```
harsh-px: we don't need a dedicated method `destroyVpsObjects`.

We would add a dedicated method where there is a group of objects in the same category that we need toc clean up.

drivers/scheduler/k8s/k8s.go
```diff
@@ -830,6 +868,21 @@ func (k *K8s) substituteNamespaceInPVC(pvc *v1.PersistentVolumeClaim, ns string)
 	}
 }
 
+func (k *K8s) addVpsVolumeLabel(pvc *v1.PersistentVolumeClaim, vlabel string) {
```
harsh-px: looks like a dead function

drivers/scheduler/k8s/k8s.go
```diff
@@ -830,6 +868,21 @@ func (k *K8s) substituteNamespaceInPVC(pvc *v1.PersistentVolumeClaim, ns string)
...
+	//pvc.Metadatas.Labels["appvps"] = vlabel
+	logrus.Infof("PVC label update label:%v  pvc:%v", vlabel, pvc)
+}
+
+//Set or disable the placement_strategy in the storageclass
+func (k *K8s) substituteVpsNameInStorageClass(sc *storageapi.StorageClass, vpsname string) {
+	//	sc.Name = namespaceRegex.ReplaceAllString(sc.Name, ns)
+	if sc.Parameters["aggregation_level"] == "2" || sc.Parameters["aggregation_level"] == "3" {
```
harsh-px: why are we doing this? We do support placement_strategy for aggregation_level 2 and 3. If we don't want to test

placement for aggregation_level = 2 or 3 for a reason, we need to handle that at a spec level and not here

drivers/scheduler/k8s/k8s.go
```diff
@@ -671,6 +705,8 @@ func (k *K8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.
 		logrus.Infof("Setting provisioner of %v to %v", obj.Name, volume.GetStorageProvisioner())
 		obj.Provisioner = volume.GetStorageProvisioner()
 
+		//VPS Name set
+		k.substituteVpsNameInStorageClass(obj, vpsmap)
```
harsh-px: why do we need this? For placement, we can simply check in the volume labels if it's present. There are use cases where the placement_strategy will only be in the PVC labels and not SC params

drivers/scheduler/k8s/k8s.go
```diff
@@ -648,6 +674,14 @@ func (k *K8s) createNamespace(app *spec.AppSpec, namespace string, options sched
 func (k *K8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.AppSpec,
 	options scheduler.ScheduleOptions) (interface{}, error) {
 
+	var vpsmap string
+	if options.VpsParameters != nil && options.VpsParameters.Enabled {
+		vpsmap = app.Key
+	} else {
+
+		vpsmap = ""
+	}
```
harsh-px: this else is not needed. `var vpsmap string` means the string is empty.

drivers/scheduler/k8s/k8s.go
```diff
@@ -96,11 +98,14 @@ const (
 )
 
 var (
+	// replace VPS name with the app name
+	placement1Regex = regexp.MustCompile("{VPS_NAME}")
```
harsh-px: use a better name that `placement1`

drivers/scheduler/k8s/k8s.go
```diff
@@ -26,8 +26,10 @@ import (
 	"github.com/portworx/sched-ops/k8s/rbac"
 	"github.com/portworx/sched-ops/k8s/storage"
 	"github.com/portworx/sched-ops/k8s/stork"
+	talisman "github.com/portworx/sched-ops/k8s/talisman"
```
harsh-px: remove extra line

drivers/scheduler/scheduler.go
```diff
@@ -46,6 +46,16 @@ func (in *Context) GetID() string {
 	return in.App.GetID(in.UID)
 }
 
+// VpsParameters , volume placement strategy parameters
+type VpsParameters struct {
```
harsh-px: instead of this, let's just use the placement spec in the schedule options to be consistent with how autopilot uses. With that, all extra schedule options will get converged under the schedule options.

drivers/volume/common.go
```diff
@@ -66,9 +66,12 @@ func (d *DefaultDriver) RecoverDriver(n node.Node) error {
...
 }
 
-// ValidateCreateVolume validates whether a volume has been created properly.
+// ValidateCreateVolume validates whether volumes of the app has been created properly.
+// appVols: dictionary of all volumes of the app
+// appName: The app name whose volumes needs to be validated
+// appToken: The token used to get volume details
 // params are the custom volume options passed when creating the volume.
-func (d *DefaultDriver) ValidateCreateVolume(name string, params map[string]string) error {
+func (d *DefaultDriver) ValidateCreateVolume(appVols map[string]map[string]string, appName string, appToken string) error {
```
harsh-px: keep the older method as is. Create a separate API to validate placement strategies. The test should first call validate create volume for each volume. And then we call validate placement strategy and provide all volumes to that method.

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
+		}
+		logrus.Infof("Successfully inspected volume: %v (%v)", vol.Locator.Name, vol.Id)
+
+	}
+
+	return nil
+}
+
+// Validate the volume replicas as per VolumePlacementStrategy rule applied to the volume
+func (d *portworx) ValidateVps(vol *api.Volume, appVols map[string]map[string]string, volNodes []api.StorageNode, nodeReplMap map[string][]*api.Volume, appVolNodes map[string]api.StorageNode) error {
```
harsh-px: this function has too many arguments

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
+
+	logrus.Debugf("Volume details: %v ===\n (%v) Volumes per Node List:%v, all nodes of the App's Volumes: %v ", vol, appVols, nodeReplMap, appVolNodes)
+	logrus.Infof("Validate VPS  for Volume:%v ,  VPS Rule :(%v)", vol.Id, vol.Spec.GetPlacementStrategy())
+
+	if vpsrule := vol.Spec.GetPlacementStrategy(); vpsrule != nil {
```
harsh-px: why do we need this check? Just check if placement_strategy is present in the volume labels. If yes, placement is enabled

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
+
+		// Get Volume Labels
+		volLabels := vol.Spec.GetVolumeLabels()
```
harsh-px: no need to save into additional variable. Use below directly

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
+		volVpsRule, err := k8sTalisman.GetVolumePlacementStrategy(volLabels["placement_strategy"])
+		logrus.Debugf("====Volume details-5: %v ===\n Placement strategy (%v)", vol.Id, volVpsRule)
+
+		if err != nil {
+			return err
+		}
+
+		// For each VPS  group rules
+		// Group1 -Replica Affinity Group
+		if volVpsRule.Spec.ReplicaAffinity != nil {
```
harsh-px: nil check not needed, range handles that. Same for all other places in review.

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
 			}
-		case api.SpecPriorityAlias:
-			// Since IO priority isn't guaranteed, we aren't validating it here.
-		case api.SpecSnapshotInterval:
-			if requestedSpec.SnapshotInterval != vol.Spec.SnapshotInterval {
-				return errFailedToInspectVolume(volumeName, k, requestedSpec.SnapshotInterval, vol.Spec.SnapshotInterval)
+		}
+
+	} else {
+		logrus.Infof("Validate VolumePlacementStrategy,  Volume (%v) doesnot have any VPS rule applied to it", vol.Id)
```
harsh-px: No need to print this

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
+							}
+						}
+
+						// Check node value is less than spec value
+						if itval <= mValue {
+							return false, &ErrFailedToInspectVolume{
+								Cause: fmt.Sprintf("VolumeAntiAffinity with matchExpression Operator:'%v', is placed on node:%v , having node value(%v) less than Spec value (%v)  for key:(%v:%v)", matchExp.Operator, node.Id, itval, mValue, matchExp.Key, tval),
+							}
+						}
+						logrus.Infof("VolumeAntiAffinity with matchExpression Operator:'%v', is placed on node:%v, having n
```
harsh-px: volNodes should not be needed. You already have the vol object which has all the replica node and pool info for the volume

harsh-px: cannot use node. It's a package name

harsh-px: variable status conflicts with a package name

harsh-px: overall the logic could be simplified. No need of a second pass here. When you iterate over nodeGrpList above, just make sure the match is within a single node group

harsh-px: we should be able to use just one function to test replica affinity and anti-affinity. Based on the type, switch some logic

tests/common.go
```diff
@@ -146,14 +146,29 @@ func ValidateCleanup() {
 // ValidateContext is the ginkgo spec for validating a scheduled context
 func ValidateContext(ctx *scheduler.Context) {
 	ginkgo.Describe(fmt.Sprintf("For validation of %s app", ctx.App.Key), func() {
+		var vState int
+		if ctx.ScheduleOptions.VpsParameters != nil {
+			vState = ctx.ScheduleOptions.VpsParameters.Vstate
+		} else {
+			vState = 1
```
harsh-px: simplify by just setting vState to 1 when declaring the variable

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1334 @@
...
+	mediaType      = "media_type"
+	liops          = "iops"
+	llatency       = "latency"
+	pxZone         = "failure-domain.beta.kubernetes.io/px_zone"
+	pxRegion       = "failure-domain.beta.kubernetes.io/px_region"
+	kubeZone       = "failure-domain.beta.kubernetes.io/zone"
+	ruleCategories = 6
+)
+
+type labelDict map[string]interface{}
```
harsh-px: reduce the number of lines here to remove the inline yamls and instead create the placement golang struct using helper functions. See how autopilot declares this using in the variable autopilotruleBasicTestCases

tests/provision/provisioning_rules.go
```diff
@@ -0,0 +1,1334 @@
...
+}
+
+//# Case-1--enforcemnt: Required
+func (v *vpscase1) GetLabels() ([]labelDict, int) {
+
+	lbldata := []labelDict{}
+	node1lbl := labelDict{mediaType: mediaSsd}
+	node2lbl := labelDict{mediaType: mediaSata}
+	node3lbl := labelDict{mediaType: mediaSsd}
+	node4lbl := labelDict{mediaType: mediaSsd}
```
harsh-px: instead of append, just initial the entire data structure. Makes it easier to read. Same for all places.

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,645 @@
...
+		var vpsSpec string
+		vpsRules := GetVpsRules(1)
+
+		for vkey, vrule := range vpsRules {
+			contexts = make([]*scheduler.Context, 0)
+			var lblData []labelDict
+			var setLabels int
+			Step("get nodes and set labels: "+vkey, func() {
+				lblData, setLabels = getTestLabels(vrule.GetLabels)
+				RemoveNodeLabels(lblData)
```
harsh-px: return err from RemoveNodeLabels and check here. Same for all places in review.

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,645 @@
...
+				if setLabels == 1 {
```
harsh-px: remove this setLabels logic and simplify it. A test case has labels, just apply it

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,645 @@
...
+					lblnode := SetNodeLabels(lblData)
```
harsh-px: return and check err here

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,645 @@
...
+					logrus.Debug("Nodes containing label", lblnode)
+					Expect(lblnode).NotTo(BeEmpty())
+				}
+			})
+
+			Step("rules of volume placement: "+vkey, func() {
+				vpsSpec = getVpsSpec(vrule.GetSpec)
```
harsh-px: no need to the getVpsSpec() function. Call vrule.GetSpec directly

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,645 @@
...
+			})
+
+			Step("launch application with new vps specs :"+vkey, func() {
+				applyVpsSpec(vpsSpec)
+				logrus.Debugf("Spec Dir to rescan: %v", Inst().SpecDir)
+				Inst().S.RescanSpecs(Inst().SpecDir)
```
harsh-px: simplify this. Don't create new files. Pass list of specs in scheduler options itself. See how we do it for autopilot rule

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,645 @@
...
+
+				// Set VPS enabled flag
+				VpsMap := &scheduler.VpsParameters{
```
harsh-px: make local variable

tests/provision/provisioning_test.go
```diff
@@ -0,0 +1,645 @@
...
+				for _, ctx := range contexts {
+					ValidateContext(ctx)
+				}
+
+			})
+
+			opts := make(map[string]bool)
+			opts[scheduler.OptionsWaitForResourceLeakCleanup] = true
+
+			vrule.CleanVps() //TODO: function arg for cleaning up the testcase environment
```
harsh-px: CleanVps not needed. The cleanup should be generic where it removes all labels with a known prefix

drivers/volume/portworx/portworx.go
harsh-px: all of these validation functions can be private (small `v`).

harsh-px: arg `volNodes` is not being used

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
 
-	logrus.Infof("Successfully inspected volume: %v (%v)", vol.Locator.Name, vol.Id)
-	return nil
+	return false, &ErrFailedToInspectVolume{
+		Cause: fmt.Sprintf("ReplicaAntiAffinity with matchExpression Operator:'%v' could not be validated for node:%v  having label key:(%v)", matchExp.Key, node.Id, matchExp.Key),
+	}
 }
 
+// VolumeMatchExpression
+func (d *portworx) VolumeMatchExpression(vol *api.Volume, matchExp *talisman_v1beta1.LabelSelectorRequirement, nodeReplList map[string][]*api.Volume, nodeGrp map[string]api.StorageNode) (bool, error) {
```
harsh-px: VolumeMatchExpression doesn't need to be public

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
+		switch matchExp.Operator {
+		case "NotIn":
+			//for each node in the node group
+			for _, node := range nodeGrp {
+
+				logrus.Debugf("VolumeMatchExpression: volume replicas on node %v: %v", node.Id, nodeReplList[node.Id])
+				// for each volume on the node
+				for _, nodeVol := range nodeReplList[node.Id] {
+					// Check whether node is having the same volume or clone
+					if nodeVol.Id == vol.Id || nodeVol.Source.Parent != "" {
```
harsh-px: this parent logic is getting repeated at many places. Just curate a list once and remove all clones.

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
 
+// VolumeMatchExpression
+func (d *portworx) VolumeMatchExpression(vol *api.Volume, matchExp *talisman_v1beta1.LabelSelectorRequirement, nodeReplList map[string][]*api.Volume, nodeGrp map[string]api.StorageNode) (bool, error) {
+	logrus.Infof("VolumeMatchExpression started nodeId: %v matchExpression:%v ", vol.Id, matchExp)
+
+	logrus.Debugf("VolumeMatchExpression rule:%v  mkey:%v mOperator:%v mValues:%v nodeReplList:%v, nodeGrp:%v", vol, matchExp.Key, matchExp.Operator, matchExp.Values, nodeReplList, nodeGrp)
+	if vol != nil {
+		// Check for Operators: In, Exists, NotIn, DoesNotExist
+		switch matchExp.Operator {
+		case "NotIn":
```
harsh-px: lot of duplicate logic for variable operators. We need to have a common piece of code to create a list of current and expected values and then later on add a switch case based on operator to make the corresponding check.

drivers/volume/portworx/portworx.go
```diff
@@ -510,174 +514,1520 @@ func (d *portworx) RecoverDriver(n node.Node) error {
...
-		case api.SpecBlockSize:
-			if requestedSpec.BlockSize != vol.Spec.BlockSize {
-				return errFailedToInspectVolume(volumeName, k, requestedSpec.BlockSize, vol.Spec.BlockSize)
+		}
+		// Group3 -Volume Affinity Group
+		if volVpsRule.Spec.VolumeAffinity != nil {
+			for _, vRule := range volVpsRule.Spec.VolumeAffinity {
+
+				logrus.Infof("Validate Volume Affinity Rule Vol:%v === VA: Spec %v", vol.Id, vRule)
+				err = d.ValidateVolumeAffinity(vol, vRule, volNodes, nodeReplMap, appVolNodes)
```
harsh-px: I don't see the part where we are looking at multiple volumes together and then checking if they satisfied the affinity requirements. `ValidateVolumeAffinity` is only being called on a single volume.

PR #397, opened 2020-01-23
PTX-2119 fix loadbalancer session affinity

thiguetta:


thiguetta:
> can you summarize how this was causing an issue?

without session affinity, we can't guarantee that load-generator will send consistently load to wordpress (which is session dependent) in case loadbalancer picks a different node than the previous requests the test breaks

thiguetta:
tested here https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/282/

PR #399, opened 2020-01-27
PTX-2235 fix torpedo not to wait when not force kill pwx driver

thiguetta:
Signed-off-by: Thiago Gonzaga <thi_gonzaga@yahoo.com.br>


drivers/volume/portworx/portworx.go
```diff
@@ -717,6 +717,8 @@ func (d *portworx) StopDriver(nodes []node.Node, force bool) error {
 				logrus.Warnf("failed to run cmd : %s. on node %s err: %v", pxCrashCmd, n.Name, err)
 				return err
 			}
+			logrus.Infof("Sleeping for %v for volume driver to go down.", waitVolDriverToCrash)
```
harsh-px: nit: rename this to `"Sleeping for %v for px-storage to get killed."`

PR #400, opened 2020-01-27
PTX-1641 Fix Init() methods for aks and gke drivers

rodrigc:
The `init()` method is called when the package is initialized: https://golang.org/doc/effective_go.html#init

However, we have our own Init() method which is called at runtime.  We need to move the SSH initialization into the Init() method so that `useSSH()`, which evaluates the `TORPEDO_SSH_KEY` and `TORPEDO_SSH_PASSWORD` variables can be evaluated properly.

### DIFF
--- a/drivers/node/aks/aks.go
+++ b/drivers/node/aks/aks.go
@@ -29,6 +29,7 @@ func (a *aks) String() string {
 }
 
 func (a *aks) Init() error {
+	a.SSH.Init()
 
 	instanceGroup := os.Getenv("INSTANCE_GROUP")
 	if len(instanceGroup) != 0 {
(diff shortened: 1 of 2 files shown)


rodrigc:
Testing here: https://jenkins.portworx.co/job/Torpedo/job/craig-tp-nextpx-aks/1/

Ran into quota limit on AKS

drivers/node/aks/aks.go
```diff
@@ -69,11 +70,8 @@ func (a *aks) GetASGClusterSize() (int64, error) {
 }
 
 func init() {
-
-	SSHDriver := ssh.SSH{}
-	SSHDriver.Init()
 	g := &aks{
```
mborodin-px: nit: rename this variable to 'a'

PR #403, opened 2020-01-30
refactor autopilot changes and other changes

harsh-px:
- Split calculation of pool and volume sizes into separate methods
- extra common pieces of autopilot tests into common methods
- rename tests
- fix calcuation of pool size check to factor in size of disks in the pool
- fix calculation of volume size to consider max size
- Add trigger callback to DeleteTasks to control when tasks are deleted


drivers/volume/portworx/portworx_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	type testCase struct {
+		rule           apapi.AutopilotRule
+		pool           node.StoragePool
+		node           node.Node
+		expectedSize   uint64
+		expectedToFail bool
+	}
+
+	testCases := []testCase{
```
trierra: no volume UT yet?

harsh-px: didn't make any logic changes to the volume function.

PR #404, opened 2020-01-30
Use portworx-service (if available) while checking if a node is metad…

vinayakshnd:
…ata node


thiguetta:
tested here https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/290/console and here https://jenkins.portworx.co/job/Torpedo/job/tp-nextpx-ocp4.3-op/16/

PR #405, opened 2020-01-30
[PTX-2178] Add termination signal handler for esload image

saheienko:
es-load container took too long to terminate continuously throwing an rpc error: code = DeadlineExceeded desc = context deadline exceeded. Torpedo has 2 minutes wait timeout for terminating containers, in this case it took over 4 minutes to kill it.

Add SIGTERM handler for esload image to fix it.

Signed-off-by: Serhii Aheienko <serhii.aheienko@gmail.com>


thiguetta:
not sure if we need this handler, first we need to understand why it is taking time, maybe some slow IO or it may even cover a bug

harsh-px:
> not sure if we need this handler, first we need to understand why it is taking time, maybe some slow IO or it may even cover a bug

@saheienko could you answer this?

saheienko:
Sure. Here is a pod termination flow: https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods

The current version of esload image ignores the SIGTERM signal so after deleting the esload pod it will be alive at least 30s and then will be killed anyway.

With this change esload image can recognize the SIGTERM signal so termination time will be decreased.

PR #407, opened 2020-01-31
PTX-2225 Add test to restart autopilot when pool expand starts

harsh-px:


drivers/volume/portworx/portworx.go
```diff
@@ -1314,17 +1321,18 @@ func (d *portworx) setDriver() error {
...
-	if err == nil && endpoint != "" {
-		if err = d.testAndSetEndpoint(endpoint); err == nil {
-			d.refreshEndpoint = false
-			return nil
-		}
-		logrus.Infof("testAndSetEndpoint failed for %v: %v", endpoint, err)
-	} else if err != nil && len(node.GetWorkerNodes()) == 0 {
-		return err
-	}
+	/*
```
vilasd: why we have commented code here ?

harsh-px: good catch. fixed. had commented for dev test to save time.

PR #408, opened 2020-01-31
PTX-2226 Added test for PVC and storage pool resize at the same time

mborodin-px:
* added test for PVC and storage pool resize at the same time
* merged all autopilot tests into one file


pkg/aututils/aututils.go
```diff
@@ -32,6 +32,40 @@ const (
 	StorageSpecAction = "openstorage.io.action.storagepool/expand"
 )
 
+// PvcRuleByUsageCapacity returns an autopilot pvc expand rule that uses usage of pvc size
+func PvcRuleByUsageCapacity(usage, scalePercentage int, maxSize string) apapi.AutopilotRule {
+	apRuleObject := apapi.AutopilotRule{
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name: fmt.Sprintf("pvc-usage-%d-scale-%d-maxsize-%s", usage, scalePercentage, maxSize),
```
harsh-px: add maxsize in the rule name only if user provided it

tests/autopilot/autopilot_test.go
```diff
@@ -0,0 +1,458 @@
...
+
+// This testsuite is used for performing basic scenarios with Autopilot rules where it
+// schedules apps and wait until workload is completed on the volumes. Restarts volume
+// driver and validates PVC sizes of the volumes
+var _ = Describe(fmt.Sprintf("{%sVolumeDriverDown}", testSuiteName), func() {
+
+	It("has to fill up the volume completely, resize the volume, validate and teardown apps", func() {
+		var err error
+		var contexts []*scheduler.Context
+		testName := strings.ToLower(fmt.Sprintf("%sBasic", testSuiteName))
```
trierra: {%sVolumeDriverDown} here too

tests/autopilot/autopilot_test.go
```diff
@@ -0,0 +1,458 @@
...
+					})
+					Expect(err).NotTo(HaveOccurred())
+					Expect(context).NotTo(BeEmpty())
+					contexts = append(contexts, context...)
+				}
+			}
+		})
+		Step("wait until workload completes on volume", func() {
+			for _, ctx := range contexts {
+				err = Inst().S.WaitForRunning(ctx, workloadTimeout, retryInterval)
```
trierra: ValidateContext(ctx *scheduler.Context)  from common.go instead. It validates volumes as well, so you can skip the following step

tests/autopilot/autopilot_test.go
```diff
@@ -0,0 +1,458 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		Step("schedule applications", func() {
+			for i := 0; i < Inst().ScaleFactor; i++ {
+				for id, apRule := range autopilotruleBasicTestCases {
+					taskName := fmt.Sprintf("%s-%d-aprule%d", testName, i, id)
+					apRule.Name = fmt.Sprintf("%s-%d", apRule.Name, i)
+					labels := map[string]string{
+						"autopilot": pvcApRule.Name,
```
harsh-px: this doesn't compile

harsh-px: pvcApRule isn't defined

tests/autopilot/autopilot_test.go
```diff
@@ -0,0 +1,458 @@
...
+		})
+
+	})
+})
+
+// This testsuite is used for performing basic scenarios with Autopilot rules where it
+// schedules apps and wait until workload is completed on the volumes and then validates
+// PVC sizes of the volumes and sizes of storage pools
+var _ = Describe(fmt.Sprintf("{%sPvcAndPoolExpand}", testSuiteName), func() {
+	It("has to fill up the volume completely, resize the storage pool(s), validate and teardown apps", func() {
```
harsh-px: `resize the storage pool(s)` => `resize the volumes and storage pool(s)`

tests/autopilot/autopilot_test.go
```diff
@@ -0,0 +1,458 @@
...
+		var contexts []*scheduler.Context
+
+		testName := strings.ToLower(fmt.Sprintf("%sPvcAndPoolExpand", testSuiteName))
+		pvcApRules := []apapi.AutopilotRule{
```
harsh-px: pvcApRules is not used

pkg/aututils/aututils.go
trierra: if you use Gi for max size, it would cause the error like
```Failed to create namespace: aut-postgres-volresizeonprem-pvc-total-10Gi  a DNS-1123 label must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character ```

File: tests/autopilot/autopilot_test.go [hunk 7]
```diff
@@ -0,0 +1,458 @@
...
+	testSuiteName            = "AUT"
+	workloadTimeout          = 5 * time.Hour
+	retryInterval            = 30 * time.Second
+	unscheduledResizeTimeout = 10 * time.Minute
+)
+
+var autopilotruleBasicTestCases = []apapi.AutopilotRule{
+	aututils.PvcRuleByUsageCapacity(50, 50, ""),
+	aututils.PvcRuleByUsageCapacity(90, 50, ""),
+	aututils.PvcRuleByUsageCapacity(50, 50, "21474836480"),
```
harsh-px: Why is `21474836480` in bytes. Since this is a helper function argument, we should use something like 20Gi here since this is the supported `maxsize` argument format. Will need changes in torpedo where this gets parsed.

We can do that after this PR.

PR #410, opened 2020-02-01
PTX-2227: added test to verify proper alerts and events are raised after each  resize attempt (successful case only)

trierra:
* Added events collector to the scheduler, which stores all events during the torpedo lifetime

### DIFF
--- a/drivers/scheduler/dcos/dcos.go
+++ b/drivers/scheduler/dcos/dcos.go
@@ -74,6 +74,19 @@ func (d *dcos) String() string {
 	return SchedName
 }
 
+// GetEvents dumps events from event storage
+func (d *dcos) GetEvents() map[string][]scheduler.Event {
+	return nil
+}
+
+// ValidateAutopilotEvents validates events for PVCs injected by autopilot
+func (d *dcos) ValidateAutopilotEvents(ctx *scheduler.Context) error {
+	return &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "ValidateAutopilotEvents()",
+	}
+}
+
 func (d *dcos) ParseSpecs(specDir, storageProvisioner string) ([]interface{}, error) {
 	fileList := []string{}
 	if err := filepath.Walk(specDir, func(path string, f os.FileInfo, err error) error {
@@ -367,7 +380,7 @@ func (d *dcos) GetVolumeParameters(ctx *scheduler.Context) (map[string]map[strin
 	return result, nil
 }
 
-func (d *dcos) InspectVolumes(ctx *scheduler.Context, timeout, retryInterval time.Duration) error {
+func (d *dcos) ValidateVolumes(ctx *scheduler.Context, timeout, retryInterval time.Duration) error {
 	inspectDockerVolumeFunc := func(volName string, _ map[string]string) error {
 		t := func() (interface{}, bool, error) {
 			out, err := d.dockerClient.VolumeInspect(context.Background(), volName)
@@ -419,7 +432,7 @@ func (d *dcos) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 }
 
 func (d *dcos) ResizeVolume(cxt *scheduler.Context, configMap string) ([]*volume.Volume, error) {
-	//TODO implement this method
+	// TODO implement this method
 	return nil, &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "ResizeVolume()",
@@ -477,7 +490,7 @@ func (d *dcos) Describe(ctx *scheduler.Context) (string, error) {
 }
 
 func (d *dcos) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string]int32) error {
-	//TODO implement this method
+	// TODO implement this method
 	return &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "ScaleApplication()",
@@ -485,7 +498,7 @@ func (d *dcos) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[strin
 }
 
 func (d *dcos) GetScaleFactorMap(ctx *scheduler.Context) (map[string]int32, error) {
-	//TODO implement this method
+	// TODO implement this method
 	return nil, &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "GetScaleFactorMap()",
@@ -493,7 +506,7 @@ func (d *dcos) GetScaleFactorMap(ctx *scheduler.Context) (map[string]int32, erro
 }
 
 func (d *dcos) StopSchedOnNode(node node.Node) error {
-	//TODO implement this method
+	// TODO implement this method
 	return &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "StopSchedOnNode()",
@@ -501,7 +514,7 @@ func (d *dcos) StopSchedOnNode(node node.Node) error {
 }
 
 func (d *dcos) StartSchedOnNode(node node.Node) error {
-	//TODO implement this method
+	// TODO implement this method
 	return &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "StartSchedOnNode()",
@@ -516,7 +529,7 @@ func (d *dcos) RescanSpecs(specDir string) error {
 }
 
 func (d *dcos) PrepareNodeToDecommission(n node.Node, provisioner string) error {
-	//TODO implement this method
+	// TODO implement this method
 	return &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "PrepareNodeToDecommission()",
@@ -578,6 +591,13 @@ func (d *dcos) GetSpecAppEnvVar(ctx *scheduler.Context, key string) string {
 	// TODO implement this method
 	return ""
 }
+
+// GetWorkloadSizeFromAppSpec gets workload size from an application spec
+func (d *dcos) GetWorkloadSizeFromAppSpec(ctx *scheduler.Context) (uint64, error) {
+	// TODO: not implemented
+	return 0, nil
+}
+
 func init() {
 	d := &dcos{}
 	scheduler.Register(SchedName, d)
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -12,11 +12,13 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"text/template"
 	"time"
 
 	snapv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	apapi "github.com/libopenstorage/autopilot-api/pkg/apis/autopilot/v1alpha1"
+	"github.com/libopenstorage/openstorage/pkg/units"
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/portworx/sched-ops/k8s/apps"
 	"github.com/portworx/sched-ops/k8s/autopilot"
@@ -26,7 +28,6 @@ import (
 	"github.com/portworx/sched-ops/k8s/rbac"
 	"github.com/portworx/sched-ops/k8s/storage"
 	"github.com/portworx/sched-ops/k8s/stork"
-
 	"github.com/portworx/sched-ops/task"
 	"github.com/portworx/torpedo/drivers/node"
 	"github.com/portworx/torpedo/drivers/scheduler"
@@ -46,7 +47,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 const (
@@ -72,12 +76,12 @@ const (
 	volDirCleanupTimeout   = 5 * time.Minute
 	k8sObjectCreateTimeout = 2 * time.Minute
 	k8sDestroyTimeout      = 2 * time.Minute
-	//FindFilesOnWorkerTimeout timeout for find files on worker
+	// FindFilesOnWorkerTimeout timeout for find files on worker
 	FindFilesOnWorkerTimeout = 1 * time.Minute
 	deleteTasksWaitTimeout   = 3 * time.Minute
-	//DefaultRetryInterval  Default retry interval
+	// DefaultRetryInterval  Default retry interval
 	DefaultRetryInterval = 10 * time.Second
-	//DefaultTimeout default timeout
+	// DefaultTimeout default timeout
 	DefaultTimeout = 2 * time.Minute
 
 	defaultTriggerCheckInterval          = 5 * time.Second
@@ -110,16 +114,17 @@ var (
 	k8sRbac            = rbac.Instance()
 )
 
-//K8s  The kubernetes structure
+// K8s  The kubernetes structure
 type K8s struct {
 	SpecFactory         *spec.Factory
 	NodeDriverName      string
 	VolDriverName       string
 	secretConfigMapName string
 	customConfig        map[string]scheduler.AppConfig
+	eventsStorage       map[string][]scheduler.Event
 }
 
-//IsNodeReady  Check whether the cluster node is ready
+// IsNodeReady  Check whether the cluster node is ready
 func (k *K8s) IsNodeReady(n node.Node) error {
 	t := func() (interface{}, bool, error) {
 		if err := k8sCore.IsNodeReady(n.Name); err != nil {
@@ -144,12 +149,13 @@ func (k *K8s) String() string {
 	return SchedName
 }
 
-//Init Initialize the driver
+// Init Initialize the driver
 func (k *K8s) Init(schedOpts scheduler.InitOptions) error {
 	k.NodeDriverName = schedOpts.NodeDriverName
 	k.VolDriverName = schedOpts.VolDriverName
 	k.secretConfigMapName = schedOpts.SecretConfigMapName
 	k.customConfig = schedOpts.CustomAppConfig
+	k.eventsStorage = make(map[string][]scheduler.Event)
 
 	nodes, err := k8sCore.GetNodes()
 	if err != nil {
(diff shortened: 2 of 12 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,54 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
 	return expectedPVCSize, nil
 }
 
+func (k *K8s) ValidateEvents(pvc *v1.PersistentVolumeClaim, apRules *apapi.AutopilotRuleList) error {
+
+	// wait until autopilot posts all events
+	time.Sleep(time.Minute)
```
mborodin-px: it is not enough to wait for one minute. If I have cool down period of 5min, you will get this event "rule transitioning from ActivePending => ActiveInProgress." after 5 min.

trierra: autopilot injects event before starting cool down period

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
...
+
+	successEvens := []string{
+		"Initializing => Normal",
+		"Normal => Triggered",
+		"Triggered => ActiveActionsPending",
+		"ActiveActionsPending => ActiveActionsInProgress",
+		"ActiveActionsInProgress => ActiveActionsTaken",
+	}
```
mborodin-px: we have to add  ActiveActionsTaken => Normal event after cool down period finished

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
...
+	return nil
+}
+
+func checkEvents(successEvens []string, eventMap map[string]bool) error {
+	for _, expectedEvent := range successEvens {
+		if _, ok := eventMap[expectedEvent]; !ok {
+			return fmt.Errorf("failed to validate events. Expected %s", expectedEvent)
+		}
+	}
+	return nil
```
mborodin-px: add logic where  ActiveActionsTaken => Normal is the last event for PVC and no more after that, otherwise it will pass here even if it has Normal => Triggered after that

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
 	return expectedPVCSize, nil
 }
 
+// ValidateEvents verifies proper alerts and events on resize completion
+func (k *K8s) ValidateEvents(pvc *v1.PersistentVolumeClaim, apRules *apapi.AutopilotRuleList) error {
```
harsh-px: method doesn't need to be public. change to `validateAutopilotPVCEvents`

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
...
+
+	// wait until autopilot posts all events
+	time.Sleep(time.Minute)
+
+	successEvens := []string{
```
harsh-px: nit: use `successEvents`

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
...
+		"Initializing => Normal",
+		"Normal => Triggered",
+		"Triggered => ActiveActionsPending",
```
harsh-px: for state names, use constants from the autopilot-api project

https://github.com/libopenstorage/autopilot-api/blob/master/pkg/apis/autopilot/v1alpha1/autopilotrule.go#L148

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
 	return expectedPVCSize, nil
 }
 
+// ValidateEvents verifies proper alerts and events on resize completion
+func (k *K8s) ValidateEvents(pvc *v1.PersistentVolumeClaim, apRules *apapi.AutopilotRuleList) error {
+
+	// wait until autopilot posts all events
+	time.Sleep(time.Minute)
```
harsh-px: The validate implementation should not sleep for this arbitrary period since this is not deterministic. The caller can sleep before invoking ValidateEvents. The implementation of ValidateEvents has no context on how long to sleep.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
...
+				if time.Since(event.CreationTimestamp.Time).Hours() > float64(1) {
+					logrus.Infof("Can't capture all events for %s:%s, since resize has started more that 1h ago", apRule.Name, pvc.Name)
+					return nil
+				}
+
+				// cut last 3 entries from a message like `rule: pvc:pvc-25c transition from ActiveActionsInProgress => ActiveActionsTaken`
+				transitionMsg := strings.Split(event.Message, " ")
+				eventMap[strings.Join(transitionMsg[len(transitionMsg)-3:], " ")] = true
+			}
+			return checkEvents(successEvens, eventMap)
```
harsh-px: nit: can avoid the extra function here and move the code inline here itself.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,6 +2936,55 @@ func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRu
...
+			}
+
+			eventMap := make(map[string]bool)
+			for _, event := range events.Items {
+				if time.Since(event.CreationTimestamp.Time).Hours() > float64(1) {
+					logrus.Infof("Can't capture all events for %s:%s, since resize has started more that 1h ago", apRule.Name, pvc.Name)
+					return nil
+				}
+
+				// cut last 3 entries from a message like `rule: pvc:pvc-25c transition from ActiveActionsInProgress => ActiveActionsTaken`
```
harsh-px: how about we  do a more loose strings.Contains check instead of cutting the last 3 entries?

drivers/scheduler/scheduler.go
```diff
@@ -124,8 +124,8 @@ type Driver interface {
 	// GetVolumeParameters Returns a maps, each item being a volume and it's options
 	GetVolumeParameters(*Context) (map[string]map[string]string, error)
 
-	// InspectVolumes inspects a storage volume.
-	InspectVolumes(cc *Context, timeout, retryInterval time.Duration) error
+	// ValidateVolumes inspects a storage volume.
```
harsh-px: `ValidateVolumes inspects a storage volume` => `ValidateVolumes validates storage volumes in the provided context`

drivers/volume/portworx/portworx.go
```diff
@@ -2083,12 +2084,19 @@ func (d *portworx) EstimateVolumeExpandSize(apRule apapi.AutopilotRule, initialS
 
 					// check if calculated size is more than maxsize
 					if actionMaxSize, ok := ruleAction.Params[aututils.RuleMaxSize]; ok {
-						maxSize, _ := strconv.ParseUint(actionMaxSize, 10, 64)
-						if maxSize != 0 && calculatedTotalSize > maxSize {
+						maxSize, err := strconv.ParseUint(actionMaxSize, 10, 64)
+						if err != nil {
+							a, err1 := resource.ParseQuantity(actionMaxSize)
```
harsh-px: nit: rename `err1` to `parseErr`.

pkg/aututils/aututils.go
```diff
@@ -133,3 +95,71 @@ func PoolRuleByAvailableCapacity(usage, scalePercentage uint64, expandType strin
 		},
 	}
 }
+
+// VolumeRuleByTotalSize resizes volume by its total size
+func VolumeRuleByTotalSize(capacity int, scalePercentage int, maxSize string) apapi.AutopilotRule {
```
harsh-px: line 134 has `PvcRuleBy` and this has `VolumeRuleBy`. For consistency, let's pick one.

harsh-px:
```
./drivers/scheduler/k8s/k8s.go:1859:1: exported method K8s.GetWorkloadSizeFromAppSpec should have comment or be unexported
./drivers/scheduler/k8s/k8s.go:3149:1: exported method K8s.GetEvents should have comment or be unexported
```

CI is failing

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12.10-alpine AS build
+FROM golang:1.13.7-alpine AS build
```
harsh-px: rebase with master, these changes are already present there

drivers/scheduler/dcos/dcos.go
```diff
@@ -74,6 +74,17 @@ func (d *dcos) String() string {
...
 
+// GetEvents dumps events from event storage
+func (d *dcos) GetEvents() map[string][]scheduler.Event {
+	return nil
+}
+
+// ValidateAutopilotPVCEvents validates events for PVCs injected by autopilot
+func (d *dcos) ValidateAutopilotPVCEvents(ctx *scheduler.Context, coolDownPeriod int) error {
+	// TODO: not implemented
+	return nil
```
harsh-px: instead of nil return `return &errors.ErrNotSupported{`

drivers/scheduler/k8s/k8s.go
```diff
@@ -166,6 +173,13 @@ func (k *K8s) Init(schedOpts scheduler.InitOptions) error {
 	if err != nil {
 		return err
 	}
+
+	go func() {
+		err = k.collectEvents()
```
harsh-px: nit: use `err :=` so that the goroutine doesn't share the `err` variable from the parent function

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,32 +2962,104 @@ func (k *K8s) destroyBackupObjects(
...
 }
 
-// getExpectedAutopilotSizeForPvc calculates exected size of PVC based on autopilot rules and workload
-func (k *K8s) getExpectedAutopilotSizeForPvc(pvc *v1.PersistentVolumeClaim, apRules *apapi.AutopilotRuleList, wSize uint64) (uint64, error) {
-	var (
-		err             error
-		expectedPVCSize uint64
-	)
+// GetExpectedAutopilotVolumeExpand calculates expected size of PVC based on autopilot rule and workload
+func (k *K8s) GetExpectedAutopilotVolumeExpand(pvc *v1.PersistentVolumeClaim, apRule apapi.AutopilotRule, wSize uint64) (uint64, int, error) {
```
harsh-px: this does not need to be a public method. Only current package is using it.

harsh-px: The name `GetExpectedAutopilotVolumeExpand`doesn't sound grammatically correct. Suggest renaming to  `estimatePVCExpansion`

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,32 +2962,104 @@ func (k *K8s) destroyBackupObjects(
...
+	pvcSize, _ := pvcObjSize.AsInt64()
+	expectedPVCSize := uint64(pvcSize)
 
 	driver, err := volume.Get(k.VolDriverName)
 	if err != nil {
-		return 0, err
+		return 0, 0, err
 	}
+	if isAutopilotMatchPvcLabels(apRule, pvc) {
+		return driver.EstimateVolumeExpand(apRule, expectedPVCSize, wSize)
```
harsh-px: nit: rename `driver` to `volDriver` to be clear

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,32 +2962,104 @@ func (k *K8s) destroyBackupObjects(
...
+	}
+	return expectedPVCSize, 0, nil
+}
 
-	pvcObjSize := pvc.Spec.Resources.Requests[v1.ResourceStorage]
-	pvcSize, _ := pvcObjSize.AsInt64()
-	expectedPVCSize = uint64(pvcSize)
+// ValidateAutopilotPVCEvents verifies proper alerts and events on resize completion
+func (k *K8s) ValidateAutopilotPVCEvents(ctx *scheduler.Context, coolDownPeriod int) error {
```
harsh-px: don't see this method being invoked anywhere

drivers/volume/portworx/portworx.go
```diff
@@ -2118,13 +2119,14 @@ func (d *portworx) EstimatePoolExpandSize(apRule apapi.AutopilotRule, pool node.
 }
 
 // EstimatePoolExpandSize calculates the expected size of a volume based on autopilot rule, initial and workload sizes
-func (d *portworx) EstimateVolumeExpandSize(apRule apapi.AutopilotRule, initialSize, workloadSize uint64) (uint64, error) {
+func (d *portworx) EstimateVolumeExpand(apRule apapi.AutopilotRule, initialSize, workloadSize uint64) (uint64, int, error) {
```
harsh-px: function name doesn't match function comment

drivers/scheduler/k8s/k8s.go
trierra: ideally we need to change events, because validation, based on messages is not reliable. 
At this point, to keep validation simple, I'm using this approach (I'm actually building the message again). Otherwise the logic will be too complicated.

I'll create a AUT to modify events to be more suitable for testing

drivers/scheduler/k8s/k8s.go
trierra: this is a method to be called in pvc tests

drivers/scheduler/k8s/k8s.go
trierra: right

drivers/scheduler/k8s/k8s.go
```diff
@@ -12,11 +12,17 @@ import (
...
 	"strings"
+	"sync"
 	"text/template"
 	"time"
 
 	snapv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
```
harsh-px: remove extra spaces

tests/autopilot/autopilot_test.go
```diff
@@ -476,6 +472,54 @@ var _ = Describe(fmt.Sprintf("{%sPvcAndPoolExpand}", testSuiteName), func() {
 	})
 })
 
+var _ = Describe(fmt.Sprintf("{%sPvcExpand}", testSuiteName), func() {
+	It("Starts resizing volumes, when the volume capacity is less than 10Gb", func() {
```
harsh-px: description (10Gb) doesn't match the actual capacity in rule for test (5).

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,32 +2966,104 @@ func (k *K8s) destroyBackupObjects(
...
+		return volDriver.EstimateVolumeExpand(apRule, expectedPVCSize, wSize)
 	}
+	return expectedPVCSize, 0, nil
+}
 
-	pvcObjSize := pvc.Spec.Resources.Requests[v1.ResourceStorage]
-	pvcSize, _ := pvcObjSize.AsInt64()
-	expectedPVCSize = uint64(pvcSize)
+// ValidateAutopilotPVCEvents verifies proper alerts and events on resize completion
+func (k *K8s) ValidateAutopilotPVCEvents(ctx *scheduler.Context, coolDownPeriod int) error {
```
harsh-px: instead of a fixed `coolDownPeriod`, we can take in a timeout argument. The implementation will retry until the timeout. By doing so, we can finish the validation earlier than the coolDownPeriod if we find the events.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2929,32 +2966,104 @@ func (k *K8s) destroyBackupObjects(
...
-			expectedPVCSize, err = driver.EstimateVolumeExpandSize(apRule, uint64(pvcSize), wSize)
+	eventMap := make(map[string]int32)
+	for _, event := range k.GetEvents()["AutopilotRule"] {
+		eventMap[event.Message] = event.Count
+	}
+
+	for _, specObj := range ctx.App.SpecList {
+		if obj, ok := specObj.(*v1.PersistentVolumeClaim); ok {
+
+			listApRules, err := autopilot.Instance().ListAutopilotRules()
```
harsh-px: you need to check if the PVC has autopilot enabled as the user could provide you a context which doesn't have autopilot enabled or has certain volumes without autopilot.

drivers/scheduler/scheduler.go
```diff
@@ -184,11 +186,17 @@ type Driver interface {
...
+	// IsAutopilotEnabledForVolume checks if autopilot enabled for a given volume
 	IsAutopilotEnabledForVolume(*volume.Volume) bool
 
-	// GetSpecAppEnvVar gets app environment variable value by given key name
-	GetSpecAppEnvVar(ctx *Context, key string) string
+	// GetEvents dumps events from event storage
+	GetEvents() map[string][]Event
+
+	// ValidateAutopilotPVCEvents validates events for PVCs injected by autopilot
+	ValidateAutopilotPVCEvents(ctx *Context, coolDownPeriod int) error
```
harsh-px: remove the word "PVC" from the interface so that we can use it for validate events for other scheduler objects in future. for e.g pods.

drivers/scheduler/k8s/k8s.go
trierra: I need this argument because autopilot posts `ActiveActionsTaken => Normal` only after the cool down period, never earlier

drivers/scheduler/scheduler.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
harsh-px: nit: remove extra line

drivers/volume/portworx/portworx.go
```diff
@@ -3,6 +3,7 @@ package portworx
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/resource"
```
harsh-px: this import should be below line 16

harsh-px:
Can you post link to the jenkins job that you ran yesterday on this PR?

trierra:
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/326/

PR #412, opened 2020-02-04
[PTX-2267] Modified rebootOneNode to restart only those worker nodes

vinayakshnd:
where storage driver is installed.


vinayakshnd:
Job where RebootOneNode test passed:
https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20NextPX/job/tp-nextpx-gke-op/4/

PR #415, opened 2020-02-05
Bump up torpedo container to use golang 1.13

rodrigc:
In "go test" 1.13 and higher, you cannot call parse.Flags in init()

In https://golang.org/doc/go1.13#testing

    Testing flags are now registered in the new Init function,
    which is invoked by the generated main function for the test.
    As a result, testing flags are now only registered when running
    a test binary, and packages that call flag.Parse during package
    initialization may cause tests to fail.

See:
  https://groups.google.com/d/msg/golang-nuts/rkCdS1EQwSM/4gbwNh0QAwAJ
  golang/go#21051
  https://go-review.googlesource.com/c/go/+/173722/
  https://www.bountysource.com/issues/79987300-flag-provided-but-not-defined-test-timeout-go-1-13
  https://golang.org/pkg/testing/#hdr-Main
 golang-1.13

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,7 @@ WORKDIR /go/src/github.com/portworx/torpedo
 RUN apk update && \
     apk add git gcc  musl-dev && \
     apk add make && \
-    apk add openssh-client && \
-    go get github.com/onsi/ginkgo/ginkgo && \
-    go get github.com/onsi/gomega && \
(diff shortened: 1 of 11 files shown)


rodrigc:
Tested here: https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/312

PR #416, opened 2020-02-05
Fix assignment of default provsioner

Rohit-PX:

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -229,7 +229,7 @@ func (d *portworx) Init(sched string, nodeDriver string, token string, storagePr
 			return fmt.Errorf("driver %s, does not support provisioner %s", DriverName, storageProvisioner)
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -229,7 +229,7 @@ func (d *portworx) Init(sched string, nodeDriver string, token string, storagePr
 			return fmt.Errorf("driver %s, does not support provisioner %s", DriverName, storageProvisioner)
 		}
 	} else {
-		torpedovolume.StorageProvisioner = torpedovolume.DefaultStorageProvisioner
+		torpedovolume.StorageProvisioner = provisioners[torpedovolume.DefaultStorageProvisioner]
```
mborodin-px: in this case we don't need torpedovolume.DefaultStorageProvisioner

Rohit-PX:
Successful run:  https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/316/

PR #419, opened 2020-02-06
Override scheduler name using options

Rohit-PX:
##### Why do we need this PR?
Allow torpedo to override the scheduler when deploying apps


Rohit-PX:
Passed here:  
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/320/

PR #424, opened 2020-02-11
PTX-2230 Added autopilot rule test on resize failure

mborodin-px:
Basic tests passed:
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/328/
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/329/

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -33,6 +33,7 @@ import (
 	"github.com/portworx/torpedo/drivers/scheduler"
 	"github.com/portworx/torpedo/drivers/scheduler/spec"
 	"github.com/portworx/torpedo/drivers/volume"
+	"github.com/portworx/torpedo/pkg/aututils"
 	tp_errors "github.com/portworx/torpedo/pkg/errors"
 	"github.com/sirupsen/logrus"
 	appsapi "k8s.io/api/apps/v1"
@@ -730,8 +731,13 @@ func (k *K8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.
 		}
 		if autopilotEnabled {
 			apRule := options.AutopilotRule
+			labels := options.Labels
 			if apRule.Name != "" {
-				aRule, err := k.createAutopilotRule(apRule, options)
+				apRule.Labels = defaultTorpedoLabel
+				labelSelector := metav1.LabelSelector{MatchLabels: labels}
+				apRule.Spec.Selector = apapi.RuleObjectSelector{LabelSelector: labelSelector}
+				apRule.Spec.NamespaceSelector = apapi.RuleObjectSelector{LabelSelector: labelSelector}
+				aRule, err := k.CreateAutopilotRule(apRule)
 				if err != nil {
 					return nil, err
 				}
@@ -1199,44 +1205,6 @@ func (k *K8s) destroyCoreObject(spec interface{}, opts map[string]bool, app *spe
 
 	return pods, nil
 
-}
-
-func (k *K8s) createAutopilotRule(autopilotRule apapi.AutopilotRule, options scheduler.ScheduleOptions) (*apapi.AutopilotRule, error) {
-	k8sOps := k8sAutopilot
-
-	t := func() (interface{}, bool, error) {
-		autopilotRule.Labels = defaultTorpedoLabel
-		autopilotRule.Spec.Selector = apapi.RuleObjectSelector{
-			LabelSelector: metav1.LabelSelector{
-				MatchLabels: options.Labels,
-			},
-		}
-		autopilotRule.Spec.NamespaceSelector = apapi.RuleObjectSelector{
-			LabelSelector: metav1.LabelSelector{
-				MatchLabels: options.Labels,
-			},
-		}
-
-		apRule, err := k8sOps.CreateAutopilotRule(&autopilotRule)
-		if errors.IsAlreadyExists(err) {
-			if apRule, err := k8sOps.GetAutopilotRule(autopilotRule.Name); err == nil {
-				logrus.Infof("Using existing AutopilotRule: %v", apRule.Name)
-				return apRule, false, nil
-			}
-		}
-		if err != nil {
-			return nil, true, fmt.Errorf("Failed to create autopilot rule: %v. Err: %v", apRule.Name, err)
-		}
-
-		return apRule, false, nil
-	}
-
-	apRuleObj, err := task.DoRetryWithTimeout(t, k8sObjectCreateTimeout, DefaultRetryInterval)
-	if err != nil {
-		return nil, err
-	}
-
-	return apRuleObj.(*apapi.AutopilotRule), nil
 }
 
 func (k *K8s) substituteNamespaceInVolumes(volumes []v1.Volume, ns string) []v1.Volume {
@@ -1795,22 +1763,14 @@ func (k *K8s) ValidateVolumes(ctx *scheduler.Context, timeout, retryInterval tim
 				if err != nil {
 					return err
 				}
-				wSize, err := k.GetWorkloadSizeFromAppSpec(ctx)
-				if err != nil {
-					return err
-				}
-				var expectedPVCSize uint64
 				for _, rule := range listApRules.Items {
-					if expectedPVCSize, _, err = k.EstimatePVCExpansion(obj, rule, wSize); err != nil {
-						return err
-					}
-				}
-				logrus.Infof("[%v] expecting PVC size: %v\n", ctx.App.Key, expectedPVCSize)
-				err = k8sCore.ValidatePersistentVolumeClaimSize(obj, int64(expectedPVCSize), timeout, retryInterval)
-				if err != nil {
-					return &scheduler.ErrFailedToValidateStorage{
-						App:   ctx.App,
-						Cause: fmt.Sprintf("Failed to validate size: %v of PVC: %v. Err: %v", expectedPVCSize, obj.Name, err),
+					for _, a := range rule.Spec.Actions {
+						if a.Name == aututils.VolumeSpecAction {
+							err := k.validatePVCSize(ctx, obj, rule, timeout, retryInterval)
+							if err != nil {
+								return err
+							}
+						}
 					}
 				}
 				logrus.Infof("[%v] Validated PVC: %v size based on Autopilot rules", ctx.App.Key, obj.Name)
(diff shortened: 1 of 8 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -3019,6 +2993,63 @@ func (k *K8s) addLabelsToPVC(pvc *v1.PersistentVolumeClaim, labels map[string]st
...
+		return nil, err
+	}
+	logrus.Infof("Created autopilot rule: %+v", apRuleObj)
+
+	return apRuleObj.(*apapi.AutopilotRule), nil
+}
+
+// GetAutopilotRule gets the AutopilotRule for the provided name
+func (k *K8s) GetAutopilotRule(name string) (*apapi.AutopilotRule, error) {
+	return k8sAutopilot.GetAutopilotRule(name)
```
trierra: as far as i see this function is called from inside the driver only. The separate function and changing the api is not needed at this point, just call k8sAutopilot.GetAutopilotRule(name) from the UpdateRule function directly

drivers/scheduler/k8s/k8s.go
```diff
@@ -3019,6 +2993,63 @@ func (k *K8s) addLabelsToPVC(pvc *v1.PersistentVolumeClaim, labels map[string]st
...
+}
+
+// UpdateAutopilotRule updates the AutopilotRule
+func (k *K8s) UpdateAutopilotRule(apRule apapi.AutopilotRule) (*apapi.AutopilotRule, error) {
+	aRule, err := k.GetAutopilotRule(apRule.Name)
+	if err != nil {
+		return nil, err
+	}
+	aRule.Spec = apRule.Spec
```
trierra: the same value has been assigned, remove this line

drivers/scheduler/k8s/k8s.go
```diff
@@ -3019,6 +2993,63 @@ func (k *K8s) addLabelsToPVC(pvc *v1.PersistentVolumeClaim, labels map[string]st
...
+
+	return k8sAutopilot.UpdateAutopilotRule(aRule)
+}
+
+// DeleteAutopilotRule deletes the AutopilotRule of the given name
+func (k *K8s) DeleteAutopilotRule(name string) error {
+	if err := k8sAutopilot.DeleteAutopilotRule(name); err != nil {
```
trierra: just 'return k8sAutopilot.DeleteAutopilotRule(name)'

drivers/scheduler/scheduler.go
```diff
@@ -189,6 +189,21 @@ type Driver interface {
...
+	GetAutopilotRule(name string) (*apapi.AutopilotRule, error)
+
+	// UpdateAutopilotRule updates the AutopilotRule
+	UpdateAutopilotRule(apapi.AutopilotRule) (*apapi.AutopilotRule, error)
+
+	// DeleteAutopilotRule deletes the AutopilotRule of the given name
+	DeleteAutopilotRule(name string) error
+
+	// ListAutopilotRules lists AutopilotRules
+	ListAutopilotRules() (*apapi.AutopilotRuleList, error)
```
trierra: it makes sense to add these api functions only if you want to give tests control on autopilot rules, otherwise if you're not calling these functions from outside the K8s driver, you don't need to modify api at this point (you can always add them later)

tests/autopilot/autopilot_test.go
```diff
@@ -476,6 +482,62 @@ var _ = Describe(fmt.Sprintf("{%sPvcAndPoolExpand}", testSuiteName), func() {
 	})
 })
 
+var _ = Describe(fmt.Sprintf("{%sPoolResizeFailure}", testSuiteName), func() {
+	It("has to start IO workloads, create rules that resize pools based on capacity, restart autopilot and validate pools have been resized once", func() {
+		var err error
```
trierra: the error can be defined at L510, remove this line

drivers/scheduler/k8s/k8s.go
mborodin-px: It is not the same value. aRule is runtime autopilot rule, and apRule is the rule which will be updated

drivers/scheduler/k8s/k8s.go
```diff
@@ -3019,6 +2993,49 @@ func (k *K8s) addLabelsToPVC(pvc *v1.PersistentVolumeClaim, labels map[string]st
 	}
 }
 
+// CreateAutopilotRule creates the AutopilotRule object
+func (k *K8s) CreateAutopilotRule(apRule apapi.AutopilotRule) (*apapi.AutopilotRule, error) {
+	t := func() (interface{}, bool, error) {
+		apRule.Labels = defaultTorpedoLabel
+		aRule, err := k8sAutopilot.CreateAutopilotRule(&apRule)
+		if errors.IsAlreadyExists(err) {
```
trierra: fix as we discussed offline

drivers/scheduler/k8s/k8s.go
```diff
@@ -717,8 +717,20 @@ func (k *K8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.
...
 		if autopilotEnabled {
 			apRule := options.AutopilotRule
+			labels := options.Labels
 			if apRule.Name != "" {
-				aRule, err := k.createAutopilotRule(apRule, options)
+				apRule.Labels = defaultTorpedoLabel
+				apRule.Spec.Selector = apapi.RuleObjectSelector{
+					LabelSelector: metav1.LabelSelector{
+						MatchLabels: labels,
+					},
```
harsh-px: since we are using the same selector here and in NamespaceSelector, save in a variable and use that. This will help avoid divergence.

drivers/scheduler/k8s/k8s.go
harsh-px: just for context, what is the fix required here?

drivers/scheduler/k8s/k8s.go
```diff
@@ -3019,6 +2993,49 @@ func (k *K8s) addLabelsToPVC(pvc *v1.PersistentVolumeClaim, labels map[string]st
...
+	apRuleObj, err := task.DoRetryWithTimeout(t, k8sObjectCreateTimeout, DefaultRetryInterval)
+	if err != nil {
+		return nil, err
+	}
+	logrus.Infof("Created autopilot rule: %+v", apRuleObj)
+
+	return apRuleObj.(*apapi.AutopilotRule), nil
+}
+
+// UpdateAutopilotRule updates the AutopilotRule
```
harsh-px: nit: no need to add function comments here. For interfaces, function comments need to be present only when we declare the interface as that provides the contract. The implementations just honor the contract.

tests/autopilot/autopilot_test.go
```diff
@@ -476,6 +482,61 @@ var _ = Describe(fmt.Sprintf("{%sPvcAndPoolExpand}", testSuiteName), func() {
 	})
 })
 
+var _ = Describe(fmt.Sprintf("{%sPoolResizeFailure}", testSuiteName), func() {
+	It("has to start IO workloads, create rules that resize pools based on capacity, restart autopilot and validate pools have been resized once", func() {
```
harsh-px: incorrect description. The test is not restarting autopilot.

tests/autopilot/autopilot_test.go
```diff
@@ -476,6 +482,61 @@ var _ = Describe(fmt.Sprintf("{%sPvcAndPoolExpand}", testSuiteName), func() {
...
+			for _, p := range node.StoragePools {
+				if smallestPoolSize == 0 {
+					smallestPoolSize = p.TotalSize
+				} else {
+					if p.TotalSize < smallestPoolSize {
+						smallestPoolSize = p.TotalSize
+					}
+				}
+			}
+		}
```
harsh-px: this logic is now present in 2 tests. Extract into a function.

tests/autopilot/autopilot_test.go
```diff
@@ -476,6 +482,61 @@ var _ = Describe(fmt.Sprintf("{%sPvcAndPoolExpand}", testSuiteName), func() {
...
+		apRules := []apapi.AutopilotRule{
+			aututils.PoolRuleByTotalSize((smallestPoolSize/units.GiB)+1, 100*16, aututils.RuleScaleTypeAddDisk, nil),
+		}
+
+		contexts := scheduleAppsWithAutopilot(testName, apRules)
+
+		err := waitForAutopilotFailedEvent(apRules, "")
+		Expect(err).NotTo(HaveOccurred())
+
+		Step("updating autopilot rules with reasonable values", func() {
```
harsh-px: change `reasonable` to `correct`. The word reasonable puts things in a grey area.

tests/autopilot/autopilot_test.go
```diff
@@ -476,6 +482,61 @@ var _ = Describe(fmt.Sprintf("{%sPvcAndPoolExpand}", testSuiteName), func() {
...
+					smallestPoolSize = p.TotalSize
+				} else {
+					if p.TotalSize < smallestPoolSize {
+						smallestPoolSize = p.TotalSize
+					}
+				}
+			}
+		}
+
+		apRules := []apapi.AutopilotRule{
```
harsh-px: add a comment that below rule is expected to fail because ...

drivers/scheduler/k8s/k8s.go
mborodin-px: if I remove comment, I'm getting this error from golint:
 ./drivers/scheduler/k8s/k8s.go:3015:1: exported method K8s.UpdateAutopilotRule should have comment or be unexported

drivers/scheduler/k8s/k8s.go
mborodin-px: return error should be inside if errors.IsAlreadyExists(err) {

trierra: move `if err != nil {` above if `errors.IsAlreadyExists(err) {` line

drivers/scheduler/k8s/k8s.go
trierra: consider using more readable names

drivers/scheduler/k8s/k8s.go
harsh-px: @mborodin-px  looks like this is still not done

drivers/scheduler/k8s/k8s.go
trierra: could you please rename `aRule -> rule`

drivers/scheduler/k8s/k8s.go
mborodin-px: @harsh-px it is there:

harsh-px:
looks like we have conflicts here now due to the other autopilot merge

harsh-px:
Let's run this branch on the tp-dev-test job before merging to be sure if we aren't breaking any existing test.

PR #409, opened 2020-01-31
PTX-2197 fix collect bundle for openshift/rke

thiguetta:
Signed-off-by: Thiago Gonzaga <thi_gonzaga@yahoo.com.br>

PR #426, opened 2020-02-11
PTX-2197 In ValidateCleanup() and PerformSystemCheck(), try to collect support bundle

rodrigc:
Refactored from https://github.com/portworx/torpedo/pull/409

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -129,17 +129,24 @@ func InitInstance() {
 
 // ValidateCleanup checks that there are no resource leaks after the test run
 func ValidateCleanup() {
-	Step(fmt.Sprintf("validate cleanup of resources used by the test suite"), func() {
-		t := func() (interface{}, bool, error) {
-			if err := Inst().V.ValidateVolumeCleanup(); err != nil {
-				return "", true, err
+	context("For validate clean up", func() {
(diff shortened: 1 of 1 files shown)


tests/common.go
```diff
@@ -129,17 +129,24 @@ func InitInstance() {
 
 // ValidateCleanup checks that there are no resource leaks after the test run
 func ValidateCleanup() {
-	Step(fmt.Sprintf("validate cleanup of resources used by the test suite"), func() {
-		t := func() (interface{}, bool, error) {
-			if err := Inst().V.ValidateVolumeCleanup(); err != nil {
-				return "", true, err
-			}
+	context("For validate clean up", func() {
```
harsh-px: nit: change `For validate clean up` to `To validate clean up` so that it reads better.

That is, `To validate clean up, validate cleanup of resources used by the test suite`

tests/common.go
```diff
@@ -493,7 +500,7 @@ func runCmd(cmd string, n node.Node) {
 // PerformSystemCheck check if core files are present on each node
 func PerformSystemCheck() {
 	context(fmt.Sprintf("checking for core files..."), func() {
```
harsh-px: not related to your change but the description here can be improved to 

`To check for core files`

`verify if cores files are present on any node`

PR #430, opened 2020-02-14
Fix port to be used for cluster pair

disrani-px:
Introduced with a102306dc9cbcc609d1ba8df8938d3bb51f9183e


Rohit-PX:
we should run  https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/ to verify

disrani-px:
No torpedo test uses it, otherwise, it would have failed already

Rohit-PX:
> No torpedo test uses it, otherwise, it would have failed already

ok

PR #433, opened 2020-02-24
Fix for getPxNode where grpc response is nil

mborodin-px:
This fix is for handling `panic: runtime error` in getPxNode function where Inspect node method returns nil.


harsh-px:
Run this on the tp-dev-test job before merging.

mborodin-px:
Test passed here: https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/354/

PR #434, opened 2020-02-26
Fix for autopilot tests

mborodin-px:
* fix for getExpectedPoolSizes when we have multiple autopilot rules and expectedPoolSizes map will be always overwritten with initial size
* fix for PVCRuleByTotalSize rule
* added more unit tests and fix existing tests

tested here:
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/364/
https://jenkins.portworx.co/job/Autopilot/job/tp-aut-nextpx-aks-dev/1/

### DIFF
--- a/pkg/aututils/aututils.go
+++ b/pkg/aututils/aututils.go
@@ -100,14 +100,14 @@ func PoolRuleByAvailableCapacity(usage, scalePercentage uint64, expandType strin
 func PVCRuleByTotalSize(capacity int, scalePercentage int, maxSize string) apapi.AutopilotRule {
 	apRuleObject := apapi.AutopilotRule{
 		ObjectMeta: meta_v1.ObjectMeta{
-			Name: fmt.Sprintf("pvc-total-size-%s", strings.ToLower(maxSize)),
+			Name: fmt.Sprintf("pool-total-%d-scale-%d", capacity, scalePercentage),
 		},
 		Spec: apapi.AutopilotRuleSpec{
 			Conditions: apapi.RuleConditions{
 				Expressions: []*apapi.LabelSelectorRequirement{
 					{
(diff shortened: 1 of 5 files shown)


pkg/aututils/aututils.go
```diff
@@ -123,6 +123,7 @@ func PVCRuleByTotalSize(capacity int, scalePercentage int, maxSize string) apapi
 		},
 	}
 	if maxSize != "" {
+		apRuleObject.Name = fmt.Sprintf("-%s-maxsize-%s", apRuleObject.Name, strings.ToLower(maxSize))
```
trierra: `"-%s-maxsize-%s"` looks like it begins from `-`

pkg/aututils/aututils.go
```diff
@@ -100,14 +100,14 @@ func PoolRuleByAvailableCapacity(usage, scalePercentage uint64, expandType strin
...
-			Name: fmt.Sprintf("pvc-total-size-%s", strings.ToLower(maxSize)),
+			Name: fmt.Sprintf("pool-total-%d-scale-%d", capacity, scalePercentage),
 		},
 		Spec: apapi.AutopilotRuleSpec{
 			Conditions: apapi.RuleConditions{
 				Expressions: []*apapi.LabelSelectorRequirement{
 					{
 						Key:      PxVolumeTotalCapacityMetric,
-						Operator: apapi.LabelSelectorOpGt,
+						Operator: apapi.LabelSelectorOpLt,
```
trierra: why Lt ?

mborodin-px: you want to resize PVC until the total size is less than provided value, otherwise it will constantly resize until maxsize if provided. We have the same logic in PoolRuleByTotalSize.

PR #436, opened 2020-02-27
Add Restore, SchedulePolicy and BackupSchedule Apis to backup driver

raviprasadsn:
Implemented new interfaces
a. Restore
b. SchedulePolicy
c. BackupSchedule

### DIFF
--- a/tests/backup/backup_test.go
+++ b/tests/backup/backup_test.go
@@ -2,10 +2,7 @@ package tests
 
 import (
 	"testing"
-	"fmt"
-	"os"
 
-	api "github.com/portworx/px-backup-api/pkg/apis/v1"
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/reporters"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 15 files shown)


vinayakshnd:
LGTM. @raviprasadsn there are no compilation/run-time errors with this PR, right?

raviprasadsn:
> LGTM. @raviprasadsn there are no compilation/run-time errors with this PR, right?

@vinayakshnd, travis builds have passed. So no compilation errors. I'm able to exercise these apis from integration side.

tests/backup/backup_test.go
```diff
@@ -32,7 +32,3 @@ var _ = AfterSuite(func() {
 	//PerformSystemCheck()
 	//ValidateCleanup()
 })
-
-func init() {
-	ParseFlags()
-}
```
thiguetta: we must include 
```
func TestMain(m *testing.M) {
	// call flag.Parse() here if TestMain uses flags
	ParseFlags()
	os.Exit(m.Run())
}
```

otherwise, the flags won't be parsed.

raviprasadsn: Yes, it should be added. It will be added as part of tests that @vinayakshnd is adding.

PR #437, opened 2020-03-03
Add WaitForRunning function for backup

stgleb:


vinayakshnd:
@raviprasadsn https://github.com/portworx/torpedo/pull/439 will address WaitforCompletion

drivers/backup/portworx/portworx.go
```diff
@@ -252,6 +255,30 @@ func (p *portworx) DeleteBackupSchedule(req *api.BackupScheduleDeleteRequest) (*
...
+		resp, err := p.backupManager.Inspect(context.Background(), req)
+
+		if err != nil {
+			return nil, true, err
+		}
+
+		// If backup is in progress - stop
+		if resp.Backup.GetStatus().Status == api.BackupInfo_StatusInfo_InProgress {
+			return "", false, nil
+		}
```
vinayakshnd: @stgleb on other terminal conditions like failed, invalid statuses function should give up checking for backup to move to InProgress state.

stgleb: gotcha thanks

stgleb: Done

vinayakshnd: @stgleb did you test these changes? With my runs, `task.DoRetryWithTimeout()` func is not returning error in case of failed, aborted, invalid cases, so I am explicitly returning  error from `WaitFor...()` function

vinayakshnd: @stgleb once we do above change, we should be good to merge this.

PR #438, opened 2020-03-03
[PTX-2232] Added backup test

vinayakshnd:


drivers/backup/backup.go
```diff
@@ -98,6 +99,11 @@ type Backup interface {
 
 	// DeleteBackup deletes backup
 	DeleteBackup(req *api.BackupDeleteRequest) (*api.BackupDeleteResponse, error)
+
+	// WaitForBackupCompletion waits for backup to complete successfully
+	// or till timeout is reached. API should poll every `timeBeforeRetry`
+	WaitForBackupCompletion(timeout time.Duration,
```
stgleb: Maybe lets put request first and timeouts and reties in the end of argument list?

drivers/backup/backup.go
```diff
@@ -66,6 +67,9 @@ type Cluster interface {
 	// CreateCluster creates a cluste object
 	CreateCluster(req *api.ClusterCreateRequest) (*api.ClusterCreateResponse, error)
 
+	// CreateNewCluster creates a cluste object
+	CreateNewCluster(name string, cloudCred string, kubeconfigPath string, orgID string) (*api.ClusterCreateResponse, error)
```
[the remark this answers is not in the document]
vinayakshnd: @raviprasadsn u can suggest better name here. Eventually all APIs we want to replace to accept actual inputs instead of directly request API objects as we discussed the call. This is because logic to construct request API objects is getting duplicated in both torpedo and in backup integration.

drivers/backup/backup.go
```diff
@@ -119,6 +128,14 @@ type Restore interface {
 
 	// DeleteRestore deletes a restore object
 	DeleteRestore(req *api.RestoreDeleteRequest) (*api.RestoreDeleteResponse, error)
+
+	CreateNewRestore(restoreName string, backupName string, namespaceMapping map[string]string,
```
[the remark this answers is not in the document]
vinayakshnd: @raviprasadsn same as above.

drivers/backup/backup.go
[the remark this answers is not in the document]
raviprasadsn: I agree partly to the need for creating request objects. But this API implementation doesn't just create a request object, it actually invokes the px-backup api to create a cluster.  I believe request creation should be a utility instead of an exported driver interface. Driver interfaces should be uniform and having a structure passed as parameter to interface avoids changing the API itself if for some reason structure of the grpc message changes in future date.

drivers/backup/portworx/portworx.go
```diff
@@ -162,6 +164,25 @@ func (p *portworx) CreateCluster(req *api.ClusterCreateRequest) (*api.ClusterCre
 	return p.clusterManager.Create(context.Background(), req)
 }
 
+func (p *portworx) CreateNewCluster(name string, cloudCred string, kubeconfigPath string, orgID string) (*api.ClusterCreateResponse, error) {
```
thiguetta: does it really need to be exported? Can we simple create `ClusterCreateRequest ` outside then call `CreateCluster`?

tests/backup/backup_test.go
```diff
@@ -292,40 +298,63 @@ func DeleteBackupLocation(name string, orgID string) {
...
-			"PXBACKUP_KUBECONFIG Environment variable should not be empty")
+	err := dumpKubeConfigs(ConfigMapName, kubeconfigList)
+	Expect(err).NotTo(HaveOccurred(),
+		fmt.Sprintf("Failed to get kubeconfigs [%v] from configmap [%s]", kubeconfigList, ConfigMapName))
 
-		kubeconfigRaw, err := ioutil.ReadFile(kubeconfigPath)
+	// Register source cluster with backup driver
+	Step(fmt.Sprintf("Create cluster [%s] in org [%s]", SourceClusterName, orgID), func() {
+		srcClusterKubeConfig := fmt.Sprintf("%s/%s", KubeconfigDirectory, kubeconfigList[0])
+		_, err = Inst().Backup.CreateNewCluster(SourceClusterName, cloudCred, srcClusterKubeConfig, orgID)
```
thiguetta: create a `ClusterCreateRequest` there then call `CreateCluster`, otherwise remove `CreateCluster` and stick to use only `CreateNewCluster ` cause I don't see `CreateCluster` being called anywhere else

tests/backup/backup_test.go
```diff
@@ -292,40 +298,63 @@ func DeleteBackupLocation(name string, orgID string) {
...
-		}
-		clusterCreateReq := &api.ClusterCreateRequest{
-			CreateMetadata: metadata,
-			Cluster:        clusterInfo,
-		}
-		_, err = backupDriver.CreateCluster(clusterCreateReq)
+	// Register destination cluster with backup driver
+	Step(fmt.Sprintf("Create cluster [%s] in org [%s]", DestinationClusterName, orgID), func() {
+		destClusterKubeConfig := fmt.Sprintf("%s/%s", KubeconfigDirectory, kubeconfigList[1])
+		_, err = Inst().Backup.CreateNewCluster(DestinationClusterName, cloudCred, destClusterKubeConfig, orgID)
```
thiguetta: same here

tests/backup/backup_test.go
raviprasadsn: CreateCluster is used by backup-integration suite

drivers/backup/backup.go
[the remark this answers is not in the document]
vinayakshnd: @raviprasadsn removed CreateNewCluster() API. keeping just `CreateCluster()` API.

drivers/backup/portworx/portworx.go
vinayakshnd: @thiguetta removed `CreateNewCluster()` API call altogether from the interface.

tests/backup/backup_test.go
vinayakshnd: @thiguetta / @raviprasadsn removed `CreateNewCluster()` API call altogather.

tests/backup/backup_test.go
vinayakshnd: Removed CreateNewCluster()

PR #439, opened 2020-03-03
Added WaitForBackupCompletion() API in Backup driver and its

vinayakshnd:
implementation.

### DIFF
--- a/drivers/backup/portworx/portworx.go
+++ b/drivers/backup/portworx/portworx.go
@@ -208,8 +208,12 @@ func (p *portworx) DeleteBackup(req *api.BackupDeleteRequest) (*api.BackupDelete
 
 // WaitForBackupCompletion waits for backup to complete successfully
 // or till timeout is reached. API should poll every `timeBeforeRetry` duration
-func (p *portworx) WaitForBackupCompletion(req *api.BackupInspectRequest,
+func (p *portworx) WaitForBackupCompletion(backupName string, orgID string,
 	timeout time.Duration, timeBeforeRetry time.Duration) error {
+	req := &api.BackupInspectRequest{
+		Name:  backupName,
+		OrgId: orgID,
+	}
 	var backupError error
 	f := func() (interface{}, bool, error) {
 		inspectBkpResp, err := p.backupManager.Inspect(context.Background(), req)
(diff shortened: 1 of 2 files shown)


drivers/backup/portworx/portworx.go
```diff
@@ -204,6 +206,37 @@ func (p *portworx) DeleteBackup(req *api.BackupDeleteRequest) (*api.BackupDelete
...
+		if err != nil {
+			// Error occured, just retry
+			return nil, true, err
+		}
+
+		// Check if backup status is complete
+		currentStatus := inspectBkpResp.GetBackup().GetStatus().GetStatus()
+		if currentStatus == api.BackupInfo_StatusInfo_Success {
+			// If backup is complete, dont retry again
+			return nil, false, nil
```
disrani-px: Stop retrying on failure and other terminal status

vinayakshnd: @disrani-px done

vinayakshnd:
@disrani-px addressed review comments.

drivers/backup/portworx/portworx.go
```diff
@@ -204,6 +206,44 @@ func (p *portworx) DeleteBackup(req *api.BackupDeleteRequest) (*api.BackupDelete
...
+				req.GetName(), currentStatus)
+			return nil, false, backupError
+		}
+		return nil,
+			true,
+			fmt.Errorf("backup [%v] is in [%s] state. Waiting to become Complete",
+				req.GetName(), currentStatus)
+	}
+
+	_, err := task.DoRetryWithTimeout(f, timeout, timeBeforeRetry)
```
stgleb: What is the reason of using backupErr through closure? Is there a case when returned error is lost inside task  code?

vinayakshnd: @stgleb yes, the returned error is getting lost inside the task code. In case of failed backups, `nil` error is getting returned.

stgleb: Wow,that is wierd. I'll update my code for that.

vinayakshnd:
@stgleb / @raviprasadsn / @disrani-px ,
In the latest commit, added `WaitForRestoreCompletion()` similar to `WaitForBackupCompletion()`. Also made both the APIs to accept `backupName/restoreName` and `orgID` instead of direct request objects.

PR #440, opened 2020-03-08
Switch kubeconfig for k8s scheduler driver

vinayakshnd:


drivers/scheduler/k8s/k8s.go
```diff
@@ -183,6 +183,33 @@ func (k *K8s) Init(schedOpts scheduler.InitOptions) error {
...
 }
 
+// SetConfig sets kubeconfig. If kubeconfigPath == "" then
+// sets it to inClusterConfig
+func (k *K8s) SetConfig(kubeconfigPath string) error {
+	var config *rest.Config
+	var err error
+
+	if kubeconfigPath == "" {
+		config = nil
```
thiguetta: I suggest that instead of setting config to nil and then calling a bunch of SetConfig(nil), return nil right away, no need to set config as nil, since sched-ops default is already inClusterConfig

PR #441, opened 2020-03-10
PTX-1952 Added stork upgrade to upgrade test

mborodin-px:
Added ENABLE_STORK_UPGRADE variable to enable stork upgrade during volume driver upgrade.
Since we added `comp=stork` to the URL as parameter, starting from version 2.1 PX, we will skip the stork update for all versions below in the test update pipelines.


pkg/osutils/osutils.go
```diff
@@ -0,0 +1,125 @@
+package osutils
```
thiguetta: I don't know if you want to execute it in a specific node, otherwise, this os utils will only execute inside torpedo pod, is that what you want?

mborodin-px: yes, this is only for torpedo pod. As I'm using in multiple places I decided to move all commands to the package.

PR #443, opened 2020-03-11
AUT-129 Added tests for fixed scale size

mborodin-px:
* Added unit tests
* Added calculation of pool size on fixed scale size
* Added tests for fixed scale size (add-disk and resize-disk)

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -2094,15 +2094,28 @@ func (d *portworx) EstimatePoolExpandSize(apRule apapi.AutopilotRule, pool node.
 					Operation: "Pool Condition Expression Key",
 				}
 			}
-
 			if doesConditionMatch(metricValue, conditionExpression) {
 				for _, ruleAction := range apRule.Spec.Actions {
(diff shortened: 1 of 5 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -2094,20 +2094,38 @@ func (d *portworx) EstimatePoolExpandSize(apRule apapi.AutopilotRule, pool node.
...
-						calculatedTotalSize += requiredNewDisks * baseDiskSize
-					} else {
-						calculatedTotalSize += uint64(requiredScaleSize)
+						requiredScaleSize := float64(calculatedTotalSize * actionScalePercentage / 100)
+						if actionScaleType == aututils.RuleScaleTypeAddDisk {
+							requiredNewDisks := uint64(math.Ceil(requiredScaleSize / float64(baseDiskSize)))
+							calculatedTotalSize += requiredNewDisks * baseDiskSize
+						} else {
+							calculatedTotalSize += uint64(requiredScaleSize)
+						}
```
harsh-px: these lines look like they are common with the same logic in else if. Can you check if we can move this common logic outside. So just calculate `requiredScaleSize` in the if and else if. And once we have that, move the logic to calculate calculatedTotalSize outside.

mborodin-px: fixed

PR #446, opened 2020-03-19
[PTX-2376] Add backup kill stork test

stgleb:


tests/backup/backup_test.go
```diff
@@ -131,6 +150,121 @@ var _ = Describe("{BackupCreateRestore}", func() {
...
+			taskID, err := sched.Instance().Schedule(t,
+				sched.Periodic(time.Second),
+				time.Now().Add(timeout), true)
+			Expect(err).NotTo(HaveOccurred())
+			defer sched.Instance().Cancel(taskID)
+		})
+
+		Step(fmt.Sprintf("Wait for Backup [%s] to complete", BackupName), func() {
+			for _, namespace := range bkpNamespaces {
+				err := Inst().Backup.WaitForBackupCompletion(
```
raviprasadsn: is it possible that the backups are already completed before we kill stork. If I understand correctly, after killing stork, we expect either backups to fail or pending. Let me know if my understanding is right about this test.

tests/backup/backup_test.go
```diff
@@ -47,9 +60,15 @@ func TestBackup(t *testing.T) {
 var _ = BeforeSuite(func() {
 	InitInstance()
 	CreateOrganization(orgID)
```
vinayakshnd: @stgleb CreateOrganization() call itself fail for other non-backup jenkins jobs (jobs that are not configured to run backup tests)

tests/backup/backup_test.go
```diff
@@ -131,6 +150,121 @@ var _ = Describe("{BackupCreateRestore}", func() {
...
+			timeout := time.Duration(rand.Int()%120)*time.Second + 60
+			time.Sleep(time.Duration(timeout) * time.Second)
+
+			t := func(interval sched.Interval) {
+				ctx := &scheduler.Context{
+					App: &spec.AppSpec{
+						SpecList: []interface{}{
+							&appsapi.Deployment{
+								ObjectMeta: meta_v1.ObjectMeta{
+									Name:      storkDeploymentName,
```
vinayakshnd: @vilasd / @stgleb the original intention of the test is to kill  stork leader pod while backup is in progress. The stork deployment will elect new leader and continue remaining backup process. User should also be able to restore such backup successfully. Deleting the entire stork deployment itself will not achieve this purpose. If we delete entire stork deployment then backup itself will discontinue. Correct me if I am wrong.

stgleb: @vinayakshnd I thought this function https://github.com/portworx/torpedo/blob/master/drivers/scheduler/k8s/k8s.go#L1550 will delete only underlying pod of deployment, but not deployment itself. Backup works fine so far.

tests/backup/backup_test.go
stgleb: well, you've got it. I was also wondering about that. Do you think is there anyother way to syncronize stork killing and backup?

harsh-px: Use trigger options in the DeleteTasks API. Supply a trigger function such that it will trigger as soon as your backup starts.

stgleb: Thank you @harsh-px , I'll check it out

tests/backup/backup_test.go
stgleb: well, I can put  a code for determining cloud provider by creds that will go  out of this BeforeSuite if this job was not run for backup.

vinayakshnd:
@gleb can u put link to passed torpedo job run 
1. with backup_test enabled 
2. without backup_test enabled.

tests/backup/backup_test.go
```diff
@@ -20,15 +61,642 @@ func TestBackup(t *testing.T) {
...
+		backupDriver.DeleteBackup(bkpDeleteRequest)
+		// Best effort cleanup, dont fail test, if deletion fails
+		//Expect(err).NotTo(HaveOccurred(),
+		//	fmt.Sprintf("Failed to delete backup [%s] in org [%s]", backupName, orgID))
+		// TODO: validate createClusterResponse also
+	})
+}
+
+func discoverProvider() string {
+	if os.Getenv("AWS_ACCESS_KEY_ID") != "" && os.Getenv("AWS_SECRET_ACCESS_KEY") != "" {
```
vinayakshnd, line 691: @stgleb mere presence of AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY  env variables should not be the sole condition for determining whether to create org, cloud-creds, backup-locations etc. in BeforeSuite(). In future, these env vars can be passed to torpedo for different purposes like cloud-drive validations in ASG test cases etc.

So, either we should have an explicit way to tell to torpedo to create px-backup objects or completely remove objects creation from BeforeSuite() and let them get created only when user wants to run backup tests.

stgleb: I agree, as we decided - lets introduce single env variable responsible for determining type of cloud provider the test is running on.

drivers/scheduler/k8s/specs/elasticsearch-aks/px-elasticdata-app.yaml
```diff
@@ -0,0 +1,152 @@
...
+        configMap:
+          name: es-config
+          items:
+          - key: elasticsearch.yml
+            path: elasticsearch.yml
+  volumeClaimTemplates:
+  - metadata:
+      name: es-data
+      annotations:
+        volume.beta.kubernetes.io/storage-class: azure-es-sc
```
vinayakshnd, line 115: @stgleb I think storage-class is the only difference between existing specs and new spec for AKS. Can we templatize this so that we do not need to maintain 3 different app specs (AKS, AWS and for Google).

deployments/deploy-ssh.sh
```diff
@@ -149,6 +153,22 @@ if [ $timeout -gt 600 ]; then
   describe_pod_then_exit
 fi
 
+# List of additional kubeconfigs of k8s clusters to register with px-backup, px-dr
+FROM_FILE=""
```
vinayakshnd, line 157: Can u check things which are already in master branch are showing up here?

stgleb: Sure

tests/backup/backup_test.go
```diff
@@ -61,8 +84,296 @@ func TestBackup(t *testing.T) {
...
+					fmt.Sprintf("Failed to wait for backup [%s] to complete. Error: [%v]",
+						BackupName, err))
+			}
+		})
+
+		Step(fmt.Sprintf("Create Restore [%s]", RestoreName), func() {
+			for _, namespace := range bkpNamespaces {
+				backupName := fmt.Sprintf("%s-%s", BackupName, namespace)
+				restoreName := fmt.Sprintf("%s-%s", RestoreName, namespace)
+				logrus.Infof("Create restore %s:%s:%s from backup %s:%s:%s",
```
vinayakshnd, line 279: We can move this description to Step() description itself so that we can remove the duplicate log here

vinayakshnd:
This PR is expected to work as is when backend is Portworx as well as AKS? If yes, then in that case, can u paste jenkins job links where:
1. Volume backend is PX.
2. Volume backend is AKS.
3. Job without backup test and without any other ENV vars required for backup test to work

stgleb:
@vinayakshnd yes, maybe except numer 2. I need a spawn PR that allows to install stork without getting porx to be merged.

tests/backup/backup_test.go
stgleb: Well, that won't work that way. I need to print every namespace and every restore created this needs to be loged from the loop.

deployments/deploy-ssh.sh
stgleb: I've just put it above in them PR.

tests/backup/backup_test.go
vinayakshnd: Got it

vinayakshnd:
For 
> @vinayakshnd yes, maybe except numer 2. I need a spawn PR that allows to install stork without getting porx to be merged.

We really don't need to wait for the spawn. Since, in case of AKS, application yaml specs are not referring to PX storage class, event though PX is installed on AKS, applications to be backed up and restored does not use PX.

stgleb:
1. https://jenkins.portworx.co/job/tp-backup-gleb/27 this is my test on ESXi machines with multiple clusters using PX volume
2. https://jenkins.portworx.co/job/PX-Backup/job/gleb-aks-backup-job/82/ AKS on multiple clusters without PX volumes

drivers/scheduler/k8s/specs/elasticsearch-aks/px-elasticdata-app.yaml
stgleb: Yes, but let it be subject of separate PR. I don't want this PR to become too big.

vinayakshnd: Sure

tests/backup/backup_test.go
```diff
@@ -5,17 +5,28 @@ import (
...
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/libopenstorage/openstorage/pkg/sched"
+	driver_api "github.com/portworx/torpedo/drivers/api"
+	"github.com/portworx/torpedo/drivers/scheduler/spec"
+	appsapi "k8s.io/api/apps/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
vinayakshnd, line 21: Lets remove blank lines

stgleb: done reorganizing imports

tests/backup/backup_test.go
```diff
@@ -340,6 +617,71 @@ func DeleteS3Bucket(bucketName string) {
...
+// CreateCloudCredential creates cloud credetials
+func CreateCloudCredential(provider, name string, orgID string) {
+	Step(fmt.Sprintf("Create cloud credential [%s] in org [%s]", name, orgID), func() {
+		logrus.Printf("Create credential name %s for org %s provider %s", name, orgID, provider)
+		backupDriver := Inst().Backup
+		switch provider {
+		case providerAws:
+			log.Printf("Create creds for aws")
+			id := os.Getenv("AWS_ACCESS_KEY_ID")
+			Expect(id).NotTo(Equal(""),
```
vinayakshnd, line 629: Can we use the function `getAWSDetailsFromEnv()` as being used in master?
https://github.com/portworx/torpedo/blob/master/tests/backup/backup_test.go#L381

PR #451, opened 2020-04-02
Modify CreateCluster() API request for px-backup

vinayakshnd:
1. Updated `github.com/portworx/px-backup-api` to latest version.
2. Fixed CreateCluster() API request for px-backup


tests/common.go
```diff
@@ -61,7 +61,7 @@ import (
 
 const (
 	// SkipClusterScopedObjects describes option for skipping deletion of cluster wide objects
-	SkipClusterScopedObjects             = "skipClusterScopedObjects"
```
stgleb, line 64: Are you sure this conforms with latest gofmt?

stgleb: which version of go do you use?

PR #452, opened 2020-04-06
PTX-2232 Added px-backup tests for portworx backend

vinayakshnd:
Test bed setup:
**Cluster 1:** PX-Backup and torpedo runs on this cluster
**Cluster 2:** Source cluster for px-backup
**Cluster 3:** Destination cluster for px-backup

Torpedo, 
1.  Deploys apps on Cluster 2*
2. Initiates backup
3. Restores apps on Cluster 3.
4. Validates restore.

***Note:** Currently apps using encrypted PVCs using K8S secrets can not be restored correctly due to 
PB-119

**Jenkins Job with backup test enabled:**
https://jenkins.portworx.co/job/Torpedo/job/vinayak-tp-backup/57/console

**Jenkins Job with backup test disabled:**
https://jenkins.portworx.co/job/Torpedo/job/vinayak-tp-backup/58/console

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -217,6 +217,8 @@ func (d *portworx) Init(sched string, nodeDriver string, token string, storagePr
 }
 
 func (d *portworx) RefreshDriverEndpoints() error {
+	// Force update px endpoints
+	d.refreshEndpoint = true
 	storageNodes, err := d.getStorageNodesOnStart()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -217,6 +217,8 @@ func (d *portworx) Init(sched string, nodeDriver string, token string, storagePr
 }
 
 func (d *portworx) RefreshDriverEndpoints() error {
+	// Force update px endpoints
```
stgleb, line 220: What this  options is stands for? Why we need this change?

vinayakshnd: @stgleb RefreshDriverEndpoints() function we are using it here to update list of storage driver endpoints since we want to switch to a different PX cluster. If this variable is not set to true, further function calls skip updating list of storage nodes.
For e.g., If we trace function calls further, we are calling [getPXnodes()](https://github.com/portworx/torpedo/blob/d95ca6922716b1886015d10ed5af3d7c5a4db9fb/drivers/volume/portworx/schedops/k8s-schedops.go#L728) which in turn calls getNodeManager() which will return existing nodeManager (manager to current cluster) instead of returning a new nodeManager for cluster we want to switch to.

PR #464, opened 2020-04-24
PTX-2229 Added autopilot upgrade test

mborodin-px:


thiguetta:
@mborodin-px have we tested this somewhere? code looks fine, just rebase against master before merging

drivers/scheduler/k8s/k8s.go
```diff
@@ -3001,6 +3000,43 @@ func (k *K8s) ValidateAutopilotEvents(ctx *scheduler.Context) error {
 	return nil
 }
 
+// UpgradeAutopilot upgrades autopilot by changing image name
+func (k *K8s) UpgradeAutopilot(image string, opts *scheduler.UpgradeAutopilotOptions) error {
```
harsh-px, line 3004: No need to create the scheduler driver API for this. This method is too specific to be added as a scheduler driver interface method and it isn't needed as Autopilot is a k8s-only product

So we can directly invoke sched-ops from autopilot test.. You can still use `api.PerformTask` from the autopilot test.

thiguetta: haven't look from this side, I agree with @harsh-px

tests/autopilot/autopilot_test.go
```diff
@@ -287,6 +290,81 @@ var _ = Describe(fmt.Sprintf("{%sRestartAutopilot}", testSuiteName), func() {
...
+				ruleEvents, err := core.Instance().ListEvents("", meta_v1.ListOptions{
+					FieldSelector: fmt.Sprintf("involvedObject.kind=AutopilotRule,involvedObject.name=%s", apRule.Name),
+				})
+				Expect(err).NotTo(HaveOccurred())
+
+				for _, ruleEvent := range ruleEvents.Items {
+					if strings.Contains(ruleEvent.Message, string(apapi.RuleStateActiveActionsInProgress)) {
+						return true, nil
+					}
+				}
```
harsh-px, line 320: check needs to be more specific

PR #466, opened 2020-04-29
Added pgbench app which becomes ready when predefine portion of IO completes

vinayakshnd:
For px-backup test, we want to run backups for apps that have some data already written and also ongoing IO. 

For that, this PR,
1. Added a new app called `pgbench-io-portion` which becomes READY when a portion of IO completes.

Jenkins Job with this PR:
https://jenkins.portworx.co/job/tp-backup-vinayak/109/console

### DIFF
--- a/tests/backup/backup_test.go
+++ b/tests/backup/backup_test.go
@@ -188,9 +188,6 @@ var _ = Describe("{BackupCreateKillStoreRestore}", func() {
 			//ValidateApplications(contexts)
 		})
 
-		// Wait for IO to run
-		time.Sleep(Inst().IOWaitTimeout)
-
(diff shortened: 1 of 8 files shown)


tests/backup/backup_test.go
```diff
@@ -371,6 +371,9 @@ var _ = Describe("{BackupCrashVolDriver}", func() {
 			ValidateApplications(contexts)
 		})
 
+		// Wait for IO to run
```
stgleb, line 374: Just a thought, since the same time.Sleep is used in BackupKillStork test we can make it configurable

vinayakshnd: Done

PR #462, opened 2020-04-21
Add AKS specs for apps

stgleb:

PR #467, opened 2020-05-04
PWX-2377 EKS specs for torpedo apps

stgleb:


stgleb:
Job Run: https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/14

deployments/deploy-ssh.sh
```diff
@@ -312,14 +320,16 @@ spec:
...
+#  affinity:
+#    nodeAffinity:
+#      preferredDuringSchedulingIgnoredDuringExecution:
+#      - weight: 100
+#        preference:
+#          matchExpressions:
+#          - key: "${K8S_VENDOR_KEY}"
+#            operator: "${K8S_VENDOR_OPERATOR}"
+#            values:
+#            - "${K8S_VENDOR_VALUE}"
```
mborodin-px, line 332: do we need this commented lines?

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
```diff
@@ -114,6 +114,7 @@ spec:
       annotations:
         volume.beta.kubernetes.io/storage-class: px-elasticdata-sc
         azure-storage-class: azure-es-sc
+        aws-storage-class: aws-sc-nginx
```
mborodin-px, line 117: I think it should be ```aws-sc-elasticsearch``` here

deployments/deploy-ssh.sh
stgleb: No, we don't. I had experimenting with different affinity policies since I can't properly label k8s node to run torpedo.

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
stgleb: thats a good point, thank you

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
```diff
@@ -112,9 +112,11 @@ spec:
   - metadata:
       name: es-data
       annotations:
-        volume.beta.kubernetes.io/storage-class: px-elasticdata-sc
         azure-storage-class: azure-es-sc
+        aws-storage-class: aws-sc-elasticsearch
```
disrani-px, line 116: There is already support for different drivers by adding specific specs in driver directories. Why have 2 ways of doing this instead of using that?

disrani-px: Support for that was added more than 3 months ago with this PR: https://github.com/portworx/torpedo/pull/382

stgleb: Well, I can spot a few problems with this approach for my case. I use portworx as a  packup driver, but in the same time volumes must be provisioned by particular cloud provider, but if I add cloud provisioner - backup won't work. Also implementing provisioner for cloud provider is in plans, but of out scope for this particular task. 
           I can see that only one provisioner from the map of provisioners is chosen, but for Azure and AWS I basically need  at least two - azure-disk and azure-file for example. But this simple mapping allows  to select only one provisioner to work with.
   So, taking this into account we decided with @harsh-px to resolve this by using annotations in PVC rather than selecting provisioners for storage classes.

stgleb:
Job passed https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/76

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
disrani-px: Not sure what you mean, we are using the same framework to test cloud provisions in the px-backup and stork tests.
Also, this is not changing the provisioner anywhere in the storageclass. Depending on the cloud to be tested it picks up the specs for the cloud from that directory. And you don't need to implement the cloud providers, they are already added for gcp, azure and aws.
If you want to have different storage classes you can add them to azure directory and use that. I don't see how you are doing that with the annotations method either.

stgleb: Okay, I put cloud specific spec to appropriate folder.But how I determine which storage class assign to each PVC?

disrani-px: In the PVC you always use a storage class with a fixed name. In each of the cloud providers' directory, you create a storage class with that name and custom parameters. You can also create other cloud-specific specs in each directory if you want to create additional pvcs for some clouds.

stgleb:
https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/83/

deployments/deploy-ssh.sh
```diff
@@ -314,12 +322,14 @@ spec:
     effect: "NoSchedule"
   affinity:
     nodeAffinity:
-      requiredDuringSchedulingIgnoredDuringExecution:
-        nodeSelectorTerms:
-        - matchExpressions:
-          - key: ${K8S_VENDOR_KEY}
-            operator: ${K8S_VENDOR_OPERATOR}
-            ${K8S_VENDOR_VALUE}
+      preferredDuringSchedulingIgnoredDuringExecution:
```
vinayakshnd, line 325: Any specific reason to change it from hard requirement to soft one here?

drivers/scheduler/k8s/specs/elasticsearch/aws-storage-class.yaml
```diff
@@ -0,0 +1,8 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: aws-sc-elasticsearch
+provisioner: kubernetes.io/aws-ebs
```
vinayakshnd, line 5: Can we have same naming convention for all storage classes?

```
00:31:18.970 INFO[2020-05-15 01:11:09] [elasticsearch] Created storage class: aws-sc-elasticsearch 

00:31:18.974 INFO[2020-05-15 01:11:09] [elasticsearch] Created storage class: azure-es-sc 

00:31:18.984 INFO[2020-05-15 01:11:09] [elasticsearch] Created storage class: px-elasticdata-sc 
```

vinayakshnd:
@stgleb can we add a debug print statement when actually stork gets killed?
From below logs it looks like stork got killed even before backup started.

```
00:39:19.817 STEP: Create backup [tp-backup-elasticsearch-backupcreaterestore-0-05-15-01h10m55s] in org [backupcreaterestore-05-15-01h10m55s] from cluster [source-cluster]

00:39:19.972 STEP: Kill stork

00:39:19.972 STEP: Wait for backup tp-backup-elasticsearch-backupcreaterestore-0-05-15-01h10m55s to complete

00:39:19.994 2020/05/15 01:19:10 backup [tp-backup-elasticsearch-backupcreaterestore-0-05-15-01h10m55s] is in [Pending] state. Waiting to become Complete Next retry in: 30s

00:39:50.010 2020/05/15 01:19:40 backup [tp-backup-elasticsearch-backupcreaterestore-0-05-15-01h10m55s] is in [InProgress] state. Waiting to become Complete Next retry in: 30s

00:40:19.999 2020/05/15 01:20:10 backup [tp-backup-elasticsearch-backupcreaterestore-0-05-15-01h10m55s] is in [InProgress] state. Waiting to become Complete Next retry in: 30s
```

drivers/scheduler/k8s/specs/elasticsearch/aws-storage-class.yaml
stgleb: yeah,this is  good point

stgleb: Done

deployments/deploy-ssh.sh
stgleb: yeah, for example there can be test environments where entire k8s cluster is dedicated to run torpedo. Also I had a problem of labeling k8s node with px/enabled false in aws due to intricating authentication process. Anyway preferred node affinity will always put torpedo pod to non px node when such node exists.

thiguetta: as per definition if we use `preferredDuringSchedulingIgnoredDuringExecution` it tries to but in case it's not possible it will use even a node which has px, which is something we don't want, we need to stick to guarantee there is a dedicated torpedo node before we deploy torpedo

drivers/backup/portworx/portworx.go
```diff
@@ -763,6 +763,7 @@ func (p *portworx) WaitForRunning(
 
 		// Check if backup in progress - stop
 		currentStatus := resp.GetBackup().GetStatus().GetStatus()
+		logrus.Debugf("Current status for backup %s is %s", req.Name, currentStatus)
```
thiguetta, line 766: do we need this debug line?

drivers/scheduler/k8s/specs/mysql/px-mysql-snapshot-clones.yaml
```diff
@@ -1,15 +0,0 @@
-##### Portworx PVC clone from VolumeSnapshot using Stork
-apiVersion: v1
-kind: PersistentVolumeClaim
```
thiguetta, line 3: why did we remove this file?

drivers/scheduler/k8s/specs/mysql/px-mysql-snapshots.yaml
```diff
@@ -1,7 +0,0 @@
-##### Portworx Stork Snapshot
-apiVersion: volumesnapshot.external-storage.k8s.io/v1
-kind: VolumeSnapshot
-metadata:
```
thiguetta, line 4: why did we remove this file?

stgleb: I'll bring it back once my new PR for spec management comes into play.

drivers/backup/portworx/portworx.go
stgleb: Well, I am not pretty sure about scheduling. If we have at least one node  that is non-px - it will always be chosen. But in case of testing backup against only single clusuter e.g. in my EKS job it is nice to have torpedo beign running in that cluster, so we can skip this hard  requirement.

deployments/deploy-ssh.sh
stgleb: Well, I have a case when I run torpedo tests on signel cluster. By default it gets labeled with px/enabled: true. So I need torpedo to run in any case.

drivers/scheduler/k8s/specs/elasticsearch/px-elasticdata-app.yaml
stgleb: @disrani-px I've made a changes for that. Now we have generic specs in app folder and a bunch of provider specific subfolders with specs. So we will never create redunant entities.

stgleb:
Job passed https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/116/

drivers/scheduler/k8s/specs/nginx/aws/aws-storage-class.yaml
```diff
@@ -0,0 +1,8 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: nginx-sc
+provisioner: kubernetes.io/aws-ebs
+parameters:
+  type: gp2
+  fsType: ext4
```
harsh-px, line 8: this doesn't look right. We might want to use an AWS storage class that supports file (not block) since multiple pods would mount the nginx volume.

drivers/scheduler/k8s/specs/nginx/px-nginx-storage.yaml
```diff
@@ -11,12 +11,10 @@ kind: PersistentVolumeClaim
...
 metadata:
   name: px-nginx-pvc
-  annotations:
-    azure-storage-class: azure-file-sc-nginx
 spec:
-  storageClassName: px-nginx-sc
+  storageClassName: nginx-sc
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
```
harsh-px, line 17: why this change?

For nginx, the PVC needs to be ReadWriteMany.

drivers/scheduler/k8s/specs/wordpress/aws/aws-storage-class.yaml
```diff
@@ -0,0 +1,17 @@
...
+  fsType: ext4
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: wordpress-sc-shared
+provisioner: kubernetes.io/aws-ebs
+parameters:
+  type: gp2
+  fsType: ext4
```
harsh-px, line 17: same comment as nginx. Does this AWS storage class allow ReadWriteMany PVCs?

tests/backup/backup_test.go
```diff
@@ -185,14 +188,22 @@ var _ = Describe("{BackupCreateKillStoreRestore}", func() {
...
+			}
+
 			ValidateApplications(contexts)
 		})
 
-		// TODO(stgleb): Parametrize this timeout
-		// Wait for some IO to run
-		time.Sleep(time.Minute * 20)
+		logrus.Info("Wait for IO to proceed\n")
+		// Wait for IO to run
```
harsh-px, line 204: nit: comment on line 204 can be removed as the log line above says the same.

tests/backup/backup_test.go
```diff
@@ -204,64 +215,44 @@ var _ = Describe("{BackupCreateKillStoreRestore}", func() {
...
+					req, BackupRestoreCompletionTimeoutMin*time.Minute,
+					RetrySeconds*time.Second)
 
-						continue
-					} else {
-						logrus.Infof("backup %s is running", backupName)
-						return true, nil
-					}
+				if err != nil {
+					logrus.Infof("backup %s wait for running err %v",
```
harsh-px, line 233: since it is in `err != nil`, change log level to error or warn.

drivers/backup/portworx/portworx.go
```diff
@@ -798,6 +798,7 @@ func (p *portworx) WaitForRunning(
 	var backupErr error
 
 	t := func() (interface{}, bool, error) {
+		logrus.Debugf("WaitForRunning inspect backup state for %s", req.Name)
```
harsh-px, line 801: is this needed or was it for debugging?

stgleb: Well, I guess it is needed to track backup which backup we are waiting for.

drivers/scheduler/k8s/specs/wordpress/aws/aws-storage-class.yaml
stgleb: This is a good point. Basically it does not. So I'll change pvc access mode to RWO.

drivers/scheduler/k8s/specs/nginx/px-nginx-storage.yaml
stgleb: Well, since aws-ebs does not support read write many we need to have least common denomitor for this or not to use this app for EKS tests untill aws bring provisioner with RWM support.

drivers/scheduler/k8s/specs/nginx/aws/aws-storage-class.yaml
stgleb: Well, installing aws file provisioner is quite a big challenge. It is not provided by aws itself and require installing it manually AFAIK. I think we can do that in separate PR if needed, so far I would either sick to RWO or skip apps with RWM pvc.

drivers/scheduler/k8s/specs/wordpress/aws/aws-storage-class.yaml
harsh-px: The right thing would be to use EFS.

https://containerjournal.com/topics/container-networking/using-ebs-and-efs-as-persistent-volume-in-kubernetes/

stgleb: okay, may I roll back changes where I've changed access modes. I'll use only specific apps for eks test and I'll bring EFS provisioner after that. Installing efs since it is non-standard provisioner can be quite tricky.

drivers/scheduler/k8s/specs/elasticsearch/csi/px-storage-class.yaml
```diff
@@ -0,0 +1,8 @@
+kind: StorageClass
```
disrani-px, line 1: This shouldn't be csi. csi is the provisioner, not the driver. You don't need a separate storage class for this. Previously if driver was pxd and provisioner was csi it would just change the povisioner from in-tree to csi.

Also, we'll be adding specs for csi with other cloud providers, so you can't just add this as the csi driver

drivers/scheduler/k8s/k8s.go
```diff
@@ -186,7 +184,7 @@ func (k *K8s) Init(schedOpts scheduler.InitOptions) error {
 		}
 	}
 
-	k.SpecFactory, err = spec.NewFactory(schedOpts.SpecDir, schedOpts.VolDriverName, k)
+	k.SpecFactory, err = spec.NewFactory(schedOpts.SpecDir, schedOpts.StorageProvisioner, k)
```
disrani-px, line 187: This needs to be driver name, not provisioner. Driver name is pxd, aws, etc. Provisioner is whether it is in-tree or csi for that driver

stgleb: storageProvisioner param of NewFactory is only used for selecting appropriate templates. We need to chose either to use portworx spec with `kubernetes.io/portworx-volume` provisioner or `pxd.portworx.com` so this param helps to resolve it. If I pass pxd I can't make any conclusion on that or I can pass them both.

drivers/scheduler/k8s/specs/elasticsearch/csi/px-storage-class.yaml
Rohit-PX: Yes we should just stick to replacing the provisioner in storage class as it's being done currently.
@harsh-px thoughts?

drivers/scheduler/k8s/k8s.go
```diff
@@ -345,13 +340,16 @@ func (k *K8s) ParseSpecs(specDir, storageProvisioner string) ([]interface{}, err
...
 }
 
-func isValidProvider(specPath string) bool {
-	for _, driver := range volume.GetVolumeDrivers() {
-		if strings.Contains(specPath, "/"+driver+"/") { // Check for directories for cloud providers, ignore files
-			return true
+func isValidProvider(specPath, storageProvisioner string) bool {
+	// Skip all storage provisioner specific spec except storageProvisioner
+	for _, driver := range volume.GetVolumeProvisioners() {
+		if driver != storageProvisioner &&
```
Rohit-PX, line 346: this would return false if driver = pxd and storageProvisioner = portworx, when it should return true

stgleb: I guess driver variable should be renamed to provisioner. I am building spec loading based on provisioner type(default porworx). I am trying to figureout with @disrani-px why is it a bad idea.

stgleb:
@Rohit-PX okay, I just wanted to have uniformal approach. But I'll do that with CSI.

drivers/scheduler/k8s/k8s.go
stgleb: @Rohit-PX I've changed that, now specs are loaded based on volume driver.

drivers/scheduler/k8s/k8s.go
stgleb: Done

stgleb:
@Rohit-PX Could you tell me why when rescan specs we use storage provisioner?
Should it be volume driver?

https://github.com/portworx/torpedo/blob/master/drivers/scheduler/k8s/k8s.go#L246

stgleb:
Passed 

Aws
https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/136/

Azure

https://jenkins.portworx.co/job/Torpedo/job/gleb-aks-backup-job/222

Non backup portworx

https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/529/

tests/common.go
```diff
@@ -802,13 +812,10 @@ func ParseFlags() {
 				logrus.Infof("Backup driver found %v", backupDriver)
 			}
 		}
-		provider, _ := os.LookupEnv("VOLUME_PROVIDER")
```
thiguetta, line 805: as we removed this where are we getting this info from?

stgleb: This env variable was introduced by https://github.com/portworx/torpedo/pull/462/files#diff-aae5bfbcffe751257212ace370c32505R798

we had volume driver portworx, but we needed to notify torpedo that it runs in other environment.It was needed because we used pxd as storage driver, but we needed to notify torpedo that it actually runs on aws or azure. Right now we pass correct info throught these variables, we don't need to duplicate it.

STORAGE_DRIVER
STORAGE_PROVISIONER

drivers/scheduler/k8s/k8s.go
```diff
@@ -82,6 +80,8 @@ const (
 	AwsStorageClassKey = "aws-storage-class"
 	// GkeStorageClassKey key for storing gke storage class
 	GkeStorageClassKey = "gke-storage-class"
+	// PortworxStorageClassKey key for storing portworx storage class
+	PortworxStorageClassKey = "portworx-storage-class"
```
disrani-px, line 84: I don't think these are being used anymore.

stgleb: done

stgleb:
AWS storage provider aws

https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/152/

AWS storage provider portworx

https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/151/

Azure storage provider azure

https://jenkins.portworx.co/job/Torpedo/job/gleb-aks-backup-job/228/

Azure storage provider portworx

https://jenkins.portworx.co/job/Torpedo/job/gleb-aks-backup-job/229/

deployments/deploy-ssh.sh
```diff
@@ -230,7 +242,7 @@ if [ -n "${K8S_VENDOR}" ]; then
         rancher)
             K8S_VENDOR_KEY=node-role.kubernetes.io/controlplane
             K8S_VENDOR_OPERATOR="In"
-            K8S_VENDOR_VALUE='values: ["true"]'
```
thiguetta, line 233: why we changed rancher? this has nothing to do with EKS

drivers/scheduler/k8s/specs/postgres/azure/azure-storage-class.yaml
```diff
@@ -2,9 +2,10 @@
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
-  name: azure-mysql-sc
```
thiguetta, line 5: shouldn't it be mysql?

deployments/deploy-ssh.sh
stgleb: by bad, I change it

stgleb: fixed

PR #470, opened 2020-05-06
add support for missing objects during spec parsing

Rohit-PX:


Rohit-PX:
Testing here: https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/492/

PR #472, opened 2020-05-07
PTX-2498 Changed the RebootOneNode test

mborodin-px:

### DIFF
--- a/tests/reboot/reboot_test.go
+++ b/tests/reboot/reboot_test.go
@@ -40,27 +40,29 @@ var _ = Describe("{RebootOneNode}", func() {
 
 		ValidateApplications(contexts)
 
-		Step("get nodes for all apps in test and reboot their nodes", func() {
+		Step("get all nodes and reboot one by one", func() {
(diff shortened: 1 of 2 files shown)


tests/reboot/reboot_test.go
```diff
@@ -144,9 +79,15 @@ var _ = Describe("{RebootAllNodes}", func() {
 							err = Inst().V.WaitDriverUpOnNode(n, Inst().DriverStartTimeout)
 							Expect(err).NotTo(HaveOccurred())
 						})
+
+						Step("validate apps", func() {
```
thiguetta, line 83: why are we validating all again? this is being validated before destroying

thiguetta: check https://github.com/portworx/torpedo/blob/eabf6abccbe3620e0cc172e814a527f0bc93dac9/tests/common.go#L368-L380

mborodin-px: done

PR #476, opened 2020-05-12
Deleted an unnecessary parameter for rabbitmq-io

mborodin-px:
Deleted an unnecessary ```--servers-startup-timeout``` parameter for rabbitmq-io which caused to rabbitmq-io pod to be in ```CrashLoopBackOff``` state

Tested on my local environment and test passed. ```rabbitmq-io``` pod was never be in the  ```CrashLoopBackOff``` state and successfully did I/O on rabbitmq stack


harsh-px:
Re-open if this is still the actual root cause

PR #479, opened 2020-05-15
Add custom app name param to schedule options

Rohit-PX:
**Why do we need this PR?**
Provide a way to provide a custom app name for deploying a spec, that way the same spec file 
can be used to deploy different apps in the same namespace

### DIFF
--- a/drivers/scheduler/scheduler.go
+++ b/drivers/scheduler/scheduler.go
@@ -88,6 +88,8 @@ type ScheduleOptions struct {
 	Scheduler string
 	// Labels is a map of {key,value} pairs for labeling spec objects
 	Labels map[string]string
+	// CustomAppName  use this custom name to deploy app, if specified
+	CustomAppName string
 }
 
 // Driver must be implemented to provide test support to various schedulers.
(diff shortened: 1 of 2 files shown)


Rohit-PX:
Testing here:  https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/507/console

drivers/scheduler/scheduler.go
```diff
@@ -88,6 +88,8 @@ type ScheduleOptions struct {
 	Scheduler string
 	// Labels is a map of {key,value} pairs for labeling spec objects
 	Labels map[string]string
+	// CustomAppName  use this custom name to deploy app, if specified
+	CustomAppName string
```
thiguetta, line 92: where does this option come from? shouldn't it have a cmd flag for this?

thiguetta:
@Rohit-PX could you please rebase against the master and rerun the test?

thiguetta:
@Rohit-PX could you check my comments above and please re-run a Jenkins job https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/ using your image, make sure it succeeded at least with these tests SetupTeardown,AppTasksDown,AppScaleUpAndDown,VolumeDriverDown,VolumeDriverAppDown,VolumeDriverDownAttachedNode,VolumeUpdate,RebootOneNode and apps elasticsearch,mysql,postgres,sysbench,nginx-sharedv4,wordpress-sharedv4 since the job you posted above is not available anymore

harsh-px:
@Rohit-PX  looks like a stable PR. Can we close?

PR #481, opened 2020-05-19
PTX-2520 Fix CSI provisioner

stgleb:


disrani-px:
Please add @Rohit-PX to PRs around provisioner changes. This PR has broken CSI tests in stork

drivers/scheduler/k8s/k8s.go
```diff
@@ -793,6 +793,12 @@ func (k *K8s) createStorageObject(spec interface{}, ns *v1.Namespace, app *spec.
 
 	if obj, ok := spec.(*storageapi.StorageClass); ok {
 		obj.Namespace = ns.Name
+
+		if strings.Contains(volume.GetStorageProvisioner(), "pxd") {
```
thiguetta, line 797: it should be GetStorageDriver instead of provisioner as per new logic change 
https://github.com/portworx/torpedo/blob/102f9f4a753cd378e8697c68fa5c3cb296056af1/drivers/volume/volume.go#L215

stgleb: @thiguetta which new logic? GetStorageDriver will give only storage driver name, but storage class needs actual name of provisioner.

PR #486, opened 2020-05-22
PTX-2543 Use storage driver name when rescan specs

stgleb:


stgleb:
CSI: https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20NextPX/job/tp-nextpx-basic-csi/121

PR #487, opened 2020-05-23
Added support for rules creation

prashanthpx:


drivers/volume/volume.go
```diff
@@ -169,7 +169,7 @@ const (
 )
 
 var (
-	volDrivers = make(map[string]Driver)
+	volDrivers   = make(map[string]Driver)
```
siva-portworx, line 172: Why this change?

prashanthpx: As aprt of make, it has done formatting and aligned this.

PR #490, opened 2020-05-27
Add support to vault secret type

thiguetta:


stgleb:
Why do we create  vault in torpedo code? Vault secrets can be created if absent in job.

thiguetta:
> Why do we create vault in torpedo code? Vault secrets can be created if absent in job.

@stgleb because torpedo manages the secret it creates, these secrets are defined in https://github.com/portworx/torpedo/blob/102f9f4a753cd378e8697c68fa5c3cb296056af1/drivers/scheduler/k8s/specs/mysql/px-mysql-storage.yaml#L1-L7 so torpedo should cope with secrets creation since it can change

thiguetta:
> move vault to portworx volume driver sched-ops.

@harsh-px tried to move to sched-ops but there we don't know basic facts like app and which secret to use, cause we define secrets inside each app specs

PR #492, opened 2020-05-29
PTX-2551 Expose portworx-service

stgleb:
portworx-service  is unaccassible for external torpedo
when we want to run test scenario with multiple clusters
we will use loadbalancer as access endpoint


drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -492,12 +492,22 @@ func isDirEmpty(path string, n node.Node, d node.Driver) bool {
...
 	svc, err := k8sCore.GetService(PXServiceName, PXNamespace)
-	if err == nil {
-		return svc.Spec.ClusterIP, nil
+
+	if err != nil {
+		return "", err
 	}
-	return "", err
+
+	// In case of portworx-service is LoadBalancer service get external IP
```
harsh-px, line 503: try using 

https://github.com/portworx/sched-ops/blob/master/k8s/core/services.go#L80

stgleb: Done

stgleb:
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/650/

PR #493, opened 2020-05-31
Increase timeout for pgbench

stgleb:
Ussually pgbench took about 19 minutes to complete data population
and pass readiness  probe. We need to increate readiness timeout
too be big enought for that.

### DIFF
--- a/tests/backup/backup_test.go
+++ b/tests/backup/backup_test.go
@@ -58,7 +58,7 @@ const (
 	defaultTimeout       = 5 * time.Minute
 	defaultRetryInterval = 5 * time.Second
 
-	appReadinessTimeout = 10 * time.Minute
+	appReadinessTimeout = 20 * time.Minute
 )
 
 var (


stgleb:
Job passed https://jenkins.portworx.co/job/Torpedo/job/gleb-eks-backup/216

tests/backup/backup_test.go
```diff
@@ -58,7 +58,7 @@ const (
 	defaultTimeout       = 5 * time.Minute
 	defaultRetryInterval = 5 * time.Second
 
-	appReadinessTimeout = 10 * time.Minute
+	appReadinessTimeout = 20 * time.Minute
```
harsh-px, line 61: let's try with a different pgbench app. @vinayakshnd  to share the app name.

vinayakshnd: `pgbench-io-portion` is the app name

PR #497, opened 2020-06-05
[PTX-2560] Added SchedulerUpgrade test for GKE

vinayakshnd:


tests/scheduler_upgrade/scheduler_upgrade_test.go
```diff
@@ -0,0 +1,143 @@
...
+					ValidateContext(ctx)
+				}
+			})
+		}
+	})
+})
+
+func ValidateClusterSize(count int64) {
+	// In multi-zone ASG cluster, node count is per zone
+	perZoneCount := count / 3
```
mborodin-px, line 90: It will fail if we run this test in non multi-zone cluster. If we assume that this test will be run in multi-zone cluster, then we have to add that restriction before we run this test.

mborodin-px: and also, we can run this test with MAX_STORAGE_NODES_PER_AZ=2 and the whole validation logic will fail here.

vinayakshnd: @mborodin-px modified test to take list of zone into account so that test would work in single zone also. 
Consideration for MAX_STORAGE_NODES_PER_AZ has been taken care of here:
https://github.com/portworx/torpedo/pull/497/files#diff-8f145a99897a3709c5c8acaa2ad504cbR105

vinayakshnd:
GKE upgrade test passed here:
https://jenkins.portworx.co/job/Torpedo/job/gke-vinayak/4/console
(**Note**: Job failed in cluster cleanup after test has passed.)

This test does not affect any other existing tests.

vinayakshnd:
Jenkins job run with single-zone GKE cluster:
https://jenkins.portworx.co/job/Torpedo/job/gke-vinayak/13/console

tests/scheduler_upgrade/scheduler_upgrade_test.go
```diff
@@ -0,0 +1,101 @@
...
+		Expect(len(upgradeHops)).NotTo(Equal(0), "No upgrade hops provided for UpgradeScheduler test")
+
+		for _, schedVersion := range upgradeHops {
+			schedVersion = strings.TrimSpace(schedVersion)
+			Step(fmt.Sprintf("start the upgrade of scheduler to version [%v]", schedVersion), func() {
+				err := Inst().N.SetClusterVersion(schedVersion, upgradeTimeoutMins)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			Step(fmt.Sprintf("wait for %s minutes for auto recovery of storeage nodes",
```
harsh-px, line 66: typo storeage

vinayakshnd: Done

thiguetta:
@vinayakshnd we need to run at least a dev-test without gke just to see if there is no regression

PR #498, opened 2020-06-09
Fix for k3s cluster

mborodin-px:
As we install portworx on master node in k3s environment, we have to update all nodes instead of only worker nodes.

Tests passed here:
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/677/ (postgres,rabbitmq,sysbench,wordpress-sharedv4)
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/671/ (elasticsearch,mysql,nginx,nginx-sharedv4)

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -239,7 +239,7 @@ func (d *portworx) RefreshDriverEndpoints() error {
 }
 
 func (d *portworx) updateNodes(pxNodes []api.StorageNode) error {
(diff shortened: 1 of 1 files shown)


thiguetta:
Why do we install Portworx in master node?

mborodin-px:
k3s deployments are normally "tiny" (minimal K8s-distro, running very few nodes)

thiguetta:
@mborodin-px why we merged this without proper testing?

PR #499, opened 2020-06-12
PTX-2579 Add IO size  to sysbench

stgleb:


thiguetta:
Why do we need this? Sysbench purpose is to run IO in a loop, so it keeps creating and cleaning infinitely, why we need to enforce size?

stgleb:
> Why do we need this PR? why do we need to enforce size?

I am currenlty in progress of adding more apps to STORK integration tests and we want to specify size of IO run against app(mysql).

PR #501, opened 2020-06-16
Fix for ValidateVolumeCleanup method for cloud runs

mborodin-px:
I have noticed that when we run torpedo on AKS, we are running ValidateVolumeCleanup function on torpedo instance which doesn't have installed portworx on it.

`2020/06/15 21:04:51 Failed to run command on: . Cause: debug pod not found in node {{ 0 0 0 0 0 STATUS_NONE map[] []    map[]  UnknownMachine {} [] 0} 10b59f59-eca3-44a6-946f-f853ebd7fb31  aks-torpedo-24102394-vmss000000 [10.240.0.7]  Worker 0 eastus false false []} Next retry in: 10s`

### DIFF
--- a/drivers/volume/portworx/schedops/k8s-schedops.go
+++ b/drivers/volume/portworx/schedops/k8s-schedops.go
@@ -408,7 +408,7 @@ func (k *k8sSchedOps) ValidateVolumeCleanup(d node.Driver) error {
 		Name:           "*portworx-volume",
 	}
 
-	for _, n := range node.GetWorkerNodes() {
+	for _, n := range node.GetStorageDriverNodes() {
 		volDirList, _ := d.FindFiles(k8sPodsRootDir, n, listVolOpts)
 		nodeToPodsMap[n.Name] = separateFilePaths(volDirList)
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -408,7 +408,7 @@ func (k *k8sSchedOps) ValidateVolumeCleanup(d node.Driver) error {
 		Name:           "*portworx-volume",
 	}
 
-	for _, n := range node.GetWorkerNodes() {
+	for _, n := range node.GetStorageDriverNodes() {
```
harsh-px, line 411: can you check if GetStorageDriverNodes return storage and storageless PX nodes (i.e nodes where PX is installed but are storageless)

mborodin-px: yes, it returns all workers and PX nodes:
```
if n.Type == TypeWorker && n.IsStorageDriverInstalled {
			nodeList = append(nodeList, n)
		}
```

PR #504, opened 2020-06-23
new volume driver: LINSTOR

chrboe:
This adds a new volume driver for LINSTOR, which is needed to be able to run the Stork integration tests against LINSTOR.

### DIFF
--- a/drivers/volume/linstor/linstor.go
+++ b/drivers/volume/linstor/linstor.go
@@ -1,0 +1,127 @@
+package linstor
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	lclient "github.com/LINBIT/golinstor/client"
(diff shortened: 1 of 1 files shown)


chrboe:
I have the LINSTOR driver for Stork ready internally and it passes all the integration tests.
@disrani-px As soon as this gets merged I can send the patch for the Stork driver

tests/common.go
```diff
@@ -56,6 +56,8 @@ import (
 	_ "github.com/portworx/torpedo/drivers/volume/aws"
 	// import azure driver to invoke it's init
 	_ "github.com/portworx/torpedo/drivers/volume/azure"
+	// import linstor driver to invoke it's init
```
disrani-px, line 59: Don't really need this since you aren't running any torpedo tests

PR #516, opened 2020-07-13
PTX-2614 Added ability to delete pods one by one

mborodin-px:
Test passed here:
https://jenkins.portworx.co/job/Torpedo/job/tp-dev-test/710/

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -92,6 +92,7 @@ const (
 
 	resizeSupportedAnnotationKey  = "torpedo.io/resize-supported"
 	autopilotEnabledAnnotationKey = "torpedo.io/autopilot-enabled"
+	deleteStrategyAnnotationKey   = "torpedo.io/delete-strategy"
 	specObjAppWorkloadSizeEnvVar  = "SIZE"
 )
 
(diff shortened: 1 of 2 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -90,9 +90,10 @@ const (
 	// DefaultTimeout default timeout
 	DefaultTimeout = 2 * time.Minute
 
-	resizeSupportedAnnotationKey  = "torpedo.io/resize-supported"
-	autopilotEnabledAnnotationKey = "torpedo.io/autopilot-enabled"
-	specObjAppWorkloadSizeEnvVar  = "SIZE"
+	resizeSupportedAnnotationKey              = "torpedo.io/resize-supported"
+	autopilotEnabledAnnotationKey             = "torpedo.io/autopilot-enabled"
+	rollingDeleteStrategyEnabledAnnotationKey = "torpedo.io/rolling-delete-strategy-enabled"
```
harsh-px, line 95: instead, call the key `"torpedo.io/delete-strategy"` and one of the supported values it takes is `rolling`. 

This will allow us to have other delete strategies.

mborodin-px: fixed

PR #523, opened 2020-07-28
[PTX-2664] Added `nfs_v4: "true"` label for Wordpress sharedv4 PVC

vinayakshnd:


vinayakshnd:
The fix was exercised by this job: https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20master/job/tp-master-basic-vol-driver/247/console

Job failed due to mysql and not related to this change.

PR #528, opened 2020-08-19
PTX-2692 fix decommission node test to wait for volume down

thiguetta:
Signed-off-by: Thiago Gonzaga <thi_gonzaga@yahoo.com.br>


drivers/volume/portworx/portworx.go
```diff
@@ -1667,6 +1667,15 @@ func (d *portworx) DecommissionNode(n *node.Node) error {
 		}
 	}
 
+	//volume driver is about to go down, need to refresh the endpoint
```
harsh-px, line 1670: fix comment before checkin

thiguetta:
closing in favor of new change to add rt-opts

PR #535, opened 2020-09-22
Storage class auth annot

Rohit-PX:
Why do we need this PR?

For auth-enabled PX clusters, security annotations must be added to storage classes under `parameters` instead of directly adding to PVC annotations or other storage object annotations


harsh-px:
@Rohit-PX the PR says annotations need to be added on the StorageClass but the PR is adding them in the MigrationSchedule. Am I missing something?

Rohit-PX:
> @Rohit-PX the PR says annotations need to be added on the StorageClass but the PR is adding them in the MigrationSchedule. Am I missing something?

I missed a commit during squash, let me correct it.

Rohit-PX:
@harsh-px I created this PR by mistakenly using an old branch. Will close this and create another.

Rohit-PX:
Was using a stale branch, create a new PR here:  https://github.com/portworx/torpedo/pull/536

PR #540, opened 2020-09-30
Longevity split triggers

vinayakshnd:


vinayakshnd:
tp-dev-test job: https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/783/console
tp-longevity job: https://jenkins.portworx.dev/job/Torpedo/view/Torpedo%20NextPX/job/tp-nextpx-longevity/83/console

tests/common.go
```diff
@@ -211,18 +211,18 @@ func ValidateContext(ctx *scheduler.Context, errChan ...*chan error) {
 			timeout = appScaleFactor * ctx.ReadinessTimeout
 		}
 
+		Step(fmt.Sprintf("wait for %s app to start running", ctx.App.Key), func() {
```
harsh-px, line 214: actually, there was a specific reason why the WaitForRunning was after the validate volumes. 

If a PVC is in pending state, the application will never come up. While test will fail regardless, it will not be able to provide the exact reason of failure (i.e the volume creation was failing). 

Since your requirement is not to return if SkipVolumeValidation is true, you can change the step below to something like.

```
if !ctx.SkipVolumeValidation {
  ValidateVolumes(ctx, errChan...)
}
```

vinayakshnd: @harsh-px done!

harsh-px (approved):

PR #543, opened 2020-10-16
PTX-2732 Add autopilot rule for pool usage mean

mborodin-px:


mborodin-px:
Closed it as I created a combined PR https://github.com/portworx/torpedo/pull/553

PR #552, opened 2020-12-05
Add daemonset object into internal Docker registry

mborodin-px:
forgot to add DaemonSet object as we use it in "debug" specs


thiguetta (approved):
LGTM

PR #555, opened 2021-01-20
[PTX-2813] Added app definitions for Cassandra and  MongoDB

vinayakshnd:


drivers/scheduler/k8s/specs/cassandra/cassandra-stress.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cassandra-stress
+  labels:
+    app: cassandra-stress
+spec:
+  replicas: 3
```
mborodin-px, line 8: why do we need 3 replicas for cassandra-stress? I think it should be one instance of cassandra-stress which will run some load using cassandra service endpoint

mborodin-px (commented):
please test that cassandra-stress pod should be in CrashLoopBackOff when cassandra service is not available

drivers/scheduler/k8s/specs/cassandra/cassandra-stress.yaml
vinayakshnd: addressed these comments in #556

vinayakshnd:
Raised another PR for this #556 .

PR #557, opened 2021-01-27
Add new vsphere node driver

Rohit-PX:
###  Why do we need this PR?
* Adding ability to interact with the ESXi VMs via the _SSH_ node drivers
* Using the govimomi API library to interact with the Vsphere SDK


drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,234 @@
...
+	// Make future calls local to this datacenter
+	f.SetDatacenter(dc)
+
+	// Find virtual machines in datacenter
+	vms, err := f.VirtualMachineList(v.ctx, "*")
+	if err != nil {
+		return fmt.Errorf("Failed to find any virtual machines on %s: %v", v.vsphereHostIP, err)
+	}
+
+	nodes := node.GetWorkerNodes()
```
Rohit-PX, line 149: Change to have all nodes including master node in the vmMap

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,234 @@
...
+
+// RebootVM reboots vsphere VM
+func (v *vsphere) RebootVM(n node.Node) error {
+	if _, ok := vmMap[n.Name]; !ok {
+		return fmt.Errorf("Could not fetch VM for node: %s", n.Name)
+	}
+
+	vm := vmMap[n.Name]
+
+	fmt.Printf("\nRebooting VM: %s  ", vm.Name())
```
Rohit-PX, line 170: logrus instead of fmt

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// ShutdownVM shutsdown the vsphere VM
+func (v *vsphere) ShutdownVM(n node.Node) error {
+	return nil
```
Rohit-PX, line 225: Implement ShutdownVM as well

drivers/scheduler/k8s/specs/mysql/pxd/px-storage-class.yaml
```diff
@@ -19,7 +19,7 @@ parameters:
   fs: "ext4"
   block_size: "512b"
```
Rohit-PX, line 20: Delete this file from the PR

tests/reboot/reboot_test.go
```diff
@@ -56,47 +56,45 @@ var _ = Describe("{RebootOneNode}", func() {
 			// Reboot node and check driver status
 			Step(fmt.Sprintf("reboot node one at a time from the node(s): %v", nodesToReboot), func() {
 				for _, n := range nodesToReboot {
-					if n.IsStorageDriverInstalled {
```
Rohit-PX, line 59: This change is not required revert this change

drivers/node/node.go
```diff
@@ -152,6 +152,15 @@ type Driver interface {
...
 	GetZones() ([]string, error)
+
+	// PowerOnVM powers VM
+	PowerOnVM(node Node) error
+
+	// PowerOffVM powers VM
+	PowerOffVM(node Node) error
+
+	// RebootVM powers VM
+	RebootVM(node Node) error
```
Rohit-PX, line 163: Use the existing RebootNode methods and override them in the vsphere driver implementation

drivers/node/vsphere/vsphere.go
Rohit-PX: Fixed.

drivers/node/vsphere/vsphere.go
Rohit-PX: Fixed in all places.

drivers/node/node.go
Rohit-PX: Fixed

drivers/node/vsphere/vsphere.go
Rohit-PX: Fixed

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,248 @@
...
+	vsphereIP    = "VSPHERE_HOST_IP"
+)
+
+const (
+	// DefaultUsername is the default username used for ssh operations
+	DefaultUsername = "root"
+	// VMReadyTimeout Timeout for checking VM power state
+	VMReadyTimeout = 3 * time.Minute
+	// VMRetryInterval interval for retry when checking VM power state
+	VMRetryInterval = 5 * time.Second
```
mborodin-px, line 39: rename it with the same name as above VMReadyRetryInterval

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,248 @@
...
+
+const (
+	vsphereUname = "VSPHERE_USER"
+	vspherePwd   = "VSPHERE_PWD"
+	vsphereIP    = "VSPHERE_HOST_IP"
+)
+
+const (
+	// DefaultUsername is the default username used for ssh operations
+	DefaultUsername = "root"
```
mborodin-px, line 35: rename it to DefaultVsphereUsername as this is a username for vSphere

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,248 @@
...
+func (v *vsphere) Init() error {
+	logrus.Infof("Using the vsphere node driver")
+	v.SSH.Init()
+
+	username := os.Getenv(vsphereUname)
+	if len(username) == 0 {
+		v.vsphereUsername = DefaultUsername
+	} else {
+		v.vsphereUsername = username
+	}
```
mborodin-px, line 70: this construction will be more readable if you use like this:
```
	v.vsphereUsername = DefaultUsername
	username := os.Getenv(vsphereUname)
	if username != "" {
		v.vsphereUsername = username
	}
```

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,248 @@
...
+				Cause: fmt.Sprintf("Failed to test connection to VM: %s Current Status: %v, error: %v", vm.Name(), powerState, err),
+			}
+		}
+		return nil, false, nil
+	}
+	if _, err := task.DoRetryWithTimeout(t, VMReadyTimeout, VMRetryInterval); err != nil {
+		return err
+	}
+	return err
+}
```
mborodin-px, line 111: The reason for this method is not just check if node is up, but also it is running. The VM could be in boot process and can't execute any commands. Add the execution of simple command like date or hostname as we use in ssh driver.

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,248 @@
...
+	return nil
+}
+
+// RebootVM reboots vsphere VM
+func (v *vsphere) RebootNode(n node.Node, options node.RebootNodeOpts) error {
+	if _, ok := vmMap[n.Name]; !ok {
+		return fmt.Errorf("Could not fetch VM for node: %s", n.Name)
+	}
+
+	vm := vmMap[n.Name]
```
mborodin-px, line 168: I would implement getVmByName and get a VM from vSphere in runtime, then you don't need to keep vmMap

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,248 @@
...
+			Cause: fmt.Sprintf("failed to reboot VM %s. cause %v", vm.Name(), err),
+		}
+	}
+	return nil
+}
+
+// PowerOnVM powers on the VM if not already on
+func (v *vsphere) PowerOnVM(n node.Node) error {
+	var err error
+	vm := vmMap[n.Name]
```
mborodin-px, line 184: see comment above

drivers/node/vsphere/vsphere.go
```diff
@@ -0,0 +1,248 @@
...
+
+	logrus.Infof("Powering on VM: %s  ", vm.Name())
+	tsk, err := vm.PowerOn(v.ctx)
+	if err != nil {
+		logrus.Errorf("Failed to power off %s: %v", vm.Name(), err)
+	} else {
```
mborodin-px, line 190: redundant else here

drivers/node/vsphere/vsphere.go
Rohit-PX: Since this is already in the vsphere driver, I think we need to call it `DefaultVsphereUsername` same convention is being followed in the ssh driver

drivers/node/vsphere/vsphere.go
Rohit-PX: makes sense, fixed.

drivers/node/vsphere/vsphere.go
Rohit-PX: Fixed.

drivers/node/ssh/vsphere/vsphere.go
```diff
@@ -0,0 +1,175 @@
...
+			tsk, err := vm.PowerOn(ctx)
+			if err != nil {
+				fmt.Printf("Failed to power off %s: %v", vm.Name(), err)
+			} else {
+				if _, err := tsk.WaitForResult(ctx); err != nil {
+					fmt.Printf("failed to wait for VM %s to be powered off. cause %v", vm.Name(), err)
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
Rohit-PX: This was from the old code as I was still rebasing, fixed now.

drivers/node/ssh/vsphere/vsphere.go
```diff
@@ -0,0 +1,175 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	c, err := govmomi.NewClient(ctx, u, true)
+	if err != nil {
+		return fmt.Errorf("logging in error: %s", err.Error())
+	}
+
+	fmt.Printf("Log in successful. \nList of VMs on %s:\n", HostIP)
```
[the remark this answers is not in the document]
Rohit-PX: This was from the old code as I was still rebasing, fixed now.

drivers/node/ssh/vsphere/vsphere.go
```diff
@@ -0,0 +1,175 @@
...
+	return err
+}
+
+func (v *Vsphere) connect(n node.Node) error {
+
+	HostIP := "70.0.0.197"
+	//protocol := "https://"
+
+	//login := "https://administrator@vsphere.local:Password1@70.0.75.173/sdk"
+	login := fmt.Sprintf("https://root:P@$$w0rd!@%s/sdk", HostIP)
```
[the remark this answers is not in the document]
Rohit-PX: This was from the old code as I was still rebasing, fixed now.

drivers/node/ssh/vsphere/vsphere.go
```diff
@@ -0,0 +1,175 @@
+package vsphere
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"text/tabwriter"
+	//	"strings"
+	//	"time"
```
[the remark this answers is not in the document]
Rohit-PX: This was from the old code as I was still rebasing, fixed now.

drivers/node/vsphere/vsphere.go
Rohit-PX: Added execution of hostname command once it's verified that VM is powered on.

drivers/node/vsphere/vsphere.go
Rohit-PX: I would say, it makes sense to have an in-memory object that contains only the VMs being used in this test rather than querying the vSphere every time. Currently there is a similar data structure nodeMap being used for the same reason.

drivers/node/vsphere/vsphere.go
Rohit-PX: Replied above.

drivers/node/vsphere/vsphere.go
Rohit-PX: yes, I need to return an error 
```if err != nil```

thiguetta (approved):
LGTM

PR #526, opened 2020-08-18
pxcentral installation and removing using helm

diptiranjanpx:
This patch contains
1. Helm chart installation and uninstallation.
2. Pxcentral installation and removal using the above.
3. Validating K8s Jobs.

Private job. - https://jenkins.portworx.co/job/Torpedo/job/central-dipti/2/

PR #563, opened 2021-03-15
[OC-590] test px-central installation using helm

pure-jliao:
* cherry-pick @diptiranjanpx's PR [here](https://github.com/portworx/torpedo/pull/526)
* 3 tests added: 
    * `InstallCentral` installs 1.2.x px-backup, px-license-server, px-monitor
    * `InstallCentralWithoutBackup` installs px-central only with backup disabled
    * `UpgradeCentralSingle` installs everything in `InstallCentral` first then upgrade to px-central 1.3.0 single chart
* before run, configmaps need to be created for each app: px-central.yaml, px-license-server.yaml, px-monitor.yaml, helm repo url and values will be provided in configmaps, check example in a [passed build here](https://jenkins.portworx.dev/job/Torpedo/job/helm-central-jliao/25/)
* last commit hide all vendor updates and will be deleted before merging

Issues and workarounds:
* due to a helm sdk bug, some resources will be created in default namespace during helm upgrade, could cause installation failures, so upgrade test will be running in default namespace directly (affect torpedo test only, not cli users), and manual cleaned required before running any px-central tests again.
* due to another helm sdk issue, extra steps are added (more annotation and labels added) during upgrade test based on `migration.sh` script (only affect torpedo but not cli users as well)


drivers/scheduler/k8s/k8s.go
```diff
@@ -534,6 +608,12 @@ func (k *K8s) Schedule(instanceID string, options scheduler.ScheduleOptions) ([]
 			return nil, err
 		}
 
+		helmSpecObjects, err := k.HelmSchedule(app, appNamespace, options)
+		if err != nil {
+			return nil, err
+		}
+
+		specObjects = append(specObjects, helmSpecObjects...)
```
mborodin-px, line 616: I think this will fail for existing tests as thy don't use helm. Try run SetupTearDown test on your branch to see if it works.

tests/pxcentral/pxcentral_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+var _ = AfterSuite(func() {
+	PerformSystemCheck()
+	ValidateCleanup()
+})
+
+func TestMain(m *testing.M) {
+	ParseFlags()
+	os.Exit(m.Run())
+}
```
mborodin-px, line 173: sorry, but I think the installation of Helm is not Torpedo responsibility. We use spawn to install any apps on environments. @harsh-px what are your thoughts?

mborodin-px (commented):
also, I woluld suggest to create separate PR for updating vendors.

drivers/scheduler/k8s/k8s.go
pure-jliao: existing tests won't use helm, so they should work fine. that's being said if the app doesn't have a helm spec, it won't be scheduled through helm.

tests/pxcentral/pxcentral_test.go
pure-jliao: the goal is to use Torpedo to test helm install an application, the app installed through helm is treated the same as other apps like mysql, only difference is that it has a different AppSpec, which Torpedo would recognize and schedule a helm install for it.

pure-jliao:
> also, I woluld suggest to create separate PR for updating vendors.

I understand but this PR contains commits from another PR by @diptiranjanpx and comments in that PR were already addressed. I should have done a checkout from latest master and cherry-pick those commits instead of a down merge from master, now it's pretty painful to resolve the conflicts.. I'll try to resolve this. but so far all changed files are above all the vendors.

nikolaypopov:
I think the installation of PX components and PX should not be done in Torpedo, it is already done in Spawn and we should just add installing these PX components in Spawn and then hand over ready environment to Torpedo to run application tests.

harsh-px:
> I think the installation of PX components and PX should not be done in Torpedo, it is already done in Spawn and we should just add installing these PX components in Spawn and then hand over ready environment to Torpedo to run application tests.

@nikolaypopov we have crossed that bridge already. The approach here is we are considering central as not a PX component but rather a stateful application. And we want to exercise it under the same testing scenarios as other statefulset applications in torpedo.

tests/pxcentral/pxcentral_test.go
harsh-px: @mborodin-px if this was discussed offline, let's resolve the conversation with a comment here. Same for other open comments in the PR. Thanks!

tests/pxcentral/pxcentral_test.go
```diff
@@ -0,0 +1,535 @@
...
+
+		Step("Install px-central with px-backup disabled then validate", func() {
+			context = installPxcentral(&centralOptions, nil, nil, false, false)
+		})
+
+		Step("destroy apps", func() {
+			destroyPxcentral(context)
+		})
+	})
+})
```
mborodin-px, line 525: can you move these tests to the top of the file

tests/pxcentral/pxcentral_test.go
```diff
@@ -0,0 +1,535 @@
...
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("not found"))
+		logrus.Infof("job %s deleted", obj.Name)
+	} else {
+		return errors.New("resource type not implemented")
+	}
+
+	return nil
+}
+
```
mborodin-px, line 203: we can move this implementation under the scheduler driver, just add two methods DeleteNamespace and DeleteJob and then call them in your test

PR #568, opened 2021-04-09
PWX-19384: Add specs for sharedv4 with service

adityadani:
Add specs for deploying sharedv4 apps nginx & wordpress with the new service mode. 

Closes #

### DIFF
--- a/drivers/scheduler/k8s/specs/wordpress-sharedv4-svc/storage.yaml
+++ b/drivers/scheduler/k8s/specs/wordpress-sharedv4-svc/storage.yaml
@@ -1,0 +1,51 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: portworx-sc-cms-repl3-sharedv4-svc
+provisioner: kubernetes.io/portworx-volume
+parameters:
+  repl: "3"
+  priority_io: "high"
+  sharedv4: "true"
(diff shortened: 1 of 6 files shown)


drivers/scheduler/k8s/specs/wordpress-sharedv4-svc/storage.yaml
```diff
@@ -0,0 +1,51 @@
...
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 1Gi
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: mysql-pvc-1-sharedv4
```
pureneelesh, line 43: SC portworx-sc-db-repl3 is not using sharedv4. So, we should get rid of "sharedv4" in mysql-pvc-1-sharedv4?

pureneelesh (approved):

mborodin-px (approved):
LGTM

drivers/scheduler/k8s/specs/wordpress-sharedv4-svc/storage.yaml
adityadani: Torpedo runs with multiple apps and volume types at the same time. I wanted to group all the PVCs used together for an app that tests sharedv4 volume. Let me change this to `mysql-pvc-wordpress` instead.

PR #572, opened 2021-04-22
PTX-3187 Need to be able to pass spec dir to nodeDriver.Init()

nikolaypopov:
Need to be able to pass spec dir to nodeDriver.Init()

### DIFF
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -51,6 +51,7 @@ type SSH struct {
 	password  string
 	keyPath   string
 	sshConfig *ssh_pkg.ClientConfig
+	specDir   string
 	// TODO keyPath-based ssh
 }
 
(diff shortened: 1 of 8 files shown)


drivers/node/ssh/ssh.go
```diff
@@ -88,14 +87,14 @@ func useSSH() bool {
...
-func (s *SSH) Init() error {
+func (s *SSH) Init(nodeOpts node.InitOptions) error {
 
 	nodes := node.GetWorkerNodes()
 	var err error
 	if useSSH() {
 		err = s.initSSH()
 	} else {
-		err = s.initExecPod()
+		err = s.initExecPod(nodeOpts.SpecDir)
```
mborodin-px, line 97: save SpecDir parameter in *SSH struct, so you don't need to pass it to initExecPod function

nikolaypopov: Done

mborodin-px (approved):

harsh-px:
build has passed on travis. But showing pending here: https://travis-ci.org/github/portworx/torpedo/builds/768120703

Travis bug. Force merging.

PR #573, opened 2021-05-01
PTX-3055 Add support for Pure FB backend

nikolaypopov:
Add support for Pure FB backend for shared and sharedv4 type volumes.

This is needed for https://portworx.atlassian.net/browse/PTX-3220

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -65,8 +65,8 @@ if [ -z "${ENABLE_STORK_UPGRADE}" ]; then
     ENABLE_STORK_UPGRADE=false
 fi
 
-if [ -z "${IS_FLASH_BLADE_PROXY_VOLUMES}" ]; then
-    IS_FLASH_BLADE_PROXY_VOLUMES=false
+if [ -z "${IS_PURE_VOLUMES}" ]; then
+    IS_PURE_VOLUMES=false
 fi
 
 if [ -n "${PROVISIONER}" ]; then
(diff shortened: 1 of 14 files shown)


drivers/scheduler/k8s/specs/wordpress-fb-nfs4/storage.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: wordpress-sc-fb-nfs4
+provisioner: kubernetes.io/portworx-volume
+parameters:
+  backend: "pure_file"
+mountOptions:
+- nfsvers=4.1
```
adityadani, line 9: @pure-jliao Is this always needed? Do we need to test for other versions as well ?

adityadani (approved):

drivers/scheduler/k8s/specs/wordpress-fb-nfs4/storage.yaml
pure-jliao: right now only support `nfsvers=3` and `nfsvers=4.1`, we could leave `mountOptions` empty to use v3 by default.

deployments/deploy-ssh.sh
```diff
@@ -415,6 +419,7 @@ spec:
             "--storage-upgrade-endpoint-version=$UPGRADE_ENDPOINT_VERSION",
             "--enable-stork-upgrade=$ENABLE_STORK_UPGRADE",
             "--secret-type=$SECRET_TYPE",
+            "--flash-blade-proxy-volumes=$IS_FLASH_BLADE_PROXY_VOLUMES",
```
pure-jliao, line 422: one suggestion, in the future we will support FA proxy volume as well, and our nsm lib is able to arrange request to either FA or FB based on the storage class, so probably a flag like `pure-volumes` might be better?

pure-jliao (approved):
looks good, only one suggestion regarding the flag naming.

deployments/deploy-ssh.sh
nikolaypopov: Done

drivers/scheduler/k8s/specs/wordpress-fb-nfs4/storage.yaml
nikolaypopov: This spec is specifically called `wordpress-fb-nfs4`, so its ok to leave the mount option, the default `wordpress` spec will use the default nfs3 as there is no `mountOptions` specified.

PR #580, opened 2021-06-14
[PTX-3449] Continue longevity in case of test failures

vinayakshnd:
**Changes as part of this PR**:
1.  Added CPU/Mem resource limits for all the apps deployed as part of torpedo tests
2. Longevity requires tests to continue and not to stop because if some Ginkgo assertion fails. For this, introduced `*chan Error` to our library functions. If error channel is supplied to these library func,  function will return error over channel and not fail the assertions. For other test suites, this `*chan Error` is not supplied and these library functions will fail the assertions as usual.
3. Added  new `TriggerDeployNewApps` trigger.
4. Changed time intervals for longevity triggers.

Closes #

tp-dev-test successful run:
https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/922/

It ran below set of tests successfully:

SetupTeardown,AppTasksDown,VolumeDriverDown,VolumeDriverAppDown,VolumeDriverCrash,RebootOneNode,VolumeUpdate

### DIFF
--- a/drivers/scheduler/k8s/specs/cassandra/cassandra-stress.yaml
+++ b/drivers/scheduler/k8s/specs/cassandra/cassandra-stress.yaml
@@ -48,6 +48,13 @@ spec:
       containers:
       - name: cassandra-stress
         image: scylladb/scylla:4.1.11
+        resources:
+          limits:
+            cpu: "2000m"
+            memory: 4Gi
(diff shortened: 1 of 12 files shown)


lsrinivas-pure (approved):
LGTM

PR #583, opened 2021-07-08
PB-1716: Change code to use secret instead of password str

kshithijiyer-px:
Problem:
Currently all the integration-test runs fail with the below error:
```
time="2021-07-07T08:44:52Z" level=error msg="FetchIDOfUser: json: cannot unmarshal object into Go value of type []backup.KeycloakUserRepresentation"
time="2021-07-07T08:44:52Z" level=error msg="AddRoleToUser: json: cannot unmarshal object into Go value of type []backup.KeycloakUserRepresentation"
time="2021-07-07T08:44:52Z" level=error msg="Setup failed: json: cannot unmarshal object into Go value of type []backup.KeycloakUserRepresentation"
```
This is mostly because px-central-admin password is now stored in secret in px-backup namespace.

Solution:
Adding a function GetPxCentralAdminPwd() and init() to fetch the password from the secret.

Signed-off-by: kshithijiyer-px <kiyer@purestorage.com>

Closes #PB-1716

### DIFF
--- a/drivers/backup/backup.go
+++ b/drivers/backup/backup.go
@@ -7,6 +7,7 @@ import (
 
 	api "github.com/portworx/px-backup-api/pkg/apis/v1"
 	"github.com/portworx/torpedo/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 // Image Generic struct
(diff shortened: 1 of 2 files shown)


drivers/backup/backup.go
```diff
@@ -309,3 +310,11 @@ func Get(name string) (Driver, error) {
 		Type: "BackupDriver",
 	}
 }
+
+func init() {
+	str, err := GetPxCentralAdminPwd()
+	if err != nil {
+		logrus.Errorf("Error fetching password from secret: %v", err)
+	}
```
prashanthpx, line 318: Let's also exit from here so that test aborts.
The rest looks good.

prashanthpx (approved):

drivers/backup/backup.go
kshithijiyer-px: Adding `panic()` for this.

vinayakshnd (approved):
LGTM

PR #594, opened 2021-07-28
:hammer: Added log for node type and node name in longevity email

lsrinivas-pure:
Added log for node type and node name in longevity email

Closes # PTx-3797

SignedOff-by : lsrinivas

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -507,7 +507,7 @@ func (k *K8s) parseK8SNode(n corev1.Node) node.Node {
 			region = value
 		}
 	}
-	logrus.Infof("Parsed node [%s] as Type: %s, Zone: %s, Regi0n %s", n.Name, nodeType, zone, region)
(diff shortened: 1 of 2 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -507,6 +507,7 @@ func (k *K8s) parseK8SNode(n corev1.Node) node.Node {
 			region = value
 		}
 	}
+	logrus.Infof("Parsed node [%s] as Type: %s, Zone: %s, Regi0n %s", n.Name, nodeType, zone, region)
```
vinayakshnd, line 510: Typo in `Region`

vinayakshnd (commented):
Should we also put the type of node in emails? Like table having 9 worker nodes and 1 master node. so that user knows which one is master and can login to that node if some debugging needs to be done.

drivers/scheduler/k8s/k8s.go
lsrinivas-pure: Fixed it

vinayakshnd (approved):
LGTM

PR #599, opened 2021-08-13
PTX-3865 fix an issue when vdbench tries to create a file on /

thiguetta:

This PR is required when running on OCP where pods are not allowed to create files on `/`, this add a parameter which you can give an alternate output dir


vilasd (approved):

nikolaypopov (approved):
If this has passed the torpedo tests, then it LGTM

thiguetta:
test passed here

PR #602, opened 2021-08-16
PTX-3915 add bonnie++ app

thiguetta:

This PR adds the bonnie++ app


drivers/scheduler/k8s/specs/bonnie/px-bonnie-app.yml
```diff
@@ -0,0 +1,31 @@
...
+    spec:
+      containers:
+        - name: bonnie
+          image: polinux/bonnie
+          imagePullPolicy: IfNotPresent
+          resources:
+            limits:
+              memory: 500Mi
+              cpu: 100m
+          args: ["-d", "/testdir", "-r", "256", "-x", "5400", "-b"]
```
vilasd, line 24: @thiguetta we need to templatize it to take -r (RAM size and keep it default to 500MB) - x (keep it default to 10 and  max to 5400) and also we need following params
-s size of the dataset (default is half of the PVCs size)
-n
the number of files for the file creation test. This is measured in multiples of 1024 files. This is because no-one will want to test less than 1024 files, and we need the extra space on braille displays.
If the specified number is 0 then this test will be skipped.

The default for this test is to test with 0 byte files. To use files of other sizes you can specify number:max:min:num-directories:chunk-size where max is the maximum size and min is the minimum size (both default to 0 if not specified). If minimum and maximum sizes are specified then every file will have a random size from the range min..max inclusive. If you specify a number of directories then the files will be evenly distributed amoung that many sub-directories.

If max is -1 then hard links will be created instead of files. If max is -2 then soft links will be created instead of files.

PR #604, opened 2021-08-18
[PWX-21061] add test to disallow nfsv4 for sharev4 svc volume

JimOuPure:
the test creates 3 sets of sc and pvc. each pvc are of sharedv4 service with nfsv4 enabled. The test makes sure that all 3 pvcs won't get created successfully

changes:
1. sv4-svc-nfsv4/px-storage.yaml specify 3 sets of sc and pvc. 
2. scheduler.go define a new method to get pvcs
3. k8s.go implement GetPVCs. also updated DeleteVolumes and GetVolumes to only return volumes that have IDs
4. shared_nfsv4_test.go the test code

Signed-off-by: Jim Ou <jqou@purestorage.com>

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2218,15 +2218,13 @@ func (k *K8s) DeleteVolumes(ctx *scheduler.Context, options *scheduler.VolumeOpt
 					Cause: fmt.Sprintf("[%s] Failed to get PVC: %v. Err: %v", ctx.App.Key, obj.Name, err),
 				}
 			}
-			// don't return the vol if VolumeName is empty
-			if pvcObj.Spec.VolumeName != "" {
-				vols = append(vols, &volume.Volume{
-					ID:        string(pvcObj.Spec.VolumeName),
-					Name:      obj.Name,
-					Namespace: obj.Namespace,
-					Shared:    k.isPVCShared(obj),
-				})
-			}
+
+			vols = append(vols, &volume.Volume{
+				ID:        string(pvcObj.Spec.VolumeName),
+				Name:      obj.Name,
+				Namespace: obj.Namespace,
+				Shared:    k.isPVCShared(obj),
+			})
 
 			if err := k8sCore.DeletePersistentVolumeClaim(obj.Name, obj.Namespace); err != nil {
 				if k8serrors.IsNotFound(err) {
@@ -2317,24 +2315,6 @@ func (k *K8s) DeleteVolumes(ctx *scheduler.Context, options *scheduler.VolumeOpt
 	return vols, nil
 }
 
-// GetPVCs returns all pvcs for the given context
-func (k *K8s) GetPVCs(ctx *scheduler.Context) ([]*volume.PVC, error) {
-	var pvcs []*volume.PVC
-	for _, specObj := range ctx.App.SpecList {
-		if obj, ok := specObj.(*corev1.PersistentVolumeClaim); ok {
-			pvcObj, err := k8sCore.GetPersistentVolumeClaim(obj.Name, obj.Namespace)
-			if err != nil {
-				return nil, err
-			}
-			pvcs = append(pvcs, &volume.PVC{
-				Name:  pvcObj.Name,
-				Phase: string(pvcObj.Status.Phase),
-			})
-		}
-	}
-	return pvcs, nil
-}
-
 // GetVolumes  Get the volumes
 //
 func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
@@ -2380,14 +2360,12 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 			}
 
 			for _, pvc := range pvcList.Items {
-				if pvc.Spec.VolumeName != "" {
-					vols = append(vols, &volume.Volume{
-						ID:        pvc.Spec.VolumeName,
-						Name:      pvc.Name,
-						Namespace: pvc.Namespace,
-						Shared:    k.isPVCShared(&pvc),
-					})
-				}
+				vols = append(vols, &volume.Volume{
+					ID:        pvc.Spec.VolumeName,
+					Name:      pvc.Name,
+					Namespace: pvc.Namespace,
+					Shared:    k.isPVCShared(&pvc),
+				})
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


tests/shared_nfsv4/shared_nfsv4_test.go
```diff
@@ -0,0 +1,79 @@
+package tests
```
harsh-px, line 1: I recently realized that adding new folders here is causing addition of new static test binaries each of which is 100MB. This increases the torpedo container image size.

Let's just add this shared_nfsv4_test.go file under the existing `basic` directory.

drivers/scheduler/scheduler.go
```diff
@@ -161,6 +161,9 @@ type Driver interface {
 	// DeleteVolumes will delete all storage volumes for the given context
 	DeleteVolumes(*Context, *VolumeOptions) ([]*volume.Volume, error)
 
+	// GetPVCs returns all pvcs for the given context
+	GetPVCs(*Context) ([]*volume.PVC, error)
```
harsh-px, line 165: Let's stick the Volume definition so we can keep the driver not directly tied with k8s. You should be able to use the existing `GetVolumes` with little or no modification required. If you need to add a phase or status, let's add that inside the Volume object.

harsh-px (changes requested):

drivers/scheduler/scheduler.go
JimOuPure: Both PVC and volume are just an object that wrap the information retrieved from k8s or dcos (not actually implemented). In k8s, PVC and volume are two different concepts, I am just afraid that combining them together in the GetVolumes can confuse people. Are you saying PVC is just a concept in k8s only? If that's the case, could we use a more generic name here?

tests/shared_nfsv4/shared_nfsv4_test.go
JimOuPure: thanks for letting me know, I will do that then. One concern is that, if there are some jenkins job that set up to run all basic tests, they will be broken, because this test would only succeed with the app sv4-svc-nfsv4

harsh-px: They won't break. Since the jobs have to include this test in the FOCUS_TESTS to actually run the test. Just adding to the _test.go file will not run the test.

drivers/scheduler/scheduler.go
harsh-px: Yes, PVC is a k8s only concept. So to keep this generic, we have to use the Volume object.

JimOuPure: in k8s, pvc != volume, right? Just like in this test, the volume (pv) didn't get created, but pvc was created with pending phase. if getVolumes returns both PVCs and Volumes, wouldn't that be confusing?

tests/shared_nfsv4/shared_nfsv4_test.go
JimOuPure: good to know, will change then.

drivers/scheduler/scheduler.go
pureneelesh: I suppose we could verify that GetVolumes() does not return any volumes? Then, we don't need GetPVCSs().

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,57 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/portworx/torpedo/tests"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ = Describe("{NFSv4NotSupported}", func() {
+	var contexts []*scheduler.Context
+
+	It("has to create sv4 svc volume with nfsv4, validate, and teardown apps", func() {
```
pureneelesh, line 19: nit: sv4 -> sharedv4

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,57 @@
...
+		time.Sleep(30 * time.Second)
+
+		for _, ctx := range contexts {
+			Step(fmt.Sprintf("validate: %s's pvcs", ctx.App.Key), func() {
+				validatePVCs(ctx)
+			})
+		}
+
+		opts := make(map[string]bool)
+		opts[scheduler.OptionsWaitForResourceLeakCleanup] = true
```
pureneelesh, line 37: nit: variable could be eliminated by passing the value directly to the function

drivers/scheduler/k8s/k8s.go
```diff
@@ -2218,13 +2218,15 @@ func (k *K8s) DeleteVolumes(ctx *scheduler.Context, options *scheduler.VolumeOpt
...
 			}
-
-			vols = append(vols, &volume.Volume{
-				ID:        string(pvcObj.Spec.VolumeName),
-				Name:      obj.Name,
-				Namespace: obj.Namespace,
-				Shared:    k.isPVCShared(obj),
-			})
+			// don't return the vol if VolumeName is empty
+			if pvcObj.Spec.VolumeName != "" {
```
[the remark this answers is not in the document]
pureneelesh: Why is the existing code using `pvc.Name` as the `Volume.Name` and `pv.Name` as `Volume.ID`? It seems a bit weird to me. It makes me wonder if this function actually expects all the PVCs to be returned as volumes even if they are not bound. Are we going to break other tests by not returning the PVCs which are not bound?

drivers/scheduler/scheduler.go
pureneelesh: Fixed by using GetVolumes().

drivers/scheduler/k8s/k8s.go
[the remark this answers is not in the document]
pureneelesh: Reverted all changes to the k8s.go and used the existing GetVolumes() instead.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2315,6 +2317,24 @@ func (k *K8s) DeleteVolumes(ctx *scheduler.Context, options *scheduler.VolumeOpt
 	return vols, nil
 }
 
+// GetPVCs returns all pvcs for the given context
+func (k *K8s) GetPVCs(ctx *scheduler.Context) ([]*volume.PVC, error) {
```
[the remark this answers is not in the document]
pureneelesh: Removed GetPVCs().

pureneelesh:
Ran the test locally and verified that it passed. Then, reverted Jim's porx change and verified that the test failed because NFSv4 was not rejected.

```
▒ Failure [508.434 seconds]
{NFSv4NotSupported}
/root/git/go/src/github.com/portworx/torpedo/tests/basic/shared_nfsv4_test.go:14
  validates that NFSv4 is not allowed with sharedv4 service [It]
  /root/git/go/src/github.com/portworx/torpedo/tests/basic/shared_nfsv4_test.go:17

  Sharedv4 service volume pvc-aed0a848-6606-4dfd-b390-83e648fd1232 for PVC sv4-svc-nfsv4-nfsv4notsupported-0-09-08-22h47m28s/px-pvc-sharedv4-vers4 should not have been created with NFSv4
```

JimOuPure:
this PR is no longer needed, due to PWX-21340. going to close it for now.

PR #623, opened 2021-09-14
PTX-4074 failed to get volume attached on node when px using two eth

thiguetta:

This PR fixes an issue when we're trying to check where the volume is mounted and the mount is using the secondary interface.


thiguetta:
tested here https://jenkins.portworx.dev/job/Torpedo/view/Torpedo%20NextPX/job/tp-ocp-candidate-csi/39/

mborodin-px (approved):
LGTM

PR #630, opened 2021-09-27
[PWX-21005] make sure nfs server pod restarted after failover

JimOuPure:
This test makes sure app pods scheduled to the nfs server, and validate
the pod would get restarted after nfs server failover.

this change also fix: portworx.GetNodeForVolume() didn't return the correct node
that the volume attached on

test notes: the test passed with Neelesh's fix for PWX-20953, and it failed
without that fix

### DIFF
--- a/tests/basic/shared_nfsv4_test.go
+++ b/tests/basic/shared_nfsv4_test.go
@@ -1,0 +1,173 @@
+package tests
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/portworx/torpedo/drivers/node"
+	"github.com/portworx/torpedo/drivers/scheduler"
+	"github.com/portworx/torpedo/drivers/volume"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/portworx/torpedo/tests"
+)
+
+const (
+	defaultCommandRetry   = 5 * time.Second
+	defaultCommandTimeout = 1 * time.Minute
+)
+
+var _ = Describe("{NFSServerFailover}", func() {
+	var contexts []*scheduler.Context
+
+	It("has to setup, validate, failover, make sure pods on old server got restarted, and teardown apps", func() {
+		contexts = make([]*scheduler.Context, 0)
+
+		for i := 0; i < Inst().GlobalScaleFactor; i++ {
+			contexts = append(contexts, ScheduleApplications(fmt.Sprintf("nfsserverfailover-%d", i))...)
+		}
+
+		ValidateApplications(contexts)
+
+		for _, ctx := range contexts {
+			oneMinute := time.Second * time.Duration(60)
+			var nodesNonReplica []node.Node
+			var volume *volume.Volume
+			Step("disable scheduling on non replica nodes", func() {
+				vols, err := Inst().S.GetVolumes(ctx)
+				if err != nil {
+					logrus.Infof("get volumes error %v", err)
+				}
+				volume = vols[0]
+				replicaSets, err := Inst().V.GetReplicaSets(volume)
+				if err != nil {
+					logrus.Infof("get replicaSets error %v", err)
+				}
+				var replicaNodes []string
+				for _, replicaSet := range replicaSets {
+					nodes := replicaSet.Nodes
+					replicaNodes = append(replicaNodes, nodes...)
+				}
+				allNodes := node.GetWorkerNodes()
+				for _, node := range allNodes {
+					if !contains(replicaNodes, node.VolDriverNodeID) {
+						nodesNonReplica = append(nodesNonReplica, node)
+					}
+				}
+				for _, node := range nodesNonReplica {
+					Inst().S.DisableSchedulingOnNode(node)
+				}
+
+			})
+
+			// scale down and then scale up the app, so that pods are only scheduled on replica nodes
+			Step(fmt.Sprintf("scale down app: %s to 0 ", ctx.App.Key), func() {
+				applicationScaleUpMap, err := Inst().S.GetScaleFactorMap(ctx)
+				Expect(err).NotTo(HaveOccurred())
+				for name := range applicationScaleUpMap {
+					applicationScaleUpMap[name] = int32(0)
+				}
+				err = Inst().S.ScaleApplication(ctx, applicationScaleUpMap)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			Step(fmt.Sprintf("scale up app: %s to 2, and re-enable scheduling on all nodes", ctx.App.Key), func() {
+				applicationScaleUpMap, err := Inst().S.GetScaleFactorMap(ctx)
+				Expect(err).NotTo(HaveOccurred())
+				for name := range applicationScaleUpMap {
+					applicationScaleUpMap[name] = int32(2)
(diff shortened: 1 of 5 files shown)


tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				for name := range applicationScaleUpMap {
+					applicationScaleUpMap[name] = int32(2)
+				}
+				err = Inst().S.ScaleApplication(ctx, applicationScaleUpMap)
+				Expect(err).NotTo(HaveOccurred())
+				ValidateApplications(contexts)
+				for _, node := range nodesNonReplica {
+					Inst().S.EnableSchedulingOnNode(node)
+				}
```
pureneelesh, line 89: Is this necessary?

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+				pods, err := core.Instance().GetPodsUsingPV(volume.ID)
+				Expect(err).NotTo(HaveOccurred())
+				var oldPodOnOldServer corev1.Pod
+				for _, pod := range pods {
+					if pod.Spec.NodeName == oldServer.Name {
+						oldPodOnOldServer = pod
+					}
+				}
+				// make sure there is a pod running on the old nfs server
+				Expect(&oldPodOnOldServer).NotTo(BeNil())
```
pureneelesh, line 105: This check is bogus? It will always succeed since `oldPodOnOldServer` is a local variable. Its address is going to be non-nil.

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+				for i := 0; i < 10; i++ {
+					server, err := Inst().V.GetNodeForVolume(volume, defaultCommandTimeout, defaultCommandRetry)
+					Expect(err).NotTo(HaveOccurred())
+					if server.Id != oldServer.Id {
+						logrus.Infof("nfs server failed over, new nfs server is %s [%s]", server.SchedulerNodeName, server.Addresses[0])
+						newServer = server
+						break
+					}
+					time.Sleep(oneMinute)
```
pureneelesh, line 125: We could reduce the retry interval to 10 seconds to speed things up a bit.

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+				}
+				// make sure nfs server failed over
+				Expect(newServer).NotTo(BeNil())
+				logrus.Infof("new nfs server is %v", newServer)
```
pureneelesh, line 129: should we log only the relevant info about the new server such as name, IP and maybe id?

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+				logrus.Infof("start px on old nfs server Id %v, Name %v", oldServer.Id, oldServer.Name)
+				Inst().V.StartDriver(*oldServer)
+				err = Inst().V.WaitDriverUpOnNode(*oldServer, Inst().DriverStartTimeout)
+				Expect(err).NotTo(HaveOccurred())
+				logrus.Infof("px is up on old nfs server Id %v, Name %v", oldServer.Id, oldServer.Name)
+
+				// make sure the pods on both old and new server are restarted
```
pureneelesh, line 137: We should add ValidateApplications() before this to make sure that the app is working. This will also give px some time to restart the pods.

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+					if pod.Spec.NodeName == oldServer.Name {
+						logrus.Infof("pod on old server %v, creation time %v", oldPodOnOldServer.Name, oldPodOnOldServer.CreationTimestamp)
+						logrus.Infof("After failover, pod on old server %v, creation time %v", pod.Name, pod.CreationTimestamp)
+						Expect(pod.CreationTimestamp.After(timestampBeforeFailOver)).To(BeTrue())
+					}
+					if pod.Spec.NodeName == newServer.Name {
+						logrus.Infof("After failover, pod on new server %v, creation time %v", pod.Name, pod.CreationTimestamp)
+						Expect(pod.CreationTimestamp.After(timestampBeforeFailOver)).To(BeTrue())
+					}
+				}
```
pureneelesh, line 150: we should assert that this loop found 2 pods. 

Also, this logic works only for repl=2 volumes. For repl=3 volumes, the new NFS server could be on any one of the remaining 2 replica nodes. So, the second pod may not need to be restarted. Maybe assert that it is a repl=2 volume at the top of the test so we don't waste time debugging failures for repl=3 vols.

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+			})
+		}
+
+		opts := make(map[string]bool)
+		opts[scheduler.OptionsWaitForResourceLeakCleanup] = true
+
+		for _, ctx := range contexts {
+			TearDownContext(ctx, opts)
```
pureneelesh, line 158: we can get rid of lines 154 and 155 by passing the map inline here.

```
map[string]bool{scheduler.OptionsWaitForResourceLeakCleanup: true}
```

tests/basic/shared_nfsv4_test.go
```diff
@@ -0,0 +1,173 @@
...
+					logrus.Infof("get replicaSets error %v", err)
+				}
+				var replicaNodes []string
+				for _, replicaSet := range replicaSets {
+					nodes := replicaSet.Nodes
+					replicaNodes = append(replicaNodes, nodes...)
+				}
+				allNodes := node.GetWorkerNodes()
+				for _, node := range allNodes {
+					if !contains(replicaNodes, node.VolDriverNodeID) {
```
pureneelesh, line 57: We could use map instead of slice for `replicaNodes` and get rid of the `contains` function.

tests/basic/shared_nfsv4_test.go
JimOuPure: in my tests, it normally takes a couple mins to fail over, so I think 1 min interval is appropriate

pureneelesh: > in my tests, it normally takes a couple mins to fail over, so I think 1 min interval is appropriate

Even if it took couple of minutes, we should check much more frequently because if the failover happens right after our check, we would wait for one minute unnecessarily.

In any case, my current PR reduces the failover delay to 30 seconds for sharedv4 service.

tests/basic/sharedv4_svc_test.go
```diff
@@ -34,34 +33,27 @@ var _ = Describe("{NFSServerFailover}", func() {
...
-					nodes := replicaSet.Nodes
-					replicaNodes = append(replicaNodes, nodes...)
+					for _, node := range replicaSet.Nodes {
+						nodeReplicaMap[node] = true
+					}
 				}
+				// make sure there are 2 replicas
+				Expect(len(nodeReplicaMap)).To(Equal(2))
 				allNodes := node.GetWorkerNodes()
 				for _, node := range allNodes {
```
pureneelesh: variable allNodes can be removed by just using `range ndoe.GetWorkerNdoes()`.

pureneelesh (approved):

tests/basic/shared_nfsv4_test.go
JimOuPure: that makes sense, updated sleep time to 10 seconds.

PR #638, opened 2021-10-19
NEWSTACK-148&149: High priority test case coverage for FB direct acce…

david-enli:
…ss volumes

adding test coverage for FB direct access volumes for FA/FB and portworx integration

### DIFF
--- a/drivers/node/node.go
+++ b/drivers/node/node.go
@@ -123,6 +123,9 @@ type Driver interface {
 	// RunCommand runs the given command on the node and returns the output
 	RunCommand(node Node, command string, options ConnectionOpts) (string, error)
 
+	// RunCommandWithNoRetry runs the given command on the node but with no retry
+	RunCommandWithNoRetry(node Node, command string, options ConnectionOpts) (string, error)
+
 	// ShutdownNode shuts down the given node
 	ShutdownNode(node Node, options ShutdownNodeOpts) error
 
@@ -219,6 +222,13 @@ func (d *notSupportedDriver) RunCommand(node Node, command string, options Conne
 	return "", &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "RunCommand()",
+	}
+}
+
+func (d *notSupportedDriver) RunCommandWithNoRetry(node Node, command string, options ConnectionOpts) (string, error) {
+	return "", &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "RunCommandWithNoRetry()",
 	}
 }
 
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -325,6 +325,14 @@ func (s *SSH) RunCommand(n node.Node, command string, options node.ConnectionOpt
 	return output.(string), nil
 }
 
+// RunCommandWithNoRetry runs given command on given node but with no retries
+func (s *SSH) RunCommandWithNoRetry(n node.Node, command string, options node.ConnectionOpts) (string, error) {
+	if useSSH() {
+		return s.doCmdSSH(n, options, command, options.IgnoreError)
+	}
+	return s.doCmdUsingPodWithoutRetry(n, command)
+}
+
 // FindFiles finds files from give path on given node
 func (s *SSH) FindFiles(path string, n node.Node, options node.FindOpts) (string, error) {
 	findCmd := "sudo find " + path
@@ -411,6 +419,30 @@ func (s *SSH) doCmd(n node.Node, options node.ConnectionOpts, cmd string, ignore
 		return s.doCmdSSH(n, options, cmd, ignoreErr)
 	}
 	return s.doCmdUsingPod(n, options, cmd, ignoreErr)
+}
+
+func (s *SSH) doCmdUsingPodWithoutRetry(n node.Node, cmd string) (string, error) {
+	cmds := []string{"nsenter", "--mount=/hostproc/1/ns/mnt", "/bin/bash", "-c", cmd}
+	allPodsForNode, err := k8sCore.GetPodsByNode(n.Name, execPodDefaultNamespace)
+	if err != nil {
+		logrus.Errorf("failed to get pods in node: %s err: %v", n.Name, err)
+		return "", err
+	}
+	var debugPod *v1.Pod
+	for _, pod := range allPodsForNode.Items {
+		if pod.Labels["name"] == execPodDaemonSetLabel && k8sCore.IsPodReady(pod) {
+			debugPod = &pod
+			break
+		}
+	}
+
+	if debugPod == nil {
+		return "", &node.ErrFailedToRunCommand{
+			Node:  n,
+			Cause: fmt.Sprintf("debug pod not found in node %v", n),
+		}
+	}
+	return k8sCore.RunCommandInPod(cmds, debugPod.Name, "", debugPod.Namespace)
 }
 
 func (s *SSH) doCmdUsingPod(n node.Node, options node.ConnectionOpts, cmd string, ignoreErr bool) (string, error) {
--- a/drivers/scheduler/dcos/dcos.go
+++ b/drivers/scheduler/dcos/dcos.go
@@ -19,6 +19,8 @@ import (
 	"github.com/portworx/torpedo/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
 const (
@@ -448,11 +450,35 @@ func (d *dcos) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 	}
 }
 
+func (d *dcos) GetPureVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
+	// TODO: Add implementation
+	return nil, &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "GetPureVolumes()",
+	}
+}
+
+func (d *dcos) GetPodsForPVC(pvcname, namespace string) ([]corev1.Pod, error) {
+	// TODO: Add implementation
+	return nil, &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "GetPodsForPVC()",
+	}
+}
+
 func (d *dcos) ResizeVolume(cxt *scheduler.Context, configMap string) ([]*volume.Volume, error) {
 	// TODO implement this method
 	return nil, &errors.ErrNotSupported{
 		Type:      "Function",
 		Operation: "ResizeVolume()",
+	}
+}
+
+func (d *dcos) ResizePureVolumes(cxt *scheduler.Context) error {
+	// TODO implement this method
+	return &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "ResizePureVolumes()",
 	}
 }
 
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -82,6 +82,8 @@ const (
 	ZoneK8SNodeLabel = "failure-domain.beta.kubernetes.io/zone"
 	// RegionK8SNodeLabel is node label describing region of the k8s node
 	RegionK8SNodeLabel = "failure-domain.beta.kubernetes.io/region"
+	// PureFile is the parameter in storageclass to represent FB volume
+	PureFile = "pure_file"
 )
 
 const (
@@ -817,8 +819,8 @@ func (k *K8s) createNamespace(app *spec.AppSpec, namespace string, options sched
 
 func convertStorageClassToPure(scParameters map[string]string) map[string]string {
 	if len(scParameters["shared"]) > 0 || len(scParameters["sharedv4"]) > 0 {
-		scParameters["backend"] = "pure_file" // This is needed to create volumes via Pure FB backend
-
+		scParameters["backend"] = PureFile // This is needed to create volumes via Pure FB backend
+		scParameters["exportrules"] = "*(rw)"
 		// These are the parameters that are not supported by Pure FB
 		if _, found := scParameters["shared"]; found {
 			delete(scParameters, "shared")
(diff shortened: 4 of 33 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -2385,6 +2385,76 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 	return vols, nil
 }
 
+// GetPureVolumes  Get Pure FB Volumes
+func (k *K8s) GetPureVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
+	k8sOps := k8sApps
```
pure-jliao, line 2390: nit: we have `k8sApps` in `ResizePureVolumes`, can we make it consistent?

drivers/scheduler/k8s/k8s.go
```diff
@@ -2385,6 +2385,76 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+		if obj, ok := specObj.(*corev1.PersistentVolumeClaim); ok {
+			pvcObj, err := k8sCore.GetPersistentVolumeClaim(obj.Name, obj.Namespace)
+			if err != nil {
+				return nil, err
+			}
+			scForPvc, err := k8sCore.GetStorageClassForPVC(pvcObj)
+			if err != nil {
+				return nil, err
+			}
+			fmt.Println("printing scforpvc params: " + scForPvc.Parameters["backend"])
```
pure-jliao, line 2402: is this a debugging message? if not we may need to log it properly.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2385,6 +2385,76 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+			if scForPvc.Parameters["backend"] == "pure_file" {
```
pure-jliao, line 2403: nit: we have `pure_file` at multiple places, consider to create a const for it

drivers/scheduler/scheduler.go
```diff
@@ -2,6 +2,7 @@ package scheduler
 
 import (
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
```
pure-jliao, line 5: nit: import group

drivers/volume/portworx/portworx.go
```diff
@@ -61,6 +61,9 @@ const (
 	clusterPort                 = "port"
 	remoteKubeConfigPath        = "/tmp/kubeconfig"
 	pxMinVersionForStorkUpgrade = "2.1"
+	pxctlLocalSnapshotCreate 	= "pxctl volume snapshot create %s --name %s"
```
pure-jliao, line 64: nit: ident may fail fmt check, and feel it's weird to use %s this way..

drivers/volume/portworx/portworx.go
```diff
@@ -670,6 +673,94 @@ func (d *portworx) ValidateCreateVolume(volumeName string, params map[string]str
...
+	var token string
+	token = d.getTokenForVolume(volumeName, params)
+	if val, hasKey := params["refresh-endpoint"]; hasKey {
+		refreshEndpoint, _ := strconv.ParseBool(val)
+		d.refreshEndpoint = refreshEndpoint
+	}
+	volDriver := d.getVolDriver()
+	_, err := volDriver.SnapshotCreate(d.getContextWithToken(context.Background(), token), &api.SdkVolumeSnapshotCreateRequest{VolumeId: volumeName, Name: volumeName+"_snapshot"})
+	if err != nil {
+		fmt.Printf("error when creating local snapshot is %v", err)
```
pure-jliao, line 686: can we use log instead of printf?

drivers/volume/portworx/portworx.go
```diff
@@ -670,6 +673,94 @@ func (d *portworx) ValidateCreateVolume(volumeName string, params map[string]str
...
+		return err
+	}
+	return nil
+}
+
+func (d *portworx) ValidateCreateSnapshotUsingPxctl(volumeName string) error {
+	nodes := node.GetStorageDriverNodes()
+	_, err := d.nodeDriver.RunCommandWithoutRetry(nodes[0], fmt.Sprintf(pxctlLocalSnapshotCreate, volumeName, volumeName+"-snapshot"), node.ConnectionOpts{
```
pure-jliao, line 694: better to have a function returning the cmd instead of concating the string here.

tests/common.go
```diff
@@ -2,8 +2,10 @@ package tests
 
 import (
 	"encoding/csv"
+	"errors"
 	"flag"
 	"fmt"
+	"github.com/portworx/torpedo/pkg/osutils"
```
pure-jliao, line 8: nit: import group

tests/common.go
```diff
@@ -133,6 +140,7 @@ var (
 	beEmpty      = gomega.BeEmpty
 	beNil        = gomega.BeNil
 	equal        = gomega.Equal
+	contain		 = gomega.ContainSubstring
```
pure-jliao, line 143: nit: change tab to spaces.

tests/common.go
```diff
@@ -291,7 +413,163 @@ func ValidateVolumes(ctx *scheduler.Context, errChan ...*chan error) {
...
+			err = Inst().V.ValidateCreateGroupSnapshotUsingPxctl()
+			if err != nil {
+				expect(err.Error()).To(contain(errPureFileSnapshotNotSupported.Error()))
+			}
+		})
+	})
+}
+
+// ValidateResizeFBPVC is the ginkgo spec for validating resize of volumes
+func ValidateResizeFBPVC(ctx *scheduler.Context, errChan ...*chan error) {
```
pure-jliao, line 501: seems we don't validate the new size after resizing?

tests/common.go
```diff
@@ -291,7 +413,163 @@ func ValidateVolumes(ctx *scheduler.Context, errChan ...*chan error) {
...
+		byteUsedafter, err := Inst().V.ValidateGetByteUsedForVolume(vols[0].ID, make(map[string]string))
+		fmt.Println(fmt.Sprintf("after writing random bytes to the file the byteUsed is %v", byteUsedafter))
+		err = expectedSizechange(byteUsedafter - byteUsedInitial)
+		expect(err).NotTo(haveOccurred())
+
+	})
+}
+
+func expectedSizechange (sizeChangeInBytes uint64) error {
+	if sizeChangeInBytes > 531 * 1048576 && sizeChangeInBytes < 511 * 1048576 {
```
pure-jliao, line 554: this size change is not "common" and not likely to share between test cases

tests/fb_directaccess/fb_directaccess_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/portworx/torpedo/drivers/node"
+	"github.com/portworx/torpedo/drivers/scheduler"
+	"github.com/portworx/torpedo/drivers/scheduler/k8s"
+	. "github.com/portworx/torpedo/tests"
+	"os"
+	"testing"
+)
+
+const (
+	// TODO: having the secret key checked in the code probably bad, wonder whats the best way to keep this info?"
```
pure-jliao, line 17: should be ok for now as long as the account and user are dedicated for test only.

tests/fb_directaccess/fb_directaccess_test.go
```diff
@@ -0,0 +1,128 @@
...
+	fbS3AccessKey = "PSFBSAZRDJMJDOMAPHHKBHFOOMIIOCMOEMMGCBPGML"
+	fbS3SecretKey = "85E6126CC4ec2e88c/579e+761CF35E3b8ed4feNKJF"
+	newstackFBUrl = "http://10.16.59.28"
+	establishBackupCredential = "pxctl credentials create --provider s3 --s3-access-key %s --s3-secret-key %s --s3-region us-east-1 --s3-endpoint %s --s3-storage-class STANDARD elitest"
```
pure-jliao, line 21: same here I feel it's so weird to have a const string here for string manipulation, a function would be much better.

pure-jliao (commented):
overall looks good to me as long as the tests are doing what we expect. left some comments but I'm ok with it if we want to merge this asap and give it more coverage on pipeline. One concern is whether we have pvc size validation after resize, didn't find it in the code.

drivers/node/node.go
```diff
@@ -123,6 +123,9 @@ type Driver interface {
 	// RunCommand runs the given command on the node and returns the output
 	RunCommand(node Node, command string, options ConnectionOpts) (string, error)
 
+	// RunCommandWithoutRetry runs the given command on the node but with no retry
+	RunCommandWithoutRetry(node Node, command string, options ConnectionOpts) (string, error)
```
Pure-AdamuKaapan, line 127: [nit] I might prefer `NoRetry`, cuz while I'm still waking up I misread this as `RunCommandWithRetry` (the exact opposite) for a couple files until I did a double-take :joy:

Half a serious nit to prevent misreadings, but feel totally free to leave it as-is

drivers/node/ssh/ssh.go
```diff
@@ -325,6 +325,14 @@ func (s *SSH) RunCommand(n node.Node, command string, options node.ConnectionOpt
 	return output.(string), nil
 }
 
+// RunCommandWithoutRetry runs given command on given node but with no retries
+func (s *SSH) RunCommandWithoutRetry(n node.Node, command string, options node.ConnectionOpts) (string, error) {
```
Pure-AdamuKaapan, line 329: Is there a way we can refactor `RunCommand` to call `RunCommandWithoutRetry`, or have them share a common function somehow? My main concern is that I see `RunCommand` has some special error returns (that `&node.ErrFailedToRunCommand`), and while we could just fix that specific case to make this function also return that, it would be better if we could have them share the same logic so we only have to change it in one place in the future.

drivers/scheduler/dcos/dcos.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	corev1 "k8s.io/api/core/v1"
```
Pure-AdamuKaapan, line 7: Move this to a separate import block (probably after the github.com block)

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: Agreed, this should be a log call or removed

drivers/volume/portworx/portworx.go
Pure-AdamuKaapan: I've seen format strings used like this before, though I'd prefer to see `Format` in the name of the constant. Also yeah, funky indents

drivers/volume/portworx/portworx.go
```diff
@@ -670,6 +673,94 @@ func (d *portworx) ValidateCreateVolume(volumeName string, params map[string]str
...
+		fmt.Printf("error when creating local snapshot using PXCTL is %v", err)
+		return err
+	}
+	return nil
+}
+
+func (d *portworx) ValidateCreateCloudsnap(volumeName string, params map[string]string) error {
+	var token string
+	token = d.getTokenForVolume(volumeName, params)
+	if val, hasKey := params["refresh-endpoint"]; hasKey {
```
Pure-AdamuKaapan, line 708: Is `refresh-endpoint` a constant somewhere, and if not can we make it one?

pkg/osutils/osutils.go
```diff
@@ -122,4 +122,4 @@ func Kubectl(arguments []string) error {
 	}
 
 	return nil
-}
+}
```
Pure-AdamuKaapan, line 125: There should be a newline at the end of this file, but more importantly it looks like nothing was actually changed here.

tests/common.go
```diff
@@ -291,7 +413,163 @@ func ValidateVolumes(ctx *scheduler.Context, errChan ...*chan error) {
...
+			processError(err, errChan...)
+		})
+
+		for vol, params := range vols {
+			if Inst().ConfigMap != "" {
+				params["auth-token"], err = Inst().S.GetTokenFromConfigMap(Inst().ConfigMap)
+				processError(err, errChan...)
+			}
+			if ctx.RefreshStorageEndpoint {
+				params["refresh-endpoint"] = "true"
```
Pure-AdamuKaapan, line 438: Use that `refresh-endpoint` constant I suggested to make earlier

tests/common.go
```diff
@@ -291,7 +413,163 @@ func ValidateVolumes(ctx *scheduler.Context, errChan ...*chan error) {
...
+
+		var vols map[string]map[string]string
+		Step(fmt.Sprintf("get %s app's volume's custom parameters", ctx.App.Key), func() {
+			vols, err = Inst().S.GetVolumeParameters(ctx)
+			processError(err, errChan...)
+		})
+
+		for vol, params := range vols {
+			if Inst().ConfigMap != "" {
+				params["auth-token"], err = Inst().S.GetTokenFromConfigMap(Inst().ConfigMap)
```
Pure-AdamuKaapan, line 434: Same here, is `auth-token` a constant and if not can we make it one?

tests/common.go
```diff
@@ -291,7 +413,163 @@ func ValidateVolumes(ctx *scheduler.Context, errChan ...*chan error) {
...
+			vols, err = Inst().S.GetPureVolumes(ctx)
+			processError(err, errChan...)
+		})
+		byteUsedInitial, err := Inst().V.ValidateGetByteUsedForVolume(vols[0].ID, make(map[string]string))
+		fmt.Println(fmt.Sprintf("initially the byteUsed is %v", byteUsedInitial))
+		// get the pod for this pvc
+		pods, err := Inst().S.GetPodsForPVC(vols[0].Name, vols[0].Namespace)
+
+		// write to the FB volume
+		cmdArgs := []string{"exec", "-it", pods[0].Name, "-n", pods[0].Namespace, "--", "bash","-c", "dd bs=521 count=1048576 </dev/urandom > "+ getMountPath(pods[0].Namespace)+ "/myfile"}
```
Pure-AdamuKaapan, line 539: Could we maybe split this out across multiple lines? It's quite an eyesore to read right now

Pure-AdamuKaapan: `bs` should be 512, not 521

tests/common.go
Pure-AdamuKaapan: Can we make `1048576` a constant somewhere, and if we made the block size a constant somewhere we could express this as `sizeChangeInBytes > (blockSize + 15) * 1048576` which is a little more readable/self-explanatory than `531`

tests/common.go
```diff
@@ -291,7 +413,163 @@ func ValidateVolumes(ctx *scheduler.Context, errChan ...*chan error) {
...
+}
+
+func expectedSizechange (sizeChangeInBytes uint64) error {
+	if sizeChangeInBytes > 531 * 1048576 && sizeChangeInBytes < 511 * 1048576 {
+		return errUnexpectedSizeChangeAfterFBIO
+	}
+	return nil
+}
+// getMountPath checkts the podname prefix, and finds the mountpath.
+//	unfortunately, the mountpath for PVCs all vary among applications, so Im getting the hardcoded value here
```
Pure-AdamuKaapan, line 560: Is there a way we could find it by running the `mount` command and looking for NFS mounts? Or is that too slow/this doesn't have access to running commands?

tests/fb_directaccess/fb_directaccess_test.go
Pure-AdamuKaapan: Torpedo should have things in place for reading k8s secrets, could we put them in there? These really really really should not be here

tests/fb_directaccess/fb_directaccess_test.go
Pure-AdamuKaapan: Yeah, this is a bit of an extreme case where I agree. Also, `elitest` should probably be a constant, and probably changed to something else more explanatory.

Pure-AdamuKaapan (changes requested):
Please remove credentials and IP address, they cannot be in here and should be provided to the tests through a secret. Also note most of Jiafeng's comments are still open and I left a few more RE constants and formatting.

drivers/node/ssh/ssh.go
```diff
@@ -413,6 +421,30 @@ func (s *SSH) doCmd(n node.Node, options node.ConnectionOpts, cmd string, ignore
 	return s.doCmdUsingPod(n, options, cmd, ignoreErr)
 }
 
+func (s *SSH) doCmdUsingPodWithoutRetry(n node.Node, cmd string) (string, error) {
+	cmds := []string{"nsenter", "--mount=/hostproc/1/ns/mnt", "/bin/bash", "-c", cmd}
```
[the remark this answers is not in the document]
david-enli: not sure, but this is existing code. I had to make a function that runs command in a pod but doesnt keep retrying on failure, instead, throw the error and verify the negative cases

dingyin (commented):
Please remove credentials and IP address as Adam suggested. Others look ok. Fix nit and minor issues if you have time.

tests/common.go
david-enli: this is calling k8s own update PVC api, so I think if it didn't succeed, errors would be returend, I feel this should be sufficient for this test, but I could add a TODO note for later to actually check the PV size again

pure-jliao: yeah but even it succeeded on k8s side, the PVC could be in state of pending expansion, it's safer to add a new size check in the test. could do it in the future.

go.mod
```diff
@@ -17,15 +17,15 @@ require (
 	github.com/libopenstorage/autopilot-api v1.3.0
 	github.com/libopenstorage/cloudops v0.0.0-20210223183702-b9c6b74cbf1d
 	github.com/libopenstorage/openstorage v8.0.1-0.20210909003102-97e11e6485ad+incompatible
-	github.com/libopenstorage/stork v1.3.0-beta1.0.20210503014041-fb328e234db7
+	github.com/libopenstorage/stork v1.4.1-0.20210903185636-5a1f8a4142bf
```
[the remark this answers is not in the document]
david-enli: I'm honestly not sure, I got some conflicts on go.mod file so I ended up just copying over the latest from master branch, so I think we should be okay there

Pure-AdamuKaapan (approved):
LGTM!

dingyin (approved):
/lgtm

PR #639, opened 2021-10-19
PTX-4323: increase volume driver up timeout

pureneelesh:
One of the jobs failed because PX took 7 minutes to become ready and the job
gave up after 6 mintues. We should increase the timeout since it is
ok for PX to take longer to get ready when there are logs to be replayed.

Signed-off-by: Neelesh Thakur <neelesh.thakur@purestorage.com>

PTX-4323

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -101,7 +101,7 @@ if [ -z "${TIMEOUT}" ]; then
 fi
 
 if [ -z "$DRIVER_START_TIMEOUT" ]; then
-    DRIVER_START_TIMEOUT="12m0s"
+    DRIVER_START_TIMEOUT="10m0s"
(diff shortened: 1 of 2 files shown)


deployments/deploy-ssh.sh
```diff
@@ -101,7 +101,7 @@ if [ -z "${TIMEOUT}" ]; then
 fi
 
 if [ -z "$DRIVER_START_TIMEOUT" ]; then
-    DRIVER_START_TIMEOUT="5m0s"
+    DRIVER_START_TIMEOUT="12m0s"
```
vilasd, line 104: Lets keep this to 10 mins for now!! in case Portworx comes up after 10 mins and before 10 mins test will resume and will hit issues in applications. Its better to have timeout at 10 mins then we know it failed because of slow restart of node.

PR #645, opened 2021-11-08
retry for scale application in sharedv4_svc_test

JimOuPure:
Inst().S.ScaleApplication fails sometime, added retry to make the test more stable

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2853,9 +2853,8 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
 				dep, err = k8sOps.UpdateDeployment(dep)
 				if err == nil {
 					return dep, false, nil // succeeded, no retry
-				} else {
-					return "", true, err // failed to update deployment, retry
 				}
+				return "", true, err // failed to update deployment, retry
 			}
 			_, err := task.DoRetryWithTimeout(t, 2*time.Minute, 10*time.Second)
 			if err != nil {
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2842,22 +2842,22 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
 		}
 		if obj, ok := specObj.(*appsapi.Deployment); ok {
 			logrus.Infof("Scale all Deployments")
-			dep, err := k8sOps.GetDeployment(obj.Name, obj.Namespace)
-			if err != nil {
-				return err
-			}
 			newScaleFactor := scaleFactorMap[obj.Name+DeploymentSuffix]
-			*dep.Spec.Replicas = newScaleFactor
 
 			t := func() (interface{}, bool, error) {
-				dep, err := k8sOps.UpdateDeployment(dep)
+				dep, err := k8sOps.GetDeployment(obj.Name, obj.Namespace)
+				if err != nil {
+					return "", false, err // do not retry if we fail to get dep
+				}
+				*dep.Spec.Replicas = newScaleFactor
+				dep, err = k8sOps.UpdateDeployment(dep)
 				if err == nil {
-					return dep, false, nil
+					return dep, false, nil // succeeded, no retry
 				} else {
-					return "", true, err
+					return "", true, err // failed to update deployment, retry
 				}
 			}
-			_, err = task.DoRetryWithTimeout(t, 2 * time.Minute, 10 * time.Second)
+			_, err := task.DoRetryWithTimeout(t, 2*time.Minute, 10*time.Second)
 			if err != nil {
 				return &scheduler.ErrFailedToUpdateApp{
 					App:   ctx.App,
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2848,7 +2848,17 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
 			}
 			newScaleFactor := scaleFactorMap[obj.Name+DeploymentSuffix]
 			*dep.Spec.Replicas = newScaleFactor
-			if _, err := k8sOps.UpdateDeployment(dep); err != nil {
+
+			t := func() (interface{}, bool, error) {
+				dep, err := k8sOps.UpdateDeployment(dep)
+				if err == nil {
+					return dep, false, nil
+				} else {
+					return "", true, err
+				}
+			}
+			_, err = task.DoRetryWithTimeout(t, 2 * time.Minute, 10 * time.Second)
+			if err != nil {
 				return &scheduler.ErrFailedToUpdateApp{
 					App:   ctx.App,
 					Cause: fmt.Sprintf("Failed to update Deployment: %v. Err: %v", obj.Name, err),
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2847,7 +2847,7 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
 			t := func() (interface{}, bool, error) {
 				dep, err := k8sOps.GetDeployment(obj.Name, obj.Namespace)
 				if err != nil {
-					return "", false, err // do not retry if we fail to get dep
+					return "", true, err // failed to get deployment, retry
 				}
 				*dep.Spec.Replicas = newScaleFactor
 				dep, err = k8sOps.UpdateDeployment(dep)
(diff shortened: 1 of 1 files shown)


tests/basic/sharedv4_svc_test.go
```diff
@@ -63,8 +63,15 @@ var _ = Describe("{NFSServerFailover}", func() {
 				for name := range applicationScaleUpMap {
 					applicationScaleUpMap[name] = int32(0)
 				}
-				err = Inst().S.ScaleApplication(ctx, applicationScaleUpMap)
-				Expect(err).NotTo(HaveOccurred())
+				var scaleErr error
+				for i := 0; i < 3; i++ {
+					scaleErr = Inst().S.ScaleApplication(ctx, applicationScaleUpMap)
```
pureneelesh, line 68: We should fix this in K8s) ScaleApplication() function. Catch the relevant k8s error and retry. That way, every job benefits from this fix.

JimOuPure: good point, will do

JimOuPure: I used wait.ExponentialBackoff in the px change. However, that module is not in torpedo, so I didn't use it here. Let me know if you think it is worth adding that module to torpedo.

pureneelesh:
Please also add one of the QE folks to the review.

tests/basic/sharedv4_svc_test.go
pureneelesh: LGTM

JimOuPure:
> Please also add one of the QE folks to the review.

Added @nikolaypopov and @mborodin-px as reviewers

drivers/scheduler/k8s/k8s.go
```diff
@@ -2848,10 +2848,22 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
...
+				_, updateError := k8sOps.UpdateDeployment(dep)
+				if updateError == nil {
+					break
+				}
+				if i < retries - 1 {
+					logrus.Warnf("Failed to update deployment. Error: %v. Going to retry", updateError)
+					time.Sleep(time.Second * 10)
+				}
+			}
+			if updateError != nil {
```
mborodin-px, line 2863: use task.DoRetryWithTimeout function to do retries.

JimOuPure: thanks for the suggestion, will change

nikolaypopov:
I think you should use task.DoRetryWithTimeout() function here. It is widely used inside torpedo for any retry, take a look at it.

JimOuPure:
I looked into the test, we scaled down the app to 0, and immediately scale it back up to 2. That might be the reason why ScaleApplication failed. So instead of retrying in ScaleApplication, we should probably validate the scale down finished first, and then scale it back up. Will update this PR based on that.

JimOuPure:
> I looked into the test, we scaled down the app to 0, and immediately scale it back up to 2. That might be the reason why ScaleApplication failed. So instead of retrying in ScaleApplication, we should probably validate the scale down finished first, and then scale it back up. Will update this PR based on that.

I looked into the torpedo code, there is no existing method to check if the scaling process of deployments is done, and it seems to be tricky to create one as scheduler.Driver doesn't have the concept of deployment. 
This issue doesn't happen most of the time, adding a 20 seconds sleep time between 2 scaling calls should resolve it, is this work around fine, @pureneelesh? If this work around doesn't resolve the issue, we could invest more time to validate the deployment.

pureneelesh:
> I looked into the torpedo code, there is no existing method to check if the scaling process of deployments is done, and it seems to be tricky to create one as scheduler.Driver doesn't have the concept of deployment. This issue doesn't happen most of the time, adding a 20 seconds sleep time between 2 scaling calls should resolve it, is this work around fine, @pureneelesh? If this work around doesn't resolve the issue, we could invest more time to validate the deployment.

The test should ensure that the change to the scale factor has taken effect before proceeding. Otherwise, we will keep seeing intermittent issues. 

Regardless of that, we need to retry the "update deployment" k8s call which is a separate issue. I would suggest that we fix the error that the job failed with by retrying the update.

Then, create a new ticket to fix the test to ensure that the deployment change has taken effect.

JimOuPure:
> > I looked into the torpedo code, there is no existing method to check if the scaling process of deployments is done, and it seems to be tricky to create one as scheduler.Driver doesn't have the concept of deployment. This issue doesn't happen most of the time, adding a 20 seconds sleep time between 2 scaling calls should resolve it, is this work around fine, @pureneelesh? If this work around doesn't resolve the issue, we could invest more time to validate the deployment.
> 
> The test should ensure that the change to the scale factor has taken effect before proceeding. Otherwise, we will keep seeing intermittent issues.
> 
> Regardless of that, we need to retry the "update deployment" k8s call which is a separate issue. I would suggest that we fix the error that the job failed with by retrying the update.
> 
> Then, create a new ticket to fix the test to ensure that the deployment change has taken effect. I think it would be better to not rely on a fixed amount of sleep. We should implement something more robust.

all right, that sounds reasonable, will do that.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2848,7 +2848,17 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
 			}
 			newScaleFactor := scaleFactorMap[obj.Name+DeploymentSuffix]
 			*dep.Spec.Replicas = newScaleFactor
-			if _, err := k8sOps.UpdateDeployment(dep); err != nil {
+
+			t := func() (interface{}, bool, error) {
+				dep, err := k8sOps.UpdateDeployment(dep)
```
pureneelesh, line 2853: I don't think this will handle the resourceVersion change. It will keep failing even with the retries. You have to do a "get" on the deployment again and modify the new object which will have the latest resourceVersion.

JimOuPure: ah right, will change

drivers/scheduler/k8s/k8s.go
```diff
@@ -2842,13 +2842,23 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
...
-				return err
-			}
 			newScaleFactor := scaleFactorMap[obj.Name+DeploymentSuffix]
-			*dep.Spec.Replicas = newScaleFactor
-			if _, err := k8sOps.UpdateDeployment(dep); err != nil {
+
+			t := func() (interface{}, bool, error) {
+				dep, err := k8sOps.GetDeployment(obj.Name, obj.Namespace)
+				if err != nil {
+					return "", false, err // do not retry if we fail to get dep
```
pureneelesh, line 2850: why not retry this as well?

drivers/scheduler/k8s/k8s.go
```diff
@@ -2842,13 +2842,23 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
...
+				}
+				*dep.Spec.Replicas = newScaleFactor
+				dep, err = k8sOps.UpdateDeployment(dep)
+				if err == nil {
+					return dep, false, nil // succeeded, no retry
+				} else {
+					return "", true, err // failed to update deployment, retry
+				}
+			}
+			_, err := task.DoRetryWithTimeout(t, 2*time.Minute, 10*time.Second)
```
pureneelesh, line 2860: What do you think of retrying every second?

pureneelesh (approved):

drivers/scheduler/k8s/k8s.go
```diff
@@ -2848,7 +2848,17 @@ func (k *K8s) ScaleApplication(ctx *scheduler.Context, scaleFactorMap map[string
...
+
+			t := func() (interface{}, bool, error) {
+				dep, err := k8sOps.UpdateDeployment(dep)
+				if err == nil {
+					return dep, false, nil
+				} else {
+					return "", true, err
+				}
+			}
+			_, err = task.DoRetryWithTimeout(t, 2 * time.Minute, 10 * time.Second)
```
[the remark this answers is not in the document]
JimOuPure: sure, I could retry every second

drivers/scheduler/k8s/k8s.go
JimOuPure: didn't see this error in the test, so not sure if it is worth retrying. I could retry since you ask

mborodin-px:
@pureneelesh @JimOuPure why can't you add GetDeployment method into scheduler driver and use it in your test? You can implement wait loop in this case unless Replicas will be 0.

JimOuPure:
> @pureneelesh @JimOuPure why can't you add GetDeployment method into scheduler driver and use it in your test? You can implement wait loop in this case unless Replicas will be 0.

do other scheduler drivers have the concept of deployment? I thought it was only k8s' concept. I could implement that in a different PR. This PR is to make sure that we will retry in the scale method when there is error.

JimOuPure:
https://portworx.atlassian.net/browse/PWX-21982 is created to verify the scaling process of applications. will close this one for now.

PR #649, opened 2021-11-15
PTX-4456 add support to ocp fast channel

thiguetta:

This PR adds support to the explicit channel parameter when using `<channel>-<version>` i.e `fast-4.9.5`. by default `stable` channel is used when the channel is not explicitly set.

Closes #


mborodin-px (approved):
LGTM

PR #650, opened 2021-11-15
PTX-3347 add support to interleaved upgrade between OCP and PX

devops-thiago:
This PR adds support for interleaved upgrade (upgrade cluster then upgrade portworx)

Closes #

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"gopkg.in/yaml.v2"
 	"math"
 	"net/http"
+	"os/exec"
 	"reflect"
 	"regexp"
 	"strconv"
@@ -21,8 +23,11 @@ import (
 	clusterclient "github.com/libopenstorage/openstorage/api/client/cluster"
 	"github.com/libopenstorage/openstorage/api/spec"
 	"github.com/libopenstorage/openstorage/cluster"
+	operatorcorev1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	"github.com/pborman/uuid"
+	"github.com/portworx/sched-ops/k8s/apiextensions"
 	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/portworx/sched-ops/k8s/operator"
 	"github.com/portworx/sched-ops/task"
 	driver_api "github.com/portworx/torpedo/drivers/api"
 	"github.com/portworx/torpedo/drivers/node"
@@ -39,7 +44,9 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -125,6 +132,8 @@ var deleteVolumeLabelList = []string{
 }
 
 var k8sCore = core.Instance()
+var pxOperator = operator.Instance()
+var apiExtentions = apiextensions.Instance()
 
 type portworx struct {
 	legacyClusterManager  cluster.Cluster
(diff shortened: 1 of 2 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1673,16 +1682,39 @@ func (d *portworx) UpgradeDriver(endpointURL string, endpointVersion string, ena
...
+		if enableStork {
+			if err := d.upgradeStork(endpointURL, endpointVersion); err != nil {
+				return err
+			}
+		} else {
+			logrus.Infof("stork upgrade is disabled, skipping...")
+		}
 	} else {
-		logrus.Infof("stork upgrade is disabled, skipping...")
+		storageClusters, err := pxOperator.ListStorageClusters(schedops.PXNamespace)
```
nikolaypopov, line 1703: Let's use Get instead of List, only 1 `StorageCluster` should be deployed.

drivers/volume/portworx/portworx.go
```diff
@@ -1673,16 +1682,39 @@ func (d *portworx) UpgradeDriver(endpointURL string, endpointVersion string, ena
 		return fmt.Errorf("no endpoint supplied for upgrading driver")
 	}
 
-	if err := d.upgradePortworx(endpointURL, endpointVersion); err != nil {
-		return err
+	_, err := apiExtentions.GetCRD("storageclusters.core.libopenstorage.org", metav1.GetOptions{})
```
nikolaypopov, line 1685: Take a look at `sched-ops` we already implemented all the `StorarageCluster` operatons there, could use get from there https://github.com/portworx/sched-ops/blob/master/k8s/operator/storagecluster.go#L50

drivers/volume/portworx/portworx.go
```diff
@@ -2736,6 +2768,40 @@ func (d *portworx) getPxctlStatus(n node.Node) (string, error) {
 	return api.Status_STATUS_NONE.String(), nil
 }
 
+func (d *portworx) upgradePortworxUsingOperator(stc *operatorcorev1.StorageCluster, url string, endpointVersion string) error {
+	imageTag := endpointVersion
+	if strings.Contains(url, "edge") {
+		url = fmt.Sprintf("%s/%s/version", url, endpointVersion)
+		stdout, err := exec.Command("curl", "-L", url).CombinedOutput()
```
nikolaypopov, line 2775: We should use `"net/http"` here instead of curl

drivers/volume/portworx/portworx.go
```diff
@@ -2736,6 +2768,40 @@ func (d *portworx) getPxctlStatus(n node.Node) (string, error) {
...
+		var data interface{}
+		err = yaml.Unmarshal(stdout, &data)
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal px version. cause: %v", err)
+		}
+		versionMap := data.(map[string]interface{})
+		imageTag = versionMap["version"].(string)
+	}
+	stcImage := strings.Split(stc.Spec.Image, ":")[0]
+	stc.Spec.Image = fmt.Sprintf("%s:%s", stcImage, imageTag)
```
nikolaypopov, line 2788: Just changing `stc.Spec.Image` isn't going to be enough for upgrading, you also need to change `PX_RELEASE_MANIFEST_URL` if we are using `edge-install`

drivers/volume/portworx/portworx.go
```diff
@@ -2736,6 +2768,40 @@ func (d *portworx) getPxctlStatus(n node.Node) (string, error) {
...
+
+	_, err := pxOperator.UpdateStorageCluster(stc)
+	if err != nil {
+		return fmt.Errorf("failed to update storagecluster. cause: %v", err)
+	}
+
+	logrus.Infof("Storage cluster updated to version: %s", endpointVersion)
+
+	for _, node := range node.GetStorageDriverNodes() {
```
nikolaypopov, line 2797: We should get rid of this and just call `ValidateStorageCluster` instead, because it also does this as well as many other checks and validations https://github.com/libopenstorage/operator/blob/3ddc52ecdb29baed3b5cc31a0375cea17ae66030/pkg/util/test/util.go#L362

nikolaypopov (changes requested):
Let me know if you wanna sync up for the comments I left

PR #651, opened 2021-11-15
Change one of the autopilot rule to use 85% of usage instead of 90%

mborodin-px:
It fixes the issue where volume is completely full, but volume usage is less than 90%.

Tested here: 
http://jenkins.pwx.dev.purestorage.com/job/Autopilot/job/tp-aut-nextpx-resize-pvc/16/


nikolaypopov (approved):
LGTM

PR #652, opened 2021-11-16
clean dirty cmd output, reduce which cmd timeout

thiguetta:

This PR reduces the number of logs produced by `runCommandInPod`, instead of printing the entire pod data, it only prints namespace, name, node, and nodeIP. Also, this reduces the time when running `which pxctl` cmd to 30sec also reducing the log output when `which` doesn't find `pxctl` 

Closes #


mborodin-px (approved):
LGTM

PR #658, opened 2021-12-01
[PTX-4584] reduce the number of cat /proc/mounts in schedops.validat…

JimOuPure:
…eMountsInPods

schedops.validateMountsInPods tries to validate multiple mount paths, and for each path,
it calls “cat /proc/mounts" once, that’s unnecessary

tested on http://jenkins.pwx.dev.purestorage.com/view/Control%20Plane/job/Torpedo/job/tp-basic-sharedv4-svc/72/

### DIFF
--- a/drivers/volume/portworx/schedops/k8s-schedops.go
+++ b/drivers/volume/portworx/schedops/k8s-schedops.go
@@ -312,7 +312,6 @@ PodLoop:
 		skipHostMountCheck := false
 		for containerName, paths := range containerPaths {
 			output, err := k8sCore.RunCommandInPod([]string{"cat", "/proc/mounts"}, pod.Name, containerName, pod.Namespace)
-			mounts := strings.Split(output, "\n")
 			if err != nil && (err == k8serrors.ErrPodsNotFound || strings.Contains(err.Error(), "container not found")) {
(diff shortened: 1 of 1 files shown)


lsrinivas-pure (approved):
LGTM

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -311,19 +311,18 @@ PodLoop:
 		containerPaths := getContainerPVCMountMap(*pod)
 		skipHostMountCheck := false
 		for containerName, paths := range containerPaths {
+			output, err := k8sCore.RunCommandInPod([]string{"cat", "/proc/mounts"}, pod.Name, containerName, pod.Namespace)
+			mounts := strings.Split(output, "\n")
```
pureneelesh, line 315: nit: Why did you move this line before checking the err? Makes more sense to do this only if the command was successful.

pureneelesh (approved):

drivers/volume/portworx/schedops/k8s-schedops.go
JimOuPure: good point! didn't do it intentionally, let me change it

PR #660, opened 2021-12-03
Changed volume size for autopilot

mborodin-px:
Changed the volume size for autopilot application, to fix the issue where volume will be resized to round volume size. Currently we have an autopilot rule which resized volume by 50%. When volume size is 7Gb it will be  next following sequence: 7 -> 10.5 -> 15.75 -> 23.625, and k8s rounds these values and the final volume size becomes 24Gb.

Tested here: http://jenkins.pwx.dev.purestorage.com/job/Autopilot/job/tp-aut-op-nextpx-security-resize-pvc/


vilasd (approved):

thiguetta (approved):
LGTM

PR #661, opened 2021-12-04
PWX-22057: Stabilize Diag test, remove test we don't need anymore.

pault84:
Signed-off-by: pault84 <paul@portworx.com>

Remove obsolete code.

A-sync diags still needs some changes to porx because we can't actually check if the file exists as we don't know the filename.
We do not allow a filename to be passed in either so that's an oversight.

### DIFF
--- a/tests/telemetry/telemetry_test.go
+++ b/tests/telemetry/telemetry_test.go
@@ -39,12 +39,9 @@ var _ = Describe("{DiagsBasic}", func() {
 
 		ValidateApplications(contexts)
 
-		opts := make(map[string]bool)
-		opts[scheduler.OptionsWaitForResourceLeakCleanup] = true
-
 		// One node at a time, collect diags and verify in S3
 		for _, currNode := range node.GetWorkerNodes() {
-			Step(fmt.Sprintf("collect diags on node: %s", currNode.Name), func() {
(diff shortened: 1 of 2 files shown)


tests/telemetry/telemetry_test.go
```diff
@@ -96,48 +93,7 @@ var _ = Describe("{DiagsAsyncBasic}", func() {
 		}
 
 		for _, ctx := range contexts {
-			TearDownContext(ctx, opts)
-		}
-	})
-
-	JustAfterEach(func() {
```
thiguetta, line 103: why are we removing "JustAfterEach"?

tests/telemetry/telemetry_test.go
```diff
@@ -53,12 +50,15 @@ var _ = Describe("{DiagsBasic}", func() {
 					OnHost:        true,
 				}
 				err := Inst().V.CollectDiags(currNode, config, torpedovolume.DiagOps{Validate: true})
+				Step(fmt.Sprintf("error is: %s", err), func() {
```
thiguetta, line 53: do we really need a step to print the error here?

jrivera-px (approved):
LGTM...

tests/telemetry/telemetry_test.go
pault84: nope removing

thiguetta (approved):
LGTM

PR #663, opened 2021-12-06
[PWX-22141] consolidate MultiVolumeMountsForSharedV4Svc and MultiVolu…

JimOuPure:
…meMountsForSharedV4

tested on jenkins against both app vdbench-sv4-multivol and vdbench-sv4-svc-multivol

### DIFF
--- a/tests/sharedv4/sharedv4_test.go
+++ b/tests/sharedv4/sharedv4_test.go
@@ -34,7 +34,7 @@ var _ = BeforeSuite(func() {
 	InitInstance()
 })
 
-// This test performs multi volume mounts to a single  deployment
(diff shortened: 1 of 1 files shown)


lsrinivas-pure (approved):
LGTM

pureneelesh (approved):
This looks much cleaner. Thanks! I assume you will be modifying the default value for the FOCUS_TESTS param in the svc scale job?

JimOuPure:
> This looks much cleaner. Thanks! I assume you will be modifying the default value for the FOCUS_TESTS param in the svc scale job?

yes, will do, otherwise the test will fail

PR #665, opened 2021-12-07
Fix the fbVolumeExpectedSizechange error condition

dingyin:
fbVolumeExpectedSizechange error condition is wrong, because (byteUsedafter - byteUsedInitial) is passed to the function.

Closes #


david-enli (approved):
Looks good!

PR #666, opened 2021-12-09
PTX-4606 add ValidateStorageCluster() after cluster upgrade

thiguetta:
This PR adds ValidateStorageCluster() after cluster upgrade to completely validate storage required components images after upgrade is complete

Closes #


drivers/volume/portworx/portworx.go
```diff
@@ -2676,6 +2685,28 @@ func (d *portworx) ToggleCallHome(n node.Node, enabled bool) error {
 	return nil
 }
 
+func (d *portworx) ValidateStorageCluster() error {
+	pxOps, err := pxOperator.ListStorageClusters(schedops.PXNamespace)
```
nikolaypopov, line 2689: Instead of List we should just do Get, only 1 StorageCluster can be installed on k8s cluster.

drivers/volume/portworx/portworx.go
```diff
@@ -2676,6 +2685,28 @@ func (d *portworx) ToggleCallHome(n node.Node, enabled bool) error {
...
+	if err != nil {
+		return err
+	}
+	if len(pxOps.Items) > 0 {
+		pxVersion, err := d.getPxVersionOnNode(node.GetStorageDriverNodes()[0])
```
nikolaypopov, line 2694: I mean this can work, but I think we need to get PX version from StorageCluster itself, if its edge-install then we should get it from env var PX_RELEASE_MANIFEST_URL and then compare it to what is actually installed on nodes.

drivers/volume/portworx/portworx.go
```diff
@@ -2809,6 +2840,24 @@ func getSDKContainerPort() (int32, error) {
 	return 0, fmt.Errorf("px-sdk target port not found in service")
 }
 
+func getImageList(pxVersion, k8sVersion string) (map[string]string, error) {
+	imageList := make(map[string]string)
+	url := fmt.Sprintf("https://install.portworx.com/%s/version?kbver=%s", pxVersion, k8sVersion)
```
nikolaypopov, line 2845: You could have issues is pxVersion is empty and you going to have https://install.portworx.com//version, take a look at how I did this here https://github.com/portworx/spawn/blob/a7fdc207ca595947670d66606a1c19cddfa54984/cmd/provision_px_k8s.go#L2251, see if this would be better here

drivers/volume/portworx/portworx.go
```diff
@@ -2809,6 +2840,24 @@ func getSDKContainerPort() (int32, error) {
...
+	stdout, err := exec.Command("curl", "-L", url).CombinedOutput()
```
nikolaypopov, line 2846: We should use "net/http" here instead of curl, see here how we call URL and format images from it https://github.com/portworx/spawn/blob/a7fdc207ca595947670d66606a1c19cddfa54984/cmd/provision_px_k8s.go#L944

nikolaypopov (changes requested):

drivers/volume/portworx/portworx.go
thiguetta: in torpedo we don't know the name of the operator, that's why we need to use list instead of getting by name

thiguetta:
tested here http://jenkins.pwx.dev.purestorage.com/job/Torpedo/job/tp-ocp-upgrade-c7/12/console

drivers/volume/portworx/portworx.go
```diff
@@ -2890,6 +2931,37 @@ func getSDKContainerPort() (int32, error) {
...
+	if err := yaml.Unmarshal(body, &yamlMap); err != nil {
+		return imageList, err
+	}
+
+	imageList = make(map[string]string)
+	for key, value := range yamlMap["components"].(map[interface{}]interface{}) {
+		imageList[key.(string)] = value.(string)
+	}
+	imageList["version"] = fmt.Sprintf("portworx/oci-monitor:%s", yamlMap["version"].(string))
+	return imageList, nil
```
nikolaypopov, line 2962: I think it would be good to do a debug print of what the test actually gets from the versions URL, it will help us debug later incase of errors

drivers/volume/portworx/portworx.go
```diff
@@ -2757,6 +2769,35 @@ func (d *portworx) ToggleCallHome(n node.Node, enabled bool) error {
...
+	pxOps, err := pxOperator.ListStorageClusters(schedops.PXNamespace)
+	if err != nil {
+		return err
+	}
+	if len(pxOps.Items) > 0 {
+		k8sVersion, err := k8sCore.GetVersion()
+		if err != nil {
+			return err
+		}
+		imageList, err := getImageList(endpointURL, endpointVersion, fmt.Sprintf("%s.%s.0", k8sVersion.Major, k8sVersion.Minor))
```
nikolaypopov, line 2790: Idk if this matters or not, but why do we hardcode 0 by default, k8s versions can consist of 1.2.3 and it might make a difference, if you hardcode 0 at the end like so 1.2.0, we might potentially have issues with some k8s version dependent components, I would ask Piyush about this.

nikolaypopov (approved):
LGTM

PR #668, opened 2021-12-16
Fix FB torpedo failure, bytes change not expected

david-enli:

Closes #

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -547,14 +547,14 @@ func ValidateVolumeStatsticsDynamicUpdate(ctx *scheduler.Context, errChan ...*ch
 
 		byteUsedafter, err := Inst().V.ValidateGetByteUsedForVolume(vols[0].ID, make(map[string]string))
 		fmt.Println(fmt.Sprintf("after writing random bytes to the file the byteUsed is %v", byteUsedafter))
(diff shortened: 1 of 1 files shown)


tests/common.go
```diff
@@ -547,14 +547,14 @@ func ValidateVolumeStatsticsDynamicUpdate(ctx *scheduler.Context, errChan ...*ch
...
+		err = fbVolumeExpectedSizechange(byteUsedafter - byteUsedInitial)
 		expect(err).NotTo(haveOccurred())
 
 	})
 }
 
-func fbVolumeExpectedSizechange(sizeChangeInBytes int64) error {
-	if sizeChangeInBytes < -15*oneMegabytes || sizeChangeInBytes > 15*oneMegabytes {
+func fbVolumeExpectedSizechange(sizeChangeInBytes uint64) error {
+	if sizeChangeInBytes < (512 - 30) *oneMegabytes || sizeChangeInBytes > (512  + 30)*oneMegabytes {
```
pure-jliao, line 557: I assume this means we are writing 512 MB to the volume and then check if the actual size change is within 512+-30?

david-enli: yes, exactly,

pure-jliao (approved):
LGTM, please make sure the test could pass.

PR #676, opened 2022-01-07
PWX-22333: Enable filename validation for a-sync diags

pault84:
Signed-off-by: pault84 <paul@portworx.com>

Enables a-sync diags validation (Requires an OSD update)


ggriffiths (approved):
lgtm :+1:

PR #684, opened 2022-01-26
PWX-22505: implement sharedv4-svc failover test cases

pureneelesh:
Added test-sharedv4 app to Torpedo.

Added the following sharedv4 service test cases:

C54374: Verify failover by stopping the node where sharedv4 volume is attached (with i/o)
C54377: Verify previously written data after failover
C54382: Verify failback by stopping failover node

PWX-22505

### DIFF
--- a/drivers/scheduler/k8s/specs/test-sv4-svc/pod.yaml
+++ b/drivers/scheduler/k8s/specs/test-sv4-svc/pod.yaml
@@ -27,7 +27,7 @@ spec:
         image: portworx/sharedv4-test:torpedo
         imagePullPolicy: Always
         command: ["python", "/app/fileio.py"]
-        args: ["--lock", "--interval=0.1", "$(MY_FILE)"]
+        args: ["--lock", "--interval=0.25", "$(MY_FILE)"]
         env:
           - name: MY_POD_NAME
             valueFrom:
--- a/drivers/scheduler/k8s/specs/test-sv4-svc/pod.yaml
+++ b/drivers/scheduler/k8s/specs/test-sv4-svc/pod.yaml
@@ -1,0 +1,44 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-sv4-dep-svc
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      app: test-sv4-app-svc
+  template:
+    metadata:
+      labels:
+        app: test-sv4-app-svc
+    spec:
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - test-sv4-app-svc
+            topologyKey: "kubernetes.io/hostname"
+      containers:
+      - name: sv4test
+        image: portworx/sharedv4-test:torpedo
+        imagePullPolicy: Always
+        command: ["python", "/app/fileio.py"]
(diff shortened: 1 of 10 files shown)


drivers/scheduler/k8s/specs/test-sv4-svc/pod.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-sv4-dep-svc
+spec:
+  replicas: 3
```
dahuang-purestorage, line 6: should the number of replica be set as a parameter?
for example:
```
{{ if .Replicas }}
replicas: {{ .Replicas }}
{{ else }}
replicas: 3{{ end }}
```

pureneelesh: That seems like a good idea. I see that some app specs have that already. Do you know how the "replicas" is passed to the job?

lsrinivas-pure: it takes from custom config yaml file, path is passed using param
CUSTOM_APP_CONFIG

trierra (approved):
LGTM (checked the go/torpedo code only)

tests/basic/sharedv4_svc_test.go
```diff
@@ -171,11 +177,331 @@ var _ = Describe("{NFSServerFailover}", func() {
...
+								3*time.Duration(numPods)*time.Second)
+						})
+
+					Step(
+						fmt.Sprintf("failover #%d for app %s by stopping the volume driver on node %s",
+							i, ctx.App.Key, attachedNodeBefore.Name),
+						func() {
+							StopVolDriverAndWait([]node.Node{*attachedNodeBefore})
+						})
+
```
adityadani, line 294: Don't we need to wait here so that we give time for failover to happen?

tests/basic/sharedv4_svc_test.go
```diff
@@ -171,11 +177,331 @@ var _ = Describe("{NFSServerFailover}", func() {
...
+							Expect(attachedNodeAfter.Name).NotTo(Equal(attachedNodeBefore.Name))
+							countersAfter = getAppCounters(apiVol, attachedNodeAfter,
+								3*time.Duration(numPods)*time.Second)
+						})
+
+					Step(fmt.Sprintf("validate no I/O disruption for app %s after failover #%d", ctx.App.Key, i),
+						func() {
+							// 2 pods (1 on the old NFS server and 1 on the new NFS server) should get restarted,
+							// others should not be interrupted
+							validateAppCounters(ctx, countersBefore, countersAfter, numPods, 2)
```
adityadani, line 327: Should we do an explicit check here that only the pods which were running on the old and the new server got restarted?

adityadani (approved):
lgtm

tests/basic/sharedv4_svc_test.go
pureneelesh: I will add a short sleep here, just in case. But in practice, the PX takes longer to restart than the time taken for the "aggressive" failover. So, we the failover is assured even if you just to "systemctl restart portworx".

tests/basic/sharedv4_svc_test.go
pureneelesh: The existing test in this file verfies that already. See line 140 above:
```
// make sure the pods on both old and new server are restarted
```

pureneelesh:
Merged this as it is. Will make the couple of suggested changes in the next PR I have underway.

PR #685, opened 2022-01-26
Add stork upgrade trigger to longevity

Rohit-PX:

Longevity currently does not have an event that does stork upgrades. This PR fixes that.

Closes # https://portworx.atlassian.net/browse/PTX-4734

Added simple torpedo test to test this change, but will also monitor longevity once merged.

### DIFF
--- a/tests/longevity/longevity_test.go
+++ b/tests/longevity/longevity_test.go
@@ -70,6 +70,7 @@ var _ = Describe("{Longevity}", func() {
 		CloudSnapShot:    TriggerCloudSnapShot,
 		PoolResizeDisk:   TriggerPoolResizeDisk,
 		PoolAddDisk:      TriggerPoolAddDisk,
+		UpgradeStork:     TriggerUpgradeStork,
 	}
 	It("has to schedule app and introduce test triggers", func() {
 		Step(fmt.Sprintf("Start watch on K8S configMap [%s/%s]",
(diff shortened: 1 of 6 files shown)


Rohit-PX:
Running test here: https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/981/console

lsrinivas-pure (commented):
LGTM
Can you run this test such like  the test is triggered 2-3 times ans see what is happening

Rohit-PX:
> LGTM Can you run this test such like the test is triggered 2-3 times ans see what is happening

Fixed to run the test with loop count as a parameter and ran the test with parameter as 3.

tests/longevity/longevity_test.go
```diff
@@ -70,6 +70,7 @@ var _ = Describe("{Longevity}", func() {
 		CloudSnapShot:    TriggerCloudSnapShot,
 		PoolResizeDisk:   TriggerPoolResizeDisk,
 		PoolAddDisk:      TriggerPoolAddDisk,
+		UpgradeStork:     TriggerUpgradeStork,
```
lsrinivas-pure, line 73: can you add the triggerinterval map?

Rohit-PX: Done.

lsrinivas-pure (approved):
LGTM

PR #691, opened 2022-01-31
PWX-22505: sharedv4-svc failover tests by rebooting the NFS server node

pureneelesh:
Automated the following test case:

C54385: Verify nfs server reboot doesn't cause app restart (with exceptions)

Also,
- added retries when getting PX pod (debug pod) after rebooting the node
- changed the test app to append to the file and use local storage to verify
  correctness (part of the test-sharedv4 repo)
- updated the app deployment to incorporate the changes to the app

PWX-22505

### DIFF
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -466,31 +466,39 @@ func (s *SSH) doCmdUsingPodWithoutRetry(n node.Node, cmd string) (string, error)
 }
 
 func (s *SSH) doCmdUsingPod(n node.Node, options node.ConnectionOpts, cmd string, ignoreErr bool) (string, error) {
-	cmds := []string{"nsenter", "--mount=/hostproc/1/ns/mnt", "/bin/bash", "-c", cmd}
+	var debugPod *v1.Pod
 
-	allPodsForNode, err := k8sCore.GetPodsByNode(n.Name, execPodDefaultNamespace)
-	if err != nil {
-		logrus.Errorf("failed to get pods in node: %s err: %v", n.Name, err)
+	findDebugPodFunc := func() (interface{}, bool, error) {
+		allPodsForNode, err := k8sCore.GetPodsByNode(n.Name, execPodDefaultNamespace)
+		if err != nil {
+			logrus.Errorf("failed to get pods in node: %s err: %v", n.Name, err)
+			return nil, true, err
+		}
+		for _, pod := range allPodsForNode.Items {
+			if pod.Labels["name"] == execPodDaemonSetLabel && k8sCore.IsPodReady(pod) {
+				debugPod = &pod
+				break
+			}
+		}
+
+		if debugPod == nil {
+			return nil, true, &node.ErrFailedToRunCommand{
+				Node:  n,
+				Cause: fmt.Sprintf("debug pod not found in node %v", n),
+			}
+		}
+		return nil, false, nil
+	}
+
+	logrus.Debugf("Finding the debug pod to run command on node %s", n.Name)
+	if _, err := task.DoRetryWithTimeout(findDebugPodFunc, options.Timeout, options.TimeBeforeRetry); err != nil {
 		return "", err
 	}
-	var debugPod *v1.Pod
-	for _, pod := range allPodsForNode.Items {
-		if pod.Labels["name"] == execPodDaemonSetLabel && k8sCore.IsPodReady(pod) {
-			debugPod = &pod
-			break
-		}
-	}
 
-	if debugPod == nil {
-		return "", &node.ErrFailedToRunCommand{
-			Node:  n,
-			Cause: fmt.Sprintf("debug pod not found in node %v", n),
-		}
-	}
-
+	cmds := []string{"nsenter", "--mount=/hostproc/1/ns/mnt", "/bin/bash", "-c", cmd}
 	t := func() (interface{}, bool, error) {
 		output, err := k8sCore.RunCommandInPod(cmds, debugPod.Name, "", debugPod.Namespace)
-		if ignoreErr == false && err != nil {
+		if !ignoreErr && err != nil {
 			return nil, true, &node.ErrFailedToRunCommand{
 				Node: n,
 				Cause: fmt.Sprintf("failed to run command in pod. command: %v , err: %v, pod: %v",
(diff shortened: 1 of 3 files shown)


tests/basic/sharedv4_svc_test.go
```diff
@@ -259,84 +266,303 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
...
+			Step(fmt.Sprintf("induce multiple sharedv4 service failovers and verify I/O for app %s", ctx.App.Key), func() {
 				vols, err := Inst().S.GetVolumes(ctx)
 				Expect(err).NotTo(HaveOccurred())
 
 				apiVol, err := Inst().V.InspectVolume(vols[0].ID)
 				Expect(err).NotTo(HaveOccurred())
 
 				// verify failover and failback by repeating the steps below
-				for i := 0; i < 2; i++ {
+				for i := 0; i < 4; i++ {
```
ggriffiths, line 278: nit: could this be configurable if trying to repro a certain case w/ a number of failovers? If that is too much overhead, not a problem - four sounds good.

tests/basic/sharedv4_svc_test.go
```diff
@@ -259,84 +266,303 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
...
 					var attachedNodeBefore, attachedNodeAfter *node.Node
+					var failoverMethod string
 
-					// get the attached node before the failover
-					attachedNodeBefore, err = Inst().V.GetNodeForVolume(vols[0],
-						defaultCommandTimeout, defaultCommandRetry)
-					Expect(err).NotTo(HaveOccurred())
+					// The first 2 failovers are done by stopping PX on the NFS server node.
+					// The next 2 failovers are done by rebooting the node.
+					if i < 2 {
```
ggriffiths, line 285: this could be failoverMax/2 if made configurable

ggriffiths (approved):
looks good to me, just one nit/question

drivers/node/ssh/ssh.go
```diff
@@ -466,31 +466,39 @@ func (s *SSH) doCmdUsingPodWithoutRetry(n node.Node, cmd string) (string, error)
...
+		if debugPod == nil {
+			return nil, true, &node.ErrFailedToRunCommand{
+				Node:  n,
+				Cause: fmt.Sprintf("debug pod not found in node %v", n),
+			}
 		}
+		return nil, false, nil
+	}
+
+	logrus.Debugf("Finding the debug pod to run command on node %s", n.Name)
```
adityadani, line 493: why not run the findDebugPod and RunCommand in the same task?

tests/basic/sharedv4_svc_test.go
```diff
@@ -259,84 +266,303 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
...
+						failoverMethod = volDriverStop
```
adityadani, line 286: Is there a reason you want to combine this in a single test and not have two separate tests? It would be more deterministic if each test does one kind of failure scenario.

tests/basic/sharedv4_svc_test.go
```diff
@@ -259,84 +266,303 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
...
+					Step(fmt.Sprintf("get counters from node %v for app %s's volume before %s",
+						attachedNodeBefore.Name, ctx.App.Key, failoverLog),
 						func() {
 							countersBefore = getAppCounters(apiVol, attachedNodeBefore,
 								3*time.Duration(numPods)*time.Second)
 						})
 
-					Step(
-						fmt.Sprintf("failover #%d for app %s by stopping the volume driver on node %s",
+					if failoverMethod == volDriverStop {
```
adityadani, line 309: nit: this should volDriverRestart

tests/basic/sharedv4_svc_test.go
```diff
@@ -259,84 +266,303 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+
+				apiVol, err := Inst().V.InspectVolume(vols[0].ID)
+				Expect(err).NotTo(HaveOccurred())
+
+				attachedNode, err = Inst().V.GetNodeForVolume(vols[0], defaultCommandTimeout, defaultCommandRetry)
+				Expect(err).NotTo(HaveOccurred())
+				logrus.Infof("volume %v (%v) is attached to node %v", vols[0].ID, apiVol.Id, attachedNode.Name)
+
+				// start PX on the client nodes one at a time
```
adityadani, line 415: nit: restart PX

tests/basic/sharedv4_svc_test.go
```diff
@@ -259,84 +266,303 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
...
+				Expect(nodeWithNoPod).NotTo(BeNil(), "did not find NFS client node whose pod was removed")
+
+				// We don't expect a failover
+				validateAttachedNode(vols[0], attachedNode)
+				exportsAfterScaleDown := getExportsOnNode(apiVol, attachedNode)
+
+				Expect(exportsBeforeScaleDown).Should(ContainElement(nodeWithNoPod.DataIp),
+					"client IP not found in the exports before scaling down the app")
+
+				Expect(exportsAfterScaleDown).ShouldNot(ContainElement(nodeWithNoPod.DataIp),
```
adityadani, line 548: Lets check if the IPs for the other nodes exist where the pods are still running.

tests/basic/sharedv4_svc_test.go
```diff
@@ -209,7 +216,7 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
 	})
 
-	It("has to schedule apps and stop volume driver on nodes where volumes are attached", func() {
+	It("has to verify no I/O disruption for test-sv4-svc apps after PX restart or reboot on NFS server node", func() {
```
adityadani, line 219: nit: you could use the `scaleApp` helper on line 238

adityadani (approved):
lgtm

drivers/node/ssh/ssh.go
pureneelesh: Once the debug pod is ready after reboot, there is no need to find it again every time we retry the specified command. But I guess we can avoid that by checking if debugPod is not nil. Will change.

tests/basic/sharedv4_svc_test.go
pureneelesh: Main reason is to avoid duplicating the code. I experimented a bit using BeforeEach(), JustBeforeEach() etc But that did not seem to fit well here.

tests/basic/sharedv4_svc_test.go
pureneelesh: Good idea. Added.

tests/basic/sharedv4_svc_test.go
pureneelesh: Yes, that might be useful. Changed to retrieve the number of failovers from an env variable.

PR #698, opened 2022-02-02
PWX-22508: sharedv4-svc client teardown test while server is offline

adityadani:
- Automated the following test case:
  https://portworx.testrail.net/index.php?/cases/view/54780
  Verify a client can Unmount a volume even if the server is offline

Steps:
- Deploy the test sharedv4 svc app
- Scale up the application to the number of nodes
- Stop the node where volume is attached
- Scale down the application to 0 replicas
- Ensure pods (except the one running on down node) get Terminated
- Scale up the application
- Ensure pods go back to running state
Repeat the above steps for all configured specs.

Added a new helper function in torpedo - SelectiveWaitForTermination
- Ensures that pods get terminated on all the nodes except the ones provided
in the exclude node list provided as an argument.

Added specs for encrypted volumes with the test sharedv4 app

Signed-off-by: Aditya Dani <aditya@portworx.com>

Closes #


drivers/scheduler/k8s/k8s.go
```diff
@@ -2237,6 +2238,95 @@ func (k *K8s) WaitForDestroy(ctx *scheduler.Context, timeout time.Duration) erro
...
+			}
+		}
+	}
+
+	// Compare the two pod maps. Create a list of pod names which are not running
+	// on the excluded nodes
+	var podNames []string
+	for _, pod := range allPods {
+		if _, exists := excludedNodePods[pod.UID]; !exists {
+			podNames = append(podNames, pod.Name)
```
pureneelesh, line 2324: nit: Probably fine in practice but it seems odd that here we are returning the pod names without namespace presumably under the assumption that all pods are in the same namespace. But the code above tries to handle multiple namespaces.

drivers/scheduler/k8s/specs/test-sv4-svc-enc/pod.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-sv4-dep-svc-enc
```
pureneelesh, line 4: Thanks for adding this. It was on my todo list.

tests/basic/sharedv4_svc_test.go
```diff
@@ -345,6 +345,136 @@ var _ = Describe("{Shared4SvcFailoverIO}", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+
+				Step(
+					fmt.Sprintf("stopping volume driver on node %s", attachedNode.Name),
+					func() {
+						StopVolDriverAndWait([]node.Node{*attachedNode})
+					},
+				)
+
+				Step(fmt.Sprintf("scale down app %s to 0", ctx.App.Key), func() {
```
pureneelesh, line 408: Could the failover have happened at this point?  This test is good, but wondering if there should also be a test that scales down app while there was no failover yet (perhaps by switching to "normal" failover strategy).

pureneelesh (approved):

tests/basic/sharedv4_svc_test.go
adityadani: Yes I was also thinking of that. While testing I did see a few occurrences where failover happened even before the scale down. Doing a failover and then scale down would probably get executed as a part of other tests too. So let me change the storage class / update the volume to use normal strategy.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2237,6 +2238,95 @@ func (k *K8s) WaitForDestroy(ctx *scheduler.Context, timeout time.Duration) erro
...
+func filterPodsByNodes(ctx *scheduler.Context, excludeList []node.Node) ([]string, error) {
+	allPods := make(map[types.UID]corev1.Pod)
+	excludedNodePods := make(map[types.UID]corev1.Pod)
+	namespaces := make(map[string]string)
+	for _, specObj := range ctx.App.SpecList {
+		var pods []corev1.Pod
+		var err error
+		if obj, ok := specObj.(*appsapi.Deployment); ok {
+			if pods, err = k8sApps.GetDeploymentPods(obj); err != nil {
+				return nil, &scheduler.ErrFailedToValidateAppDestroy{
```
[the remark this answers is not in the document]
adityadani: ah! I copied it from another function and just changed the cause. Let me fix it.

drivers/scheduler/k8s/k8s.go
adityadani: I was also thinking of adding node names as well to indicate where the pods are running. Will do.

tests/basic/sharedv4_svc_test.go
adityadani: Added steps to change the failover strategy to normal

PR #699, opened 2022-02-02
PWX-22447: added sharedv4-svc test for PX offline too long on the client node

pureneelesh:
C54778: Verify an offline client (using NFS mountpoint) after restart can access
the NFS mountpoint even if the server has moved to another node.

PWX-22447


tests/basic/sharedv4_svc_test.go
```diff
@@ -749,6 +695,157 @@ var _ = Describe("{Shared4SvcUnexportExport}", func() {
...
+			contexts = append(contexts, ScheduleApplications(fmt.Sprintf("client-offline-%d", i))...)
+		}
+		testSv4Contexts = getTestSv4Contexts(contexts)
+		if len(testSv4Contexts) == 0 {
+			Skip("No test-sv4-svc apps were found")
+		}
+		workers = node.GetWorkerNodes()
+
+		Step("scale the test-sv4-svc apps so that one pod runs on each worker node", func() {
+			scaleApps(testSv4Contexts, len(workers))
```
dahuang-purestorage, line 723: quick question, this is assuming k8s scales evenly to the worker nodes. There won't be a case where one node runs two pod. Is this assumption correct?
If not, is there a way we can ensure each node runs a pod?

tests/basic/sharedv4_svc_test.go
```diff
@@ -749,6 +695,157 @@ var _ = Describe("{Shared4SvcUnexportExport}", func() {
...
+				failover = len(workers) >= 5
+
+				apiVol, err = Inst().V.InspectVolume(vols[0].ID)
+				Expect(err).NotTo(HaveOccurred())
+
+				attachedNode, err = Inst().V.GetNodeForVolume(vols[0], defaultCommandTimeout, defaultCommandRetry)
+				Expect(err).NotTo(HaveOccurred())
+				logrus.Infof("volume %v (%v) is attached to node %v", vols[0].ID, apiVol.Id, attachedNode.Name)
+
+				// Choose a client node to stop PX on.
```
dahuang-purestorage, line 759: is there a criteria for choosing the client node to stop? 
From reading the code. Seems like there are two types of nodes we don't stop on. When:
1. the node is a replica node.
2. the node is a metadata node (and when failover is true)

I'm just hoping there's a one-line description on what client node we are looking for. For example, "client node that's not a replica node and is not metadata node".

tests/basic/sharedv4_svc_test.go
```diff
@@ -749,6 +695,157 @@ var _ = Describe("{Shared4SvcUnexportExport}", func() {
...
+					restartVolumeDriverOnNode(attachedNode)
+					attachedNodeAfter, err := Inst().V.GetNodeForVolume(vols[0],
+						defaultCommandTimeout, defaultCommandRetry)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(attachedNodeAfter.Name).NotTo(Equal(attachedNode.Name))
+					logrus.Infof("volume %v (%v) is attached to node %v after failover",
+						vols[0].ID, apiVol.Id, attachedNodeAfter.Name)
+					attachedNode = attachedNodeAfter
+				} else {
+					logrus.Infof("skipping the failover since there are not enough nodes")
```
dahuang-purestorage, line 807: should we move this check to an earlier step?

tests/basic/sharedv4_svc_test.go
adityadani: The Deployment spec of the application used in these tests have a podAntiAffinity which ensures that only one pod runs on a node.

tests/basic/sharedv4_svc_test.go
```diff
@@ -749,6 +695,157 @@ var _ = Describe("{Shared4SvcUnexportExport}", func() {
...
+				Expect(exports).Should(ContainElement(clientNode.DataIp),
+					"client IP not found in the exports before stopping PX on the client node")
+
+				logrus.Infof("stopping volume driver on node %s", clientNode.Name)
+				StopVolDriverAndWait([]node.Node{*clientNode})
+
+				// TODO: offlineClientTimeout = 15 * time.Minute in ref.go
+				// need to make it configurable so that we don't have to sleep for that long in the test
+				logrus.Infof("sleep to allow the server to remove client %v's export", clientNode.Name)
+				time.Sleep(16 * time.Minute)
```
adityadani, line 785: nit: add this as a constant

adityadani (approved):
lgtm

tests/basic/sharedv4_svc_test.go
```diff
@@ -749,6 +695,157 @@ var _ = Describe("{Shared4SvcUnexportExport}", func() {
...
+
+				// Verify server removed the export
+				exports = getExportsOnNode(apiVol, attachedNode)
+				Expect(exports).ShouldNot(ContainElement(clientNode.DataIp),
+					"client IP still present in the exports after stopping PX on client node for a long time")
+			})
+
+			Step(fmt.Sprintf("do a failover if possible for app %s", ctx.App.Key), func() {
+				if failover {
+					restartVolumeDriverOnNode(attachedNode)
```
adityadani, line 798: Can we do the before and after counter checks? Is that the refactoring you are planning to do?

tests/basic/sharedv4_svc_test.go
```diff
@@ -749,6 +695,157 @@ var _ = Describe("{Shared4SvcUnexportExport}", func() {
...
+				// TODO: offlineClientTimeout = 15 * time.Minute in ref.go
+				// need to make it configurable so that we don't have to sleep for that long in the test
+				logrus.Infof("sleep to allow the server to remove client %v's export", clientNode.Name)
+				time.Sleep(16 * time.Minute)
+
+				// We don't expect a failover
+				validateAttachedNode(vols[0], attachedNode)
+
+				// Verify server removed the export
+				exports = getExportsOnNode(apiVol, attachedNode)
```
adityadani, line 791: should we also check the annotations on the k8s service?

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: ah got it. thanks! :)

dahuang-purestorage (approved):
lgtm. just minor comments

tests/basic/sharedv4_svc_test.go
pureneelesh: added more descriptive comment

tests/basic/sharedv4_svc_test.go
pureneelesh: Not sure what you mean. We are not skipping the whole test here. We skip the failover part which is done in this step.

tests/basic/sharedv4_svc_test.go
pureneelesh: We verify that all the app pods are incrementing their counters after brining up all the nodes. Do you think additional checks are needed?

pureneelesh:
Merging this now. Comments addressed in the follow-up PR.

PR #704, opened 2022-02-04
Adding IPV6 support for torpedo

aghodke1312:

Closes #


drivers/volume/portworx/portworx.go
```diff
@@ -1854,6 +1854,9 @@ func (d *portworx) testAndSetEndpointUsingNodeIP(ip string) error {
 }
 
 func (d *portworx) testAndSetEndpoint(endpoint string, sdkport, apiport int32) error {
+	if strings.Contains(endpoint, ":"){
```
jrivera-px, line 1857: Even though at this point you are sure that the address is just an IP you may want to add a check for '['  just in case someone sends in an IP with brackets around it already.  I have seen this in parsed out IP addresses.  Although if the IP is obtained using net.SplitHostPort() then you should be good 👍.  
if strings.Contains(endpoint, ":") && !strings.Contains(endpoint, "[") {
		host = fmt.Sprintf("[%s]", endpoint)
	}

jrivera-px (approved):
Looks good,  nice job!

thiguetta:
do we have a jenkins job where we can validate this change? if so please post the link

PR #705, opened 2022-02-04
PWX-22802: Refactored to remove Pure-specific volume driver functions

Pure-AdamuKaapan:
Partly a practice PR just for getting used to torpedo code and layout, and cleaning up some code duplication I noticed that seemed relatively easy to fix ("good first issue").

Currently running this on http://jenkins.pwx.dev.purestorage.com/blue/organizations/jenkins/Pure-FAFB%2Fdev-test-fa/detail/dev-test-fa


drivers/scheduler/k8s/k8s.go
```diff
@@ -1099,33 +1119,46 @@ func (k *K8s) createNamespace(app *spec.AppSpec, namespace string, options sched
...
-		}
-		if _, found := scParameters["io_profile"]; found {
-			delete(scParameters, "io_profile")
-		}
-		if _, found := scParameters["priority_io"]; found {
-			delete(scParameters, "priority_io")
-		}
+		return convertStorageClassToPureFB(scParameters)
+	}
+	// TODO: what parameters define Pure FA passthrough volumes?
```
Pure-AdamuKaapan, line 1124: TODO, @david-enli or someone else you could you advise me here? (See the above if check for FB, I'm looking for what might be a comparable check for FA).

drivers/scheduler/k8s/k8s.go
```diff
@@ -2709,6 +2742,13 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 			if err != nil {
 				return nil, err
 			}
+			related, err := k.relatedVolume(pvcObj)
+			if err != nil {
+				return nil, err
+			}
+			if !related {
+				continue
+			}
```
piyush-nimbalkar, line 2751: Not sure if all volumes will have a backend. This will get called for non-pure volumes as well. In that case it would not return that volume.

drivers/scheduler/k8s/k8s.go
```diff
@@ -591,6 +593,24 @@ func validateSpec(in interface{}) (interface{}, error) {
 	return nil, fmt.Errorf("unsupported object: %v", reflect.TypeOf(in))
 }
 
+// relatedVolume will test if the given volume should be included. Currently that just relates to
+// if we're filtering by Pure volumes, but can be expanded later.
+func (k *K8s) relatedVolume(claim *v1.PersistentVolumeClaim) (bool, error) {
```
piyush-nimbalkar, line 598: What does 'related' mean? It does not convey the intention clearly.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2847,89 +2818,51 @@ func (k *K8s) ResizeVolume(ctx *scheduler.Context, configMapName string) ([]*vol
...
-			if err != nil || pvcList == nil {
-				return nil, &scheduler.ErrFailedToResizeStorage{
-					App:   ctx.App,
-					Cause: fmt.Sprintf("Failed to get PVC from StatefulSet: %v. Err: %v", ss.Name, err),
-				}
-			}
-
-			for _, pvc := range pvcList.Items {
-				vol, err := k.resizePVCBy1GB(ctx, &pvc)
+			if related {
```
piyush-nimbalkar, line 2825: Do we need to check this now? Can we directly call `resizePVCBy1GB`?

drivers/scheduler/k8s/k8s.go
```diff
@@ -2847,89 +2818,51 @@ func (k *K8s) ResizeVolume(ctx *scheduler.Context, configMapName string) ([]*vol
...
 			for _, pvc := range pvcList.Items {
-				scForPvc, err := k8sCore.GetStorageClassForPVC(&pvc)
+				related, err := k.relatedVolume(&pvc)
 				if err != nil {
-					return err
+					return nil, err
 				}
-				if scForPvc.Parameters["backend"] == PureFile {
-					_, err := k.resizePVCBy1GB(ctx, &pvc)
+				if related {
```
piyush-nimbalkar, line 2854: Same as above.

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: Agreed, I was struggling to think of a good term at the time. Will revise

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: I think we do still, unless you're saying this also gets checked somewhere above us? I'm still learning my way around the code

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: For cases of Torpedo running on non-Pure backends, the "relatedVolume" check will do a fast exit. If we have Pure backends enabled, I _think_ we want to always only include Pure volumes, at least in terms of the current functionality I'm refactoring.

drivers/scheduler/k8s/k8s.go
david-enli: I need to check with Nikolday, @nikolaypopov do you remember why we specifically look for shared / sharedv4 parameter here?

drivers/scheduler/k8s/k8s.go
```diff
@@ -591,6 +593,25 @@ func validateSpec(in interface{}) (interface{}, error) {
 	return nil, fmt.Errorf("unsupported object: %v", reflect.TypeOf(in))
 }
 
+// filterPureVolumesIfEnabled will return whether or not the given PVC is a Pure
+// direct access volume, only if PureVolumes is true. If PureVolumes is false, all
+// volumes are valid.
+func (k *K8s) filterPureVolumesIfEnabled(claim *v1.PersistentVolumeClaim) (bool, error) {
```
david-enli, line 599: yea this is a lot better way to handle it, I'd just make sure the resize volumes till work before merging it. But i dont see any problems assuming we converting the right storageclass. also, since we are here, is it better to have a filter for pure_file volumes and pure_block volumes ?

Pure-AdamuKaapan: I think having the same one is fine, we can do more specific checks later.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2709,6 +2743,13 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 			if err != nil {
 				return nil, err
 			}
+			related, err := k.filterPureVolumesIfEnabled(pvcObj)
```
piyush-nimbalkar, line 2746: nit: rename `related`

drivers/scheduler/k8s/k8s.go
piyush-nimbalkar: I see. So you want to resize the pure volume only if it is either pure_file/pure_block.

piyush-nimbalkar (approved):

drivers/scheduler/k8s/k8s.go
piyush-nimbalkar: Got it. I did not realize that we are using`k.PureVolumes` to just run pure volume tests.

PR #708, opened 2022-02-08
PWX-22840: refactor sharedv4 svc tests

pureneelesh:
- common setup and teardown instead of duplicating it for each test

- separate out failover/failback tests by reboot and PX-restart without
  duplicating the test flow using "shared behaviors" pattern

- added env variable to skip collecting diags (useful when running the tests locally)

- other minor cleanup pointed out by static checks

- addresssed comments from the previous review

PWX-22840

The tests structured differently now but there are no changes to the test logic.

### DIFF
--- a/tests/basic/sharedv4_svc_test.go
+++ b/tests/basic/sharedv4_svc_test.go
@@ -240,7 +240,7 @@ var _ = Describe("{Shared4 service apps}", func() {
 	var contexts, testSv4Contexts []*scheduler.Context
 	var workers []node.Node
 	var numPods int
-	var namespace_prefix string
+	var namespacePrefix string
 
 	JustBeforeEach(func() {
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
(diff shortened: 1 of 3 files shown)


tests/basic/sharedv4_svc_test.go
```diff
@@ -199,648 +196,603 @@ type appCounter struct {
...
+	var namespacePrefix string
 
 	JustBeforeEach(func() {
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
-	})
 
-	It("has to verify no I/O disruption for test-sv4-svc apps after PX restart or reboot on NFS server node", func() {
-		contexts = make([]*scheduler.Context, 0)
+		// Set up all apps
+		contexts = nil
```
dahuang-purestorage, line 249: Is setting contexts to nil necessary here? I don't see it being populated before this test step. Since it's nil by default when it's defined, i don't see a need to set it to nil again.

dahuang-purestorage (approved):

dahuang-purestorage:
lgtm. nice abstraction of shared method.

PR #710, opened 2022-02-09
PWX-22802: Pure FA direct access test preparation

Pure-AdamuKaapan:
The main feature of this PR is a new volume driver `pure` which composes all the functionality of the original `portworx` driver, but has a different name. This will allow us to provide Pure-specific storage class specs, so we can test specific cases like filesystem options, raw block, etc.

PWX-22802


drivers/scheduler/k8s/specs/elasticsearch/pure/pure-storage-class.yaml
```diff
@@ -0,0 +1,15 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+    name: elasticsearch-sc
+provisioner: kubernetes.io/portworx-volume
```
piyush-nimbalkar, line 5: Are we planning to run this using the portworx in-tree driver or the CSI driver?

drivers/volume/portworx/portworx.go
```diff
@@ -3150,4 +3155,5 @@ func getImageList(endpointURL, pxVersion, k8sVersion string) (map[string]string,
 
 func init() {
 	torpedovolume.Register(DriverName, provisioners, &portworx{})
+	torpedovolume.Register(PureDriverName, provisioners, &pure{portworx: portworx{}})
```
piyush-nimbalkar, line 3158: Same here. I am not sure if we support volume provisioning with portworx in-tree driver. If not, we will have to only pass the csi provisioner here.

drivers/volume/portworx/pure.go
```diff
@@ -0,0 +1,20 @@
...
+	PureDriverName = "pure"
+)
+
+// pure is essentially the same as the portworx volume driver, just different in name. This way,
+// we can have separate specs for pure volumes vs. normal portworx ones
+type pure struct {
+	portworx
+}
+
+func (p *pure) Init(sched, nodeDriver, token, storageProvisioner, csiGenericDriverConfigMap string) error {
```
piyush-nimbalkar, line 14: This is a very nice way to implement this! 👏

drivers/volume/portworx/pure.go
```diff
@@ -0,0 +1,20 @@
...
+	return p.portworx.init(sched, nodeDriver, token, storageProvisioner, csiGenericDriverConfigMap, PureDriverName)
+}
+
+func (p *pure) String() string {
```
piyush-nimbalkar, line 18: Let's also move the `ValidatePureVolumesNoReplicaSets` method from the portworx driver to this driver if possible. 
I would also recommend renaming it to `ValidateVolumeHasNoReplicaSets` and then let the portworx driver implement the DefaultDriver to use the NotImplemented version like other volume drivers. 

Does not have to be part of this PR.

drivers/volume/portworx/pure.go
Pure-AdamuKaapan: Thank @nikolaypopov , he had the original idea :grin: I agree, made it muuuuch easier than I thought

drivers/scheduler/k8s/specs/elasticsearch/pure/pure-storage-class.yaml
Pure-AdamuKaapan: I think you're right, checking with @david-enli to be sure

Pure-AdamuKaapan: Confirmed, will go ahead and fix this

piyush-nimbalkar (approved):

nikolaypopov (approved):

drivers/volume/portworx/schedops/k8s-schedops.go
```diff
@@ -67,6 +67,11 @@ const (
 	talismanServiceAccount      = "talisman-account"
 	talismanImage               = "portworx/talisman:latest"
 	rancherControlPlaneLabelKey = "node-role.kubernetes.io/controlplane"
+
+	// pureVolumeOUID is used to identify if a mapper device is of Pure origin
+	// (if it leads with /dev/mapper/3624a9370, it's a Pure volume, if it's some other prefix it's not)
+	pureVolumeOUID = "3624a9370"
```
piyush-nimbalkar, line 73: Is this always true or specific to an array? What is special about this number?

piyush-nimbalkar (approved):

drivers/volume/portworx/schedops/k8s-schedops.go
Pure-AdamuKaapan: This is always true for Pure devices: it's our OUID, our company-specific identifier. It's also what we use in Portworx for FA cloud drives.

PR #713, opened 2022-02-11
PWX-22508: add sharedv4 svc resync test

pureneelesh:
Added test:
- C54379: Verify failover by stopping the node in resync

Also,

- added "SKIP_PX_SERVICE_ENDPOINT" env variable to speed up running the tests locally
  using ginko where the svc endpoint is not accessible

- fixed nfs server failover test to work with test-shaerdv4 app by setting
  the HA level to 2 when needed

- refactored some more common code into a utility function that returns
  vol, apivol and attached node for test-sharedv4 app

- fixed the UnexportExport test to wait long enough for the pods to be terminated

PWX-22508

### DIFF
--- a/tests/basic/sharedv4_svc_test.go
+++ b/tests/basic/sharedv4_svc_test.go
@@ -913,7 +913,7 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
 				})
 
 				Step(fmt.Sprintf("verify that the volume %v (%v) is still in resync", vol.ID, apiVol.Id), func() {
-					cmd := fmt.Sprintf("pxctl volume inspect %s | grep 'Replication Status'", apiVol.Id)
+					cmd := fmt.Sprintf("pxctl volume inspect %s | grep \"Replication Status\"", apiVol.Id)
 					replStatus, err := runCmd(cmd, *resyncReplica)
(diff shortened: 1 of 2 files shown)


tests/basic/sharedv4_svc_test.go
```diff
@@ -802,16 +815,198 @@ var _ = Describe("{Shared4 service apps}", func() {
 		})
 	})
 
+	// Induce a failover while the NFS server node is in resync.
+	Context("{Sharedv4SvcResync}", func() {
+		// There are 3 replica nodes for the test-sharedv4 app's volume. We designate them as:
+		//
+		// - attachedNodeOrig: node where the volume was attached initially when the app was deployed
+		//
+		// - resyncReplica: node that will be resync'ing the volume while the volume fails over to
```
dahuang-purestorage, line 824: nit: resync'ing typo?

dahuang-purestorage (approved):
lgtm

tests/basic/sharedv4_svc_test.go
pureneelesh: Apostrophe is sometimes used to indicate missing letters. In this case, it is short for resynchronizing. But "resyncing" is considered a valid word, so may be not needed. I will leave it like that so I don't have to with for the tests before merging.

PR #714, opened 2022-02-15
PWX-22929: fix the sharedv4 svc pod restart test

pureneelesh:
The test was working fine when there was only 1 app specified in the jenkins job.
It started failing intermittently after addition of 2 more apps.

- scale all the apps to 2 initially, to ensure no pending pods when test cordons all but 2 nodes

- after scaling the apps down to 0 wait until all the pods are gone

PWX-22929

### DIFF
--- a/tests/basic/sharedv4_svc_test.go
+++ b/tests/basic/sharedv4_svc_test.go
@@ -45,7 +45,27 @@ var _ = Describe("{Sharedv4SvcPodRestart}", func() {
 			contexts = append(contexts, ScheduleApplications(fmt.Sprintf("nfsserverfailover-%d", i))...)
 		}
 
+		// scale all the apps to have 2 pods since later we will have only 2 nodes where the pods can run
+		for _, ctx := range contexts {
+			Step(fmt.Sprintf("scale app %s to 2", ctx.App.Key), func() {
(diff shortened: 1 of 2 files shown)


tests/basic/sharedv4_svc_test.go
```diff
@@ -45,8 +45,28 @@ var _ = Describe("{Sharedv4SvcPodRestart}", func() {
...
+		for _, ctx := range contexts {
+			Step(fmt.Sprintf("scale app %s to 2", ctx.App.Key), func() {
+				applicationScaleUpMap, err := Inst().S.GetScaleFactorMap(ctx)
+				Expect(err).NotTo(HaveOccurred())
+				for name := range applicationScaleUpMap {
+					applicationScaleUpMap[name] = int32(2)
+				}
+				err = Inst().S.ScaleApplication(ctx, applicationScaleUpMap)
+				Expect(err).NotTo(HaveOccurred())
+			})
```
dahuang-purestorage, line 58: I think we already have scaleApp() function that does the same functionality. Can we replace this with scaleApp() function to reduce duplication?

pureneelesh: good point. Fixed.

PR #715, opened 2022-02-16
PWX 22789 failover test for maintenance mode

dahuang-purestorage:
This PR adds failover test for maintenance mode.
- We added a 30 sec sleep in the recover mode for the failover mode to operate. We should refactor this out later and put the sleep in the test instead of the driver function. 

Closes # PWX-22798

Testing note:
run with local torpedo with PX version 2.10.0-4c8fb75
```ginkgo --focus Shared4SvcRecoverNode  -v bin/basic.test -- -spec-dir `pwd`/drivers/scheduler/k8s/specs --app-list test-sv4-svc```

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -620,6 +620,18 @@ func (d *portworx) GetStorageDevices(n node.Node) ([]string, error) {
 
 func (d *portworx) RecoverDriver(n node.Node) error {
 
+	if err := d.EnterMaintenance(n); err != nil {
+		return err
+	}
+
+	if err := d.ExitMaintenance(n); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (d *portworx) EnterMaintenance(n node.Node) error {
 	t := func() (interface{}, bool, error) {
 		if err := d.maintenanceOp(n, enterMaintenancePath); err != nil {
 			return nil, true, err
@@ -647,11 +659,11 @@ func (d *portworx) RecoverDriver(n node.Node) error {
 			Cause: err.Error(),
 		}
 	}
-	// TODO: refactor the function into two pieces, entermaintenance and exitmaintenance. So we don't need to have
-	// random sleep interval in the middle.
-	time.Sleep(30 * time.Second)
+	return nil
+}
 
-	t = func() (interface{}, bool, error) {
+func (d *portworx) ExitMaintenance(n node.Node) error {
+	t := func() (interface{}, bool, error) {
 		if err := d.maintenanceOp(n, exitMaintenancePath); err != nil {
 			return nil, true, err
 		}
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -124,7 +124,7 @@ var provisioners = map[torpedovolume.StorageProvisionerType]torpedovolume.Storag
 }
 
 var csiProvisionerOnly = map[torpedovolume.StorageProvisionerType]torpedovolume.StorageProvisionerType{
-	PortworxCsi:     "pxd.portworx.com",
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: drivers/volume/portworx/portworx.go [hunk 1]
```diff
@@ -647,6 +647,10 @@ func (d *portworx) RecoverDriver(n node.Node) error {
 			Cause: err.Error(),
 		}
 	}
+	// TODO: refactor the function into two pieces, entermaintenance and exitmaintenance. So we don't need to have
```
pureneelesh, line 650: I like the idea of creating 2 new functions and calling them from here. Then, we can call the new functions directly from our tests. If it is not too much work, we should just refactor now. Otherwise, maybe we could add a parameter for the sleep interval?

tests/basic/sharedv4_svc_test.go
```diff
@@ -272,6 +272,27 @@ func (fm *failoverMethodReboot) sleepBetweenFailovers() time.Duration {
 	return 2 * time.Minute
 }
 
+type failoverMethodRecover struct {
```
pureneelesh, line 275: The term "Recover" does not make sense in this context to me at least. That term was used in the context of recovering the node which is in StorageDown state. We are not doing that here. We should probably call  it failoverMethodMaintenance instead.

tests/basic/sharedv4_svc_test.go
```diff
@@ -272,6 +272,27 @@ func (fm *failoverMethodReboot) sleepBetweenFailovers() time.Duration {
...
+
+func (fm *failoverMethodRecover) String() string {
+	return "recover"
+}
+
+func (fm *failoverMethodRecover) getExpectedPodDeletions() []int {
+	// 1 or 2 pods may get deleted depending on what kubelet does on the rebooted node.
+	// The kubelet could set up the mount for the same pod or it could create a new pod.
+	return []int{1, 2}
+}
```
pureneelesh, line 290: This is not applicable here. We should expect 2 pods to get deleted.

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: updated

tests/basic/sharedv4_svc_test.go
```diff
@@ -272,6 +272,27 @@ func (fm *failoverMethodReboot) sleepBetweenFailovers() time.Duration {
...
+	maintainVolumeDriverOnNode(attachedNode)
+}
+
+func (fm *failoverMethodMaintenance) String() string {
+	return "maintenance"
+}
+
+func (fm *failoverMethodMaintenance) getExpectedPodDeletions() []int {
+	// 1 or 2 pods may get deleted depending on what kubelet does on the rebooted node.
+	// The kubelet could set up the mount for the same pod or it could create a new pod.
```
pureneelesh, line 288: remove the comment?

tests/basic/sharedv4_svc_test.go
```diff
@@ -420,6 +441,16 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
 			})
 			testFailoverFailback()
 		})
+
+		// test failover/failback by recover the node
+		Context("{Shared4SvcRecoverNode}", func() {
+			BeforeEach(func() {
+				namespacePrefix = "recovernode-"
```
pureneelesh: recovernode -> nodemaintenance 

Also, we don't need - in the end any more.

pureneelesh (approved):

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: updated

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: updated

PR #718, opened 2022-02-17
PWX-22804: Converted FB direct access tests to generic Pure

Pure-AdamuKaapan:
Enables the FB Direct Access tests to also work with FA.

General summary of what I had to change:
* Removed the "storage class conversion" methods, as the specs will now just be created specifically for Pure
* Overrides the "ValidateCreateSnapshot" method for the Pure volume driver, as it cannot have hyphens in the snapshot name
* Fleshed out some log statements
* Renamed some errors/variables to be non-FB-specific
* Commented out the volume statistics checks for now as FlashArray hasn't implemented it yet (if I have to merge this before the FA feature has merged, I'll find a way to make it conditional for FB only)
* The snapshot test now expects no error for FA snapshots, expects an error for FB snapshots
* Commented out the "groupsnap" test as it's currently broken: it prints the correct error message but returns exit code 0

PWX-22804


david-enli:
LGTM, the actual snap/restore will be in a different PR?

Pure-AdamuKaapan:
Yeah, snapshot restore will be a different test. It's currently marked as **not automated** in testrail because of this, but at least we have automated coverage of part of it.

david-enli (approved):
LGTM

drivers/volume/portworx/pure.go
```diff
@@ -18,3 +25,21 @@ func (p *pure) Init(sched, nodeDriver, token, storageProvisioner, csiGenericDriv
...
+	var token string
+	token = p.getTokenForVolume(volumeName, params)
+	if val, hasKey := params[refreshEndpointParam]; hasKey {
+		refreshEndpoint, _ := strconv.ParseBool(val)
+		p.refreshEndpoint = refreshEndpoint
+	}
+
+	volDriver := p.getVolDriver()
+	// This is the only difference: we have to name snapshots with hyphens, not underscores
+	_, err := volDriver.SnapshotCreate(p.getContextWithToken(context.Background(), token), &api.SdkVolumeSnapshotCreateRequest{VolumeId: volumeName, Name: volumeName + "-snapshot"})
```
piyush-nimbalkar, line 39: Why do we need to create the snapshot when validating a snapshot?

Pure-AdamuKaapan: I was kinda wondering that originally myself too. I think it's "Validating creating a snapshot". e.g. validating that the creation is successful?

piyush-nimbalkar: @david-enli, do you have the context on this?

Pure-AdamuKaapan: Discussed offline in Slack, this has been resolved. Added comments for notes to refactor later

PR #719, opened 2022-02-18
PWX-22789 Remove using service Endpoints for maintainOp

dahuang-purestorage:
In Enter/Exit Maintenance Mode, we make two calls: `maintenanceOp` for
putting the node in maintenance mode, and `getPxNode` for checking the
status of the node.

The issue with `maintenanceOp` with using serviceEndpoint was that it
calls to the services and return a random node endpoint the service choose.
Therefore, putting a random node in maintenance and failing the status check.

This PR aims to resolve the issue. For now, we remove getting the service
endpoint and use the node's address. If there's any concern or better way to do this,
feel free to comment.

Closes # PWX22798

Testing note:
Tested with Shared4SvcMaintainNode
`ginkgo --focus Shared4SvcMaintainNode -v bin/basic.test -- -spec-dir `pwd`/drivers/scheduler/k8s/specs --app-list test-sv4-svc`

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -2340,25 +2340,14 @@ func (d *portworx) getNodeManagerByAddress(addr string) (api.OpenStorageNodeClie
 
 func (d *portworx) maintenanceOp(n node.Node, op string) error {
 	var err error
-	// TODO replace by sdk call whenever it is available
-	pxdRestPort, err := getRestPort()
+
+	logrus.Warnf("unable to get service endpoint falling back to node addr: err=%v, skipPXSvcEndpoint=%v", err, d.skipPXSvcEndpoint)
+	pxdRestPort, err := getRestContainerPort()
 	if err != nil {
 		return err
 	}
-	var url, endpoint string
-	if !d.skipPXSvcEndpoint {
-		endpoint, err = d.schedOps.GetServiceEndpoint()
-	}
-	if err != nil || endpoint == "" {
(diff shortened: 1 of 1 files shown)


thiguetta (changes requested):
we need to find a better solution for this since we use this in Cloud jobs and we don't have access to node IP directly since it's not public, we do it through Load Balancers, we need to figure it out for both scenarios

dahuang-purestorage:
hey @thiguetta can you share which cloud job/jenkins job you are referring to? I think in general if you set hostNetwork to true, pods should have access to the node via node's address.

drivers/volume/portworx/portworx.go
```diff
@@ -2340,25 +2340,14 @@ func (d *portworx) getNodeManagerByAddress(addr string) (api.OpenStorageNodeClie
 
 func (d *portworx) maintenanceOp(n node.Node, op string) error {
 	var err error
-	// TODO replace by sdk call whenever it is available
-	pxdRestPort, err := getRestPort()
+
+	logrus.Warnf("unable to get service endpoint falling back to node addr: err=%v, skipPXSvcEndpoint=%v", err, d.skipPXSvcEndpoint)
```
pureneelesh, line 2344: this warning is misleading; need to reword it

also, add a comment about why we cannot use the service endpoint here

pureneelesh (approved):

drivers/volume/portworx/portworx.go
dahuang-purestorage: updated

PR #727, opened 2022-02-24
PWX-20722: add pool expand test suite to the deploy file

trierra:

Closes #

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -167,7 +167,8 @@ esac
 done
 
 if [[ -z "$TEST_SUITE" || "$TEST_SUITE" == "" ]]; then
-    TEST_SUITE='"bin/asg.test",
+    TEST_SUITE='
+            "bin/asg.test",
             "bin/autopilot.test",
             "bin/basic.test",
             "bin/backup.test",
@@ -183,7 +184,9 @@ if [[ -z "$TEST_SUITE" || "$TEST_SUITE" == "" ]]; then
             "bin/sharedv4.test",
             "bin/telemetry.test",
             "bin/upgrade_cluster.test",
-            "bin/pxcentral.test",'
+            "bin/pxcentral.test",
+            "bin/storage_pool.test",
+'
 else
   TEST_SUITE=$(echo \"$TEST_SUITE\" | sed "s/,/\",\n\"/g")","
 fi
--- a/tests/storage_pool/storage_pool_test.go
+++ b/tests/storage_pool/storage_pool_test.go
@@ -10,65 +10,199 @@ import (
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/reporters"
 	. "github.com/onsi/gomega"
+	"github.com/portworx/sched-ops/task"
+	"github.com/portworx/torpedo/drivers/scheduler"
+	"github.com/portworx/torpedo/pkg/units"
 	. "github.com/portworx/torpedo/tests"
+	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+const poolResizeTimeout = time.Minute * 90
+const retryTimeout = time.Minute
 
 func TestStoragePool(t *testing.T) {
 	RegisterFailHandler(Fail)
 	var specReporters []Reporter
 	junitReporter := reporters.NewJUnitReporter("/testresults/junit_basic.xml")
 	specReporters = append(specReporters, junitReporter)
-	RunSpecs(t, "StoragePoolResize Suite")
+	RunSpecs(t, "StoragePoolExpand Suite")
 }
 
 var _ = BeforeSuite(func() {
 	InitInstance()
 })
 
-var _ = Describe("{StoragePoolResize}", func() {
+var _ = Describe("{StoragePoolExpandDiskResize}", func() {
+	var contexts []*scheduler.Context
 
-	It("should get the existing pool", func() {
+	It("has to schedule apps, and expand it by resizing a disk", func() {
+		contexts = make([]*scheduler.Context, 0)
+
+		for i := 0; i < Inst().GlobalScaleFactor; i++ {
+			contexts = append(contexts, ScheduleApplications(fmt.Sprintf("voldriverdown-%d", i))...)
+		}
+
+		ValidateApplications(contexts)
+
+		var poolIDToResize string
+
 		pools, err := Inst().V.ListStoragePools(metav1.LabelSelector{})
-		if err != nil {
-			Expect(err).NotTo(HaveOccurred())
-		}
-
-		Expect(len(pools)).ShouldNot(BeEmpty())
-
-		var poolID string
-		for _, pool := range pools {
-			poolID = pool.Uuid
-			break
-		}
-		// Pick any pool and trigger pool resize command
-
-		poolToResize := pools[poolID]
-		oldSize := poolToResize.TotalSize
-		newSize := oldSize * 2
-
-		err = Inst().V.ExpandPool(poolID, api.SdkStoragePool_RESIZE_TYPE_AUTO, newSize)
 		Expect(err).NotTo(HaveOccurred())
 
-		time.Sleep(time.Second * 40)
+		Expect(len(pools)).NotTo(Equal(0))
 
-		It(fmt.Sprintf("triggers pool resize %s to the size of %v", poolToResize.Uuid, newSize), func() {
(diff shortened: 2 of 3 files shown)


deployments/deploy-ssh.sh
```diff
@@ -167,7 +167,8 @@ esac
 done
 
 if [[ -z "$TEST_SUITE" || "$TEST_SUITE" == "" ]]; then
-    TEST_SUITE='"bin/asg.test",
+    TEST_SUITE='
+    "bin/asg.test",
```
mborodin-px, line 171: nit indentation

tests/storage_pool/storage_pool_test.go
```diff
@@ -26,33 +26,71 @@ var _ = BeforeSuite(func() {
 	InitInstance()
 })
 
-var _ = Describe("{StoragePoolResize}", func() {
+var _ = Describe("{StoragePoolExpandDiskResize}", func() {
+
+	Context("When the author has both names", func() {
```
mborodin-px, line 31: what is the author?

tests/storage_pool/storage_pool_test.go
```diff
@@ -26,33 +26,71 @@ var _ = BeforeSuite(func() {
...
-	It("should get the existing pool", func() {
-		pools, err := Inst().V.ListStoragePools(metav1.LabelSelector{})
-		if err != nil {
+			poolToResize := pools[poolID]
+			oldSize := poolToResize.TotalSize
+			newSize := oldSize * 2 / units.GiB
+
+			err = Inst().V.ExpandPool(poolID, api.SdkStoragePool_RESIZE_TYPE_RESIZE_DISK, newSize)
+			Expect(err).NotTo(HaveOccurred())
+			time.Sleep(time.Second * 90)
```
mborodin-px, line 52: can we avoid sleeps and add wait until pool is resized? In some cases it could take more than 90 secods

tests/storage_pool/storage_pool_test.go
```diff
@@ -26,33 +26,71 @@ var _ = BeforeSuite(func() {
...
+			var newPool *api.StoragePool
+			for _, pool := range newPools {
+				if poolID == pool.Uuid {
+					newPool = pool
+					break
+				}
+			}
+			Expect(newPool).ShouldNot(BeNil())
+			newTotalSize := newPool.TotalSize / units.GiB
+			Expect(newTotalSize).Should(BeNumerically(">=", oldSize/units.GiB))
```
mborodin-px, line 65: I think we have to use exact "=" otherwise it will be negative pass test

tests/storage_pool/storage_pool_test.go
```diff
@@ -26,33 +26,71 @@ var _ = BeforeSuite(func() {
...
+
+var _ = Describe("{StoragePoolExpandDiskAdd}", func() {
 
-		var poolID string
-		for _, pool := range pools {
-			poolID = pool.Uuid
-			break
-		}
-		// Pick any pool and trigger pool resize command
+	Context("When the author has both names", func() {
```
mborodin-px, line 72: the same as above, don't know what is the author

tests/storage_pool/storage_pool_test.go
```diff
@@ -26,33 +26,71 @@ var _ = BeforeSuite(func() {
...
 
-		time.Sleep(time.Second * 40)
+			poolToResize := pools[poolID]
+			oldSize := poolToResize.TotalSize
+			newSize := oldSize * 2 / units.GiB
 
-		It(fmt.Sprintf("triggers pool resize %s to the size of %v", poolToResize.Uuid, newSize), func() {
+			err = Inst().V.ExpandPool(poolID, api.SdkStoragePool_RESIZE_TYPE_ADD_DISK, newSize)
+			Expect(err).NotTo(HaveOccurred())
+			time.Sleep(time.Second * 90)
```
mborodin-px, line 93: try to avoid sleeps

tests/storage_pool/storage_pool_test.go
```diff
@@ -64,10 +102,10 @@ var _ = Describe("{StoragePoolResize}", func() {
 				}
 			}
 			Expect(newPool).ShouldNot(BeNil())
-			Expect(newPool.TotalSize).Should(Equal(newSize))
+			newTotalSize := newPool.TotalSize / units.GiB
+			Expect(newTotalSize).Should(BeNumerically(">=", oldSize/units.GiB))
```
mborodin-px, line 106: the same as above comment

tests/storage_pool/storage_pool_test.go
trierra: ah, sorry, I've copied it from the example and forgotten to remove

tests/storage_pool/storage_pool_test.go
trierra: like add the backoff?

mborodin-px: not a backoff, but kind of ValidateStoragePool function which will expect the size of the pool and timeout if pool size will not be of that size.

mborodin-px: I have the same concept in storage pool validation for autoipilot

tests/storage_pool/storage_pool_test.go
trierra: we can't do the exact size for the `disk-add`. If we run this test on a multi disk pool, the test will fail, because the test doesn't determine the disk size

tests/storage_pool/storage_pool_test.go
```diff
@@ -2,16 +2,16 @@ package tests
...
 	"github.com/libopenstorage/openstorage/api"
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/reporters"
 	. "github.com/onsi/gomega"
+	"github.com/portworx/torpedo/pkg/units"
 	. "github.com/portworx/torpedo/tests"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	"testing"
+	"time"
```
mborodin-px, line 14: remove these changes  from PR

tests/storage_pool/storage_pool_test.go
```diff
@@ -26,48 +26,104 @@ var _ = BeforeSuite(func() {
...
-		time.Sleep(time.Second * 40)
+			Expect(expandedPool).ShouldNot(BeNil())
+			if expandedPool.TotalSize != poolToResize.TotalSize {
+				newTotalSize = expandedPool.TotalSize / units.GiB
+				break
+			}
+			time.Sleep(time.Second * 90)
+		}
+		Expect(newTotalSize).NotTo(BeNumerically("==", 0), "Pool size should not be 0. Pool expand operation has failed")
+		Expect(newTotalSize).Should(BeNumerically("==", expectedSize), fmt.Sprintf("expected new size '%d' to be equal to the expected size  '%d'", newTotalSize, expectedSize))
```
mborodin-px, line 76: move these steps to the volume driver like ValidateStoragePools and use that method in each test

tests/storage_pool/storage_pool_test.go
```diff
@@ -26,48 +26,104 @@ var _ = BeforeSuite(func() {
...
-		})
+			Expect(expandedPool).ShouldNot(BeNil())
+			if expandedPool.TotalSize != poolToResize.TotalSize {
+				newTotalSize = expandedPool.TotalSize / units.GiB
+				break
+			}
+			time.Sleep(time.Second * 90)
+		}
+		Expect(newTotalSize).NotTo(BeNumerically("==", 0), "Pool size should not be 0. Pool expand operation has failed")
+		Expect(newTotalSize).Should(BeNumerically(">=", expectedSize), fmt.Sprintf("expected new size '%d' to be greater then the expected size '%d'", newTotalSize, expectedSize))
```
mborodin-px, line 125: the same as above, a lot of copy/paste here

tests/storage_pool/storage_pool_test.go
trierra: why?

tests/storage_pool/storage_pool_test.go
trierra: This looks more like a helper function and we should not mix it with the driver behavior. But I agree, I'll remove the code duplication

ggriffiths (changes requested):
Looks like you're still getting some linter failures
```
./tests/storage_pool/storage_pool_test.go:173:6: func getStoragePoolByUUid should be getStoragePoolByUUID
./tests/storage_pool/storage_pool_test.go:173:27: func parameter poolUUid should be poolUUID
make: *** [Makefile:88: lint] Error 1
```

Looks good once you address though

ggriffiths (approved):

PR #728, opened 2022-02-27
PWX-22988: wait for pod termination and increase the timeouts

pureneelesh:
Added a new check in WaitForRunning() function to wait for the deleted
pods to terminate fully. This helps with debugging the test failures
by catching the pods stuck terminating state earlier and avoids
secondary issues such as "device or resource busy" errors during
sharedv4 failover.

Increased the app readiness timeouts by 50% in MultiVolumeMountsForSharedV4
test.

PWX-22988

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2133,7 +2133,7 @@ func (k *K8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time
 		}
 	}
 
-	t := func() (interface{}, bool, error) {
+	isPodTerminating := func() (interface{}, bool, error) {
 		var terminatingPods []string
 		pods, err := k.getPodsForApp(ctx)
 		// sadly, getPodsForApp returns an error if there are no pods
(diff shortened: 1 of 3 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -2133,6 +2133,28 @@ func (k *K8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time
 		}
 	}
 
+	t := func() (interface{}, bool, error) {
```
ggriffiths, line 2136: nit: maybe rename to `isAppTerminating` or something similar? this func is short enough to read and find out what it does, but a name would've made it easier to read

tests/sharedv4/sharedv4_test.go
```diff
@@ -49,41 +49,17 @@ var _ = Describe("{MultiVolumeMountsForSharedV4}", func() {
...
-			timeout = 3 * time.Minute
-		case 2:
-			frequency = 20
-			timeout = 2 * time.Minute
-		case 1:
-			frequency = 10
-			timeout = 1 * time.Minute
-		default:
+
+		chaosLevel := Inst().ChaosLevel
```
ggriffiths, line 53: nice refactor here :+1:

ggriffiths (approved):
mostly lgtm, just one nit

drivers/scheduler/k8s/k8s.go
pureneelesh: It's a good suggestion. I was just trying to be consistent with how these functions are named in torpedo repo. But will change.

PR #729, opened 2022-03-02
PWX-22445 shorten scheduled application prefix

dahuang-purestorage:

we were getting namespace chars exceeds 63 chars error. shortening the prefix to fit the requirement.

Closes # PWX-22445 

Testing on user jenkins upgrade from 2.8.0 to 2.10.0:
http://jenkins.pwx.dev.purestorage.com/job/Users/job/Daniel/job/dahaung-tp-upgrade-pipeline02/48/
~There's currently an issue with memory. will wait for that to be RCA before merging in.~ 
More memory is added to the upgrade test ([PR](https://github.com/portworx/jenkinsfiles/pull/224)).

### DIFF
--- a/tests/upgrade/upgrade_test.go
+++ b/tests/upgrade/upgrade_test.go
@@ -65,7 +65,7 @@ var _ = Describe("{UpgradeVolumeDriver}", func() {
 
 		Step("reinstall and validate all apps after upgrade", func() {
 			for i := 0; i < Inst().GlobalScaleFactor; i++ {
-				contexts = append(contexts, ScheduleApplications(fmt.Sprintf("upgradevolumedriverreinstall-%d", i))...)
+				contexts = append(contexts, ScheduleApplications(fmt.Sprintf("upgradedvolumedriver-%d", i))...)
 			}
 			ValidateApplications(contexts)
 		})


tests/upgrade/upgrade_test.go
```diff
@@ -65,7 +65,7 @@ var _ = Describe("{UpgradeVolumeDriver}", func() {
 
 		Step("reinstall and validate all apps after upgrade", func() {
 			for i := 0; i < Inst().GlobalScaleFactor; i++ {
-				contexts = append(contexts, ScheduleApplications(fmt.Sprintf("upgradevolumedriverreinstall-%d", i))...)
+				contexts = append(contexts, ScheduleApplications(fmt.Sprintf("upgradedvolumedriver-%d", i))...)
```
mborodin-px, line 68: I think it will not schedule new applications as namespaces will be the same as previous scheduled application

dahuang-purestorage: the prefix is different than the initial one. upgradevolumedriver vs upgrade`d`volumedriver.
that would make the app new right?

mborodin-px: ah, missed that

mborodin-px (approved):
LGTM

PR #731, opened 2022-03-03
PWX-22804: Converted some Pure specs to use FB

Pure-AdamuKaapan:
Quick fixup from the last PR, now that we don't have a "convert storage class" method we need to have different specs for both FA and FB. This sets up:
* Elasticsearch on FA DA with some filesystem options
* MySQL on FA DA with no extra options
* NGINX on FB DA


david-enli:
have some discussion off line with app list, we will give this a run as it is self isolated and not impacting other tests LGTM

david-enli (approved):
LGTM

PR #732, opened 2022-03-04
PWX-22804: Removed pure mysql, updated pure wordpress

Pure-AdamuKaapan:
Turns out that Pure cannot use the mysql app as it uses encryption. I've removed those files and instead updated the wordpress app pure specs.


david-enli (approved):
LGTM

PR #735, opened 2022-03-09
PWX-23093 Validates all app counter is active before going into failover

dahuang-purestorage:

Reboot tests were sometimes failing because tests are not stable before going into the failover stage. This PR makes sure the number of active pods are the same as expected number of running pods before going into failover stage.

Closes # PWX-23093

Testing notes: 
Ran torpedo test locally and passed.

### DIFF
--- a/tests/basic/sharedv4_svc_test.go
+++ b/tests/basic/sharedv4_svc_test.go
@@ -400,7 +400,20 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
 
 							Step(fmt.Sprintf("get counters from node %v before %s", attachedNodeBefore.Name, failoverLog),
 								func() {
-									countersBefore = getAppCounters(apiVol, attachedNodeBefore, counterCollectionInterval)
+									activeKeys := 0
+									Eventually(func() bool {
+										countersBefore = getAppCounters(apiVol, attachedNodeBefore, counterCollectionInterval)
+										activeKeys = 0
+										for key := range countersBefore {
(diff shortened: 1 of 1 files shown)


tests/basic/sharedv4_svc_test.go
```diff
@@ -400,7 +400,20 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
...
+										countersBefore = getAppCounters(apiVol, attachedNodeBefore, counterCollectionInterval)
+										activeKeys = 0
+										for key := range countersBefore {
+											if countersBefore[key].active {
+												activeKeys++
+											}
+										}
+										return (activeKeys == numPods)
+									}, 3*time.Minute, 10*time.Second).Should(BeTrue(),
+										"number of active keys did not match %v/%vfor volume %v (%v) for app %v",
```
pureneelesh, line 414: need space before "for volume %v"

There is no need to log the values of activeKeys and numPods since gomego will print those in the failure message But we should probably log the entire countersBefore map to help debug the failure.

pureneelesh (commented):
If the same issue happens on the other side of the failover, we could mistakenly think that the pod has been terminated while it is actually running. We should probably add some code to verify that the pods we think are terminated are really gone.

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: updated.

pureneelesh (approved):

PR #737, opened 2022-03-16
PWX-23108 add sharedv4 mount recovery test

dahuang-purestorage:
This adds a new test to validate px restart recovers after setting a device path to RO. Also, added a new test for test-sharedv4.
Closes # PWX-23108

Tested locally using ginkgo

### DIFF
--- a/drivers/scheduler/k8s/specs/test-sharedv4/pod.yaml
+++ b/drivers/scheduler/k8s/specs/test-sharedv4/pod.yaml
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: test-sv4-dep-svc
+  name: test-sharedv4-dep
 spec:
   {{ if .Replicas }}
   replicas: {{ .Replicas }}
@@ -9,11 +9,11 @@ spec:
   replicas: 3{{ end }}
   selector:
     matchLabels:
-      app: test-sv4-app-svc
+      app: test-sharedv4-app
   template:
     metadata:
       labels:
-        app: test-sv4-app-svc
+        app: test-sharedv4-app
     spec:
       affinity:
         podAntiAffinity:
@@ -23,7 +23,7 @@ spec:
               - key: app
                 operator: In
                 values:
-                - test-sv4-app-svc
+                - test-sharedv4-app
             topologyKey: "kubernetes.io/hostname"
       containers:
       - name: sv4test
@@ -37,17 +37,17 @@ spec:
               fieldRef:
                 fieldPath: metadata.name
           - name: SHARED_FILE
-            value: "/test-sv4-vol-svc/$(MY_POD_NAME)"
+            value: "/test-sharedv4-vol/$(MY_POD_NAME)"
           - name: LOCAL_FILE
             value: "/local-vol/$(MY_POD_NAME)"
         volumeMounts:
-        - name: test-sv4-vol-svc
-          mountPath: /test-sv4-vol-svc
+        - name: test-sharedv4-vol
+          mountPath: /test-sharedv4-vol
         - name: local-vol
           mountPath: /local-vol
       volumes:
-      - name: test-sv4-vol-svc
+      - name: test-sharedv4-vol
         persistentVolumeClaim:
-          claimName: test-sv4-pvc-svc
+          claimName: test-sharedv4-pvc
       - name: local-vol
         emptyDir: {}
--- a/drivers/scheduler/k8s/specs/test-sharedv4/pod.yaml
+++ b/drivers/scheduler/k8s/specs/test-sharedv4/pod.yaml
@@ -1,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-sv4-dep-svc
+spec:
+  {{ if .Replicas }}
+  replicas: {{ .Replicas }}
+  {{ else }}
+  replicas: 3{{ end }}
+  selector:
+    matchLabels:
+      app: test-sv4-app-svc
+  template:
+    metadata:
+      labels:
+        app: test-sv4-app-svc
+    spec:
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - test-sv4-app-svc
+            topologyKey: "kubernetes.io/hostname"
+      containers:
+      - name: sv4test
+        image: portworx/sharedv4-test:torpedo
+        imagePullPolicy: Always
+        command: ["python", "/app/fileio.py"]
+        args: ["--lock", "--interval=0.25", "$(SHARED_FILE)", "$(LOCAL_FILE)"]
+        env:
+          - name: MY_POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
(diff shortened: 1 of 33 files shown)


pureneelesh:
what are the vendor changes for? If you really need them, they will have to come from the openstorage repo.

drivers/scheduler/k8s/specs/test-sharedv4/pod.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-sv4-dep-svc
```
pureneelesh, line 4: Everything in this app should be named test-sharedv4-dep, test-sharedv4-app etc to avoid confusion with the svc app.

tests/basic/sharedv4_svc_test.go
```diff
@@ -1110,7 +1229,7 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
 func getTestSv4Contexts(contexts []*scheduler.Context) []*scheduler.Context {
 	var testSv4Contexts []*scheduler.Context
 	for _, ctx := range contexts {
-		if !strings.HasPrefix(ctx.App.Key, "test-sv4-svc") {
+		if !strings.HasPrefix(ctx.App.Key, "test-sv4-svc") && !strings.HasPrefix(ctx.App.Key, "test-sharedv4") {
```
pureneelesh, line 1232: This will cause the other tests to fail since they assume that the apps in testSv4Contexts are using "service"?

pureneelesh (commented):
partial review. Have not looked at the test itself yet.

dahuang-purestorage:
> what are the vendor changes for? If you really need them, they will have to come from the openstorage repo.

I needed a libopenstorage package for getting the mount info. And after i ran make vendor, it pulled other deps as well.

It's not a necessary import as i can just copy the command that needs to be run.

drivers/scheduler/k8s/specs/test-sharedv4/pod.yaml
dahuang-purestorage: updated

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: added a new context for sharedv4 app

tests/basic/sharedv4_svc_test.go
```diff
@@ -309,7 +313,7 @@ func (fm *failoverMethodMaintenance) sleepBetweenFailovers() time.Duration {
 
 var _ = Describe("{Sharedv4SvcFunctional}", func() {
 	var testrailID, runID int
-	var contexts, testSv4Contexts []*scheduler.Context
+	var contexts, testSv4Contexts, testSharedV4Contexts []*scheduler.Context
```
pureneelesh, line 316: rename testSv4Contexts to testSv4SvcContexts?

Another thing that is a bit odd is that this file, the job everything is named as sharedv4-svc functional tests. So, this new test does not fit well in that scheme. Any thoughts on how we can clean that up?

tests/basic/sharedv4_svc_test.go
```diff
@@ -325,8 +329,9 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
 
 		// Skip the test if there are no test-sharedv4 apps
 		testSv4Contexts = getTestSv4Contexts(contexts)
```
pureneelesh, line 331: getSv4Contexts -> getSv4SvcContexts

tests/basic/sharedv4_svc_test.go
```diff
@@ -325,8 +329,9 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
...
-		if len(testSv4Contexts) == 0 {
-			Skip("No test-sv4-svc apps were found")
+		testSharedV4Contexts = getTestSharedV4Contexts(contexts)
+		if len(testSv4Contexts) == 0 && len(testSharedV4Contexts) == 0 {
+			Skip("No test-sv4-svc or test-sharedv4 apps were found")
 		}
```
pureneelesh, line 335: This is not ideal. If there are only sharedv4 contexts, we will proceed with the sharedv4-svc tests which will be no-ops. But we will mark them as "passed" even thought they were actually "skipped"?

tests/basic/sharedv4_svc_test.go
```diff
@@ -1055,6 +1060,121 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
...
+						}
+					})
+				}
+			})
+
+			JustAfterEach(func() {
+				for _, ctx := range testSharedV4Contexts {
+					_, _, attachedNode := getSv4TestAppVol(ctx)
+					Step(fmt.Sprintf("remove label on node for %s", ctx.App.Key), func() {
+						err := core.Instance().RemoveLabelOnNode(attachedNode.Name, "attachedNode")
```
pureneelesh, line 1171: since "attachedNode" is being used in more than one place, it should be a const

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: good point. i can move it to its own scheme. maybe even to a new file, since this file is also called sharedv4_svc test

tests/basic/sharedv4_svc_test.go
```diff
@@ -1055,6 +1060,121 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
...
+							Step(fmt.Sprintf("scale down app: %s to 0 ", ctx.App.Key), func() {
+								scaleApp(ctx, 0)
+							})
+
+							// wait until all pods are gone
+							Step(fmt.Sprintf("wait for app %s to have 0 pods", ctx.App.Key), func() {
+								waitForNumPodsToEqual(ctx, 0)
+							})
+
+							Step(fmt.Sprintf("scale it backsssss %s", ctx.App.Key), func() {
```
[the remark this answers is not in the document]
dahuang-purestorage: updated

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: do you mean to set it right at the beginning of context? if so, isn't it going to change for each additional context we run with?

dahuang-purestorage: "attachedNode" is removed since we don't need to run on single node.

tests/basic/sharedv4_svc_test.go
```diff
@@ -1055,6 +1060,121 @@ var _ = Describe("{Sharedv4SvcFunctional}", func() {
...
+			BeforeEach(func() {
+				namespacePrefix = "sharedv4mountrecovery"
+			})
+
+			JustBeforeEach(func() {
+				for _, ctx := range testSharedV4Contexts {
+					vol, _, attachedNode := getSv4TestAppVol(ctx)
+					k8sApps := apps.Instance()
+					Step(
+						fmt.Sprintf("setup app %s to deploy on single node", ctx.App.Key),
```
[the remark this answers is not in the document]
dahuang-purestorage: removed.

tests/basic/sharedv4_svc_test.go
dahuang-purestorage: as discussed, sharedv4 apps are moved to its own file and context

pureneelesh (approved):

PR #739, opened 2022-03-17
PWX-22448: add test-sharedv4 service app with aggr=2 and io_profile=db

pureneelesh:
Add another app to the sharedv4 function tests for the test cases
C55359 and C55360.

PWX-22448

I am not able to run this app locally because I don't have enough nodes. I have started a test job on jenkins to see if we need more changes.

Job passed here: http://jenkins.pwx.dev.purestorage.com/view/Control%20Plane/job/Torpedo/job/tp-sharedv4-svc-functional/217/

### DIFF
--- a/drivers/scheduler/k8s/specs/test-sv4-svc-2/pod.yaml
+++ b/drivers/scheduler/k8s/specs/test-sv4-svc-2/pod.yaml
@@ -1,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-sv4-dep-svc
+spec:
+  {{ if .Replicas }}
+  replicas: {{ .Replicas }}
+  {{ else }}
+  replicas: 3{{ end }}
+  selector:
+    matchLabels:
+      app: test-sv4-app-svc
+  template:
+    metadata:
+      labels:
+        app: test-sv4-app-svc
+    spec:
+      affinity:
+        podAntiAffinity:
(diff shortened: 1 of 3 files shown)


drivers/scheduler/k8s/specs/test-sv4-svc-2/pod.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-sv4-2-dep-svc
```
dahuang-purestorage, line 4: is it possible to have a different name other than test-sv4-2-. ?

dahuang-purestorage (approved):
LGTM just minor comment

drivers/scheduler/k8s/specs/test-sv4-svc-2/pod.yaml
pureneelesh: yeah, I don't like that name as well. But I could not come up with something better. This app is similar to test-sv4-svc but with slightly different settings. We may want to add more different settings in the future. Do you have any suggestion?

dahuang-purestorage: yeah i don't know if there's a better name for it now. In the future, i think it's possible to parameterize the different settings. But it might take some time to do it.

PR #752, opened 2022-04-12
PTX-6088: wait for the context to stabilize

pureneelesh:
We should wait for the context to stabilize before starting a new
iteration of the for loop since the context may have become unstable
due to the node down and up for the previous context.

PTX-6088


ggriffiths (approved):
lgtm

PR #755, opened 2022-04-20
Writes more data to Pure volumes using Elasticsearch

Pure-AdamuKaapan:
Currently, the Pure test suite runs a test where we write data to the data directory of an app, and then validate that the bytesUsed has increased. This works in general, but Elasticsearch uses the data directory quite heavily due to the esload app that's running, so sometimes we actually see the bytesUsed decrease instead of increase (due to either the load app or over-time compression?) which causes the test to fail.

I'm hoping that writing much more data (e.g. 5 GiB instead of .5 GiB) will make a more noticeable difference. I had to do some refactoring to make it so that we still only wrote a bit of data to nginx and wordpress, but I think this will be better in the long run.

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -31,6 +31,7 @@ import (
 	"github.com/portworx/sched-ops/task"
 	"github.com/portworx/torpedo/drivers/node"
 	"github.com/portworx/torpedo/pkg/testrailuttils"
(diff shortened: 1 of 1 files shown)


david-enli (approved):
lol nice, LGTM

PR #758, opened 2022-04-20
Ipv6 functional changes

dahuang-purestorage:
Adds changes to torpedo to run on ipv6 clusters.
Closes #

### DIFF
--- a/drivers/backup/portworx/portworx.go
+++ b/drivers/backup/portworx/portworx.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"net"
+	"strconv"
 	"strings"
 	"time"
 
@@ -134,7 +136,7 @@ func (p *portworx) Init(schedulerDriverName string, nodeDriverName string, volum
 }
 
 func (p *portworx) constructURL(ip string) string {
-	return fmt.Sprintf("%s:%d", ip, defaultPxbServicePort)
+	return net.JoinHostPort(ip, strconv.Itoa(int(defaultPxbServicePort)))
 }
 
 func (p *portworx) testAndSetEndpoint(endpoint string) error {
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -3,6 +3,7 @@ package ssh
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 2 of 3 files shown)


drivers/backup/portworx/portworx.go
```diff
@@ -134,7 +136,7 @@ func (p *portworx) Init(schedulerDriverName string, nodeDriverName string, volum
 }
 
 func (p *portworx) constructURL(ip string) string {
-	return fmt.Sprintf("%s:%d", ip, defaultPxbServicePort)
+	return net.JoinHostPort(ip, strconv.Itoa(int(defaultPxbServicePort)))
```
jrivera-px, line 139: Don't need to type cast int(defaultPxbServicePort) since its already an int...

drivers/node/ssh/ssh.go
```diff
@@ -576,7 +577,8 @@ func (s *SSH) getConnectionOnUsableAddr(n node.Node, options node.ConnectionOpts
 	for _, addr := range n.Addresses {
 		t := func() (interface{}, bool, error) {
 			// check if address is responding on port 22
-			conn, err := ssh_pkg.Dial("tcp", fmt.Sprintf("%s:%d", addr, DefaultSSHPort), s.sshConfig)
+			endpoint := net.JoinHostPort(addr, strconv.Itoa(int(DefaultSSHPort)))
```
jrivera-px, line 580: Don't need to type cast int(DefaultSSHPort) since its already an int...

drivers/scheduler/k8s/specs/debug/debug.yaml
```diff
@@ -24,7 +24,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
       - name: debug
-        image: ubuntu:latest
+        image: harbor.pwx.dev.purestorage.com:443/px-image/ubuntu:latest
```
jrivera-px, line 27: Not sure if we want to commit harbor pulls since its actually only accessible from certain network.

drivers/backup/portworx/portworx.go
dahuang-purestorage: i think it's int32, which `strconv.Itoa` complain about. that's why i recaste it to int.

jrivera-px: I thought it was const?

jrivera-px (approved):
LGTM

drivers/scheduler/k8s/specs/debug/debug.yaml
lsrinivas-pure: this might fail existig runs

lsrinivas-pure: can u run this job with utr private image
https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/

lsrinivas-pure (changes requested):
please run the job with your private image so that we dont have any regression
https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/

drivers/scheduler/k8s/specs/debug/debug.yaml
dahuang-purestorage: removed the charts. will work on running the job

dahuang-purestorage:
> please run the job with your private image so that we dont have any regression https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/

job passed: https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/999/

lsrinivas-pure (approved):
LGTM

PR #760, opened 2022-04-21
Shrunk Pure Elastic data file size down to 2 GiB

Pure-AdamuKaapan:
Looks like my last PR of writing 5 GiB was a bit too much, depending on how big elastics data volume was. Let's try 2 GiB instead.


david-enli (approved):
LGTM

PR #762, opened 2022-04-22
Changed a fmt.Println for a log line

Pure-AdamuKaapan:
Fixes a log line which bled over into the next step


david-enli (approved):
LG

PR #765, opened 2022-04-22
If no pods are found for the specs in current context, bail out during pod terminating check.

diptiranjanpx:
Signed-off-by: Diptiranjan

After doing the restore, for checking if the related resources are properly restored, we call WaitForRunning()
Now because pod terminating check got added, it is not finding the pods with owner reference as the specs in current context(which are basically the resources that torpedo created in backup cluster).

Closes # PB-2327

*Test*
https://jenkins.portworx.dev/job/PX-Backup/view/PX-Master/job/px-backup-integration/874/

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2085,10 +2085,8 @@ func (k *K8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time
 		var terminatingPods []string
 		pods, err := k.getPodsForApp(ctx)
 		// sadly, getPodsForApp returns an error if there are no pods
-		if err != schederrors.ErrPodsNotFound {
+		if err != nil {
 			return nil, false, nil
-		} else if err != nil {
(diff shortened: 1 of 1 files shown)


diptiranjanpx:
Torpedo run with the pvt torpedo image: https://jenkins.portworx.dev/job/Torpedo/job/tp-dev-test/1001/

siva-portworx (approved):
LGTM

drivers/scheduler/k8s/k8s.go
```diff
@@ -2085,10 +2085,8 @@ func (k *K8s) WaitForRunning(ctx *scheduler.Context, timeout, retryInterval time
 		var terminatingPods []string
 		pods, err := k.getPodsForApp(ctx)
 		// sadly, getPodsForApp returns an error if there are no pods
-		if err != schederrors.ErrPodsNotFound {
+		if err != nil {
 			return nil, false, nil
```
pureneelesh, line 2089: This returns success after there is an error while getting the pods. What if there was an error such as API server was busy, or communication error etc.? We would falsely think that all the pods are gone?

Is it possible to ignore only the specific error that you are hitting in the backup scenario?

PR #767, opened 2022-04-26
Fixed wrong error check for Pure groupsnaps

Pure-AdamuKaapan:
Looks like the groupsnap check has been commented out so long that it still had the wrong error message check. This adds the correct error message


david-enli (approved):
LGTM

PR #770, opened 2022-04-26
Splits specs up so there are distinct FA and FB ones

Pure-AdamuKaapan:
FA apps: elasticsearch, mysql-without-enc
FB apps: nginx-without-enc, wordpress

Also adds a new "mysql-without-enc" app so we can test using mysql for FA direct access (which is why there's a bunch of AWS and Azure specs in this PR too).

After this is merged I'll copy the existing jobs, and then make one run the FA apps and one run the FB apps.


drivers/scheduler/k8s/specs/wordpress/pure/pure-storage-class.yaml
```diff
@@ -18,8 +17,7 @@ metadata:
 provisioner: pxd.portworx.com
 parameters:
   # Tests:
-  # * FlashArray Direct Access w/ filesystem
-  # * Default filesystem type (ext4), mount options and create options
-  # * No QoS
-  backend: "pure_block"
+  # * FlashBlade Direct Access
+  # * Default export rules
+  backend: "pure_file"
```
dsupure, line 22: Why we want to remove FA DA from the wordpress test?
Why we need two SCs (wordpress-sc-legacy-shared, wordpress-sc) for FBDA?

Pure-AdamuKaapan: The storage classes are specified as the same for all providers (portworx, azure, aws, etc.) so this file has to provide both.

Right now wordpress has both block and file so we can't split it and run "only FA" or "only FB". To fix that I'm making wordpress FB only and making another mysql app that's FA only. Then, to run FA-only or FB-only tests, we'll use the combo in the PR description.

dsupure: Gocha, thanks for the explanation.

dsupure (approved):
LGTM!

PR #771, opened 2022-04-26
Add netutil pkg and MakeURL func to create URIs which handle IPv6 add…

jrivera-px:
…resses.

Fix conflicts...
Add pkg to be  used to create URI's which take into consideration IPv6 host addresses in the URI.


lsrinivas-pure (approved):
LGTM

PR #772, opened 2022-04-27
Adds a data directory for Pure mysql app

Pure-AdamuKaapan:
BytesUsed wasn't changing for mysql because it didn't have a data directory, so we were just writing to the container root. We will now write to the correct directory, so that check should work now.


dsupure (approved):
LGTM! Thanks for finding the root casue!

PR #774, opened 2022-04-27
Reverted wordpress to FA+FB, and added pure specs to rabbitmq

Pure-AdamuKaapan:
Turns out that MySQL (in Wordpress) has some issues running on FB. They are fixable in a customer env but not worth our time here.
1. MySQL tries to chown the `.snapshot` directory which is a special directory
2. After manually disabling snapshots (which can also be done with an env variable), MySQL had some issues with block sizes likely since it's running on NFS

I decided to replace this with rabbitmq (just scrolled the list until I found a simple one that I liked)


dsupure (approved):
LGTM!

PR #775, opened 2022-04-27
PTX-6403: fix nil panic caused by nil parameter

trierra:
the nil parameter in line causes nil panic later in the thread, fixed the parameter check 

 
Closes #

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -1386,7 +1386,7 @@ func (k *K8s) createVolumeSnapshotRestore(specObj interface{},
 }
 
 func (k *K8s) addSecurityAnnotation(spec interface{}, configMap *corev1.ConfigMap) error {
-	//logrus.Debugf("Config Map details: %v", configMap.Data)
+	// logrus.Debugf("Config Map details: %v", configMap.Data)
 	secretNameKeyFlag := false
 	secretNamespaceKeyFlag := false
 	encryptionFlag := false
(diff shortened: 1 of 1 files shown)


nikolaypopov (approved):
LGTM

dahuang-purestorage (approved):
LGTM

drivers/scheduler/k8s/k8s.go
```diff
@@ -2613,10 +2613,14 @@ func (k *K8s) ValidateVolumes(ctx *scheduler.Context, timeout, retryInterval tim
...
 		} else if obj, ok := specObj.(*v1.PersistentVolumeClaim); ok {
 			err := k8sCore.ValidatePersistentVolumeClaim(obj, timeout, retryInterval)
-			if err != nil && !options.ExpectError {
-				return &scheduler.ErrFailedToValidateStorage{
-					App:   ctx.App,
-					Cause: fmt.Sprintf("Failed to validate PVC: %v. Err: %v", obj.Name, err),
+			if err != nil {
+				if options != nil && options.ExpectError {
+					// ignore
+				} else {
```
piyush-nimbalkar, line 2619: nit: We could also write it like this - 
```
if err != nil && (options == nil || !options.ExpectError) {
    return &scheduler.ErrFailedToValidateStorage{...}
}
```

trierra: yep, I could do that too, it just a bit easier to read for me in this way

PR #778, opened 2022-04-28
Migrate Ginkgo to v2

fmilichovsky:
Transitions ginkgo to v2 following https://onsi.github.io/ginkgo/MIGRATING_TO_V2.

No special deprecation warnings found from `ginkgo ./...`, I only had to rename one type alias to avoid a dot-imported naming conflict.

Summary on `master` (before migration):
```
$ ginkgo ./...
[...]
Ginkgo ran 25 suites in 2m31.035273817s

There were failures detected in the following suites:
    asg ./tests/asg
  basic ./tests/basic [Compilation failure]
```

Summary after migration:
```
$ ginkgo ./...
Ginkgo ran 25 suites in 3m2.689931081s

There were failures detected in the following suites:
    asg ./tests/asg
  basic ./tests/basic [Compilation failure]
```


fmilichovsky:
Hmm looks like `CustomReporters` [were deprecated](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-custom-reporters).

If they're not just unused boilerplate and we actually do have reports relying on them, they'd need to get migrated:
https://github.com/portworx/torpedo/blob/a490e2e89867c8e972a78a02ba198b70565cc302/tests/backup/backup_test.go#L60-L62

fmilichovsky:
Something seems off in that job - it succeeds pretty much instantly, nothing substantial is actually getting executed by the looks of it:
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Aniket/job/tp-basic-sharedv4-aghodke/lastBuild/

But anyway, I'm not sure if a single focused job is a good enough check for a repo-wide upgrade, but I take it we don't have that kind of all-encompassing job that would do some kind of dry-run with mocked-out inputs and focus more on the spec validity itself :/

aghodke1312:
I have started a run: https://jenkins.pwx.dev.purestorage.com/job/Users/job/Aniket/job/tp-basic-sharedv4/1/console

fmilichovsky:
Alright, so the run with the `master` version passed while the migrated version is less tolerant of the current structure and complained here:
```
20:51:31 STEP: validate applications 04/28/22 18:51:31.814

20:51:31 Ginkgo detected an issue with your spec structure

20:51:31 /go/src/github.com/portworx/torpedo/tests/common.go:366

20:51:31   It looks like you are trying to add a [Container] node

20:51:31   to the Ginkgo spec tree in a leaf node after the specs started running.
```
And that's coming from just one focused package, no info about all the others. Currently it doesn't look like it's gonna be worth pushing the migration further, unless a stronger incentive comes.

PR #780, opened 2022-04-28
Attempted fix for pxctl being missing after successful startup

Pure-AdamuKaapan:
We've seen a few test runs where PX supposedly installed successfully, and pxctl did eventually become available, but there seems to be some period in the middle where pxctl isn't available in the PATH. This PR tries replacing the short `pxctl` with the full path.

### DIFF
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -579,7 +579,7 @@ func (s *SSH) doCmdSSH(n node.Node, options node.ConnectionOpts, cmd string, ign
 		return "", fmt.Errorf("fail to setup stdout")
 	}
 	if options.Sudo {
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -3420,6 +3422,15 @@ func (d *portworx) getPxctlPath(n node.Node) string {
 	return strings.TrimSpace(out)
 }
 
+func (d *portworx) updatePxctlPath(n node.Node, cmd string) string {
+	if !strings.HasPrefix(cmd, "pxctl") {
```
dsupure, line 3426:
```suggestion
	if !strings.Contains(cmd, "pxctl") {
```

dsupure (approved):
LGTM. Make sure we don't cause regression before merging.

PR #781, opened 2022-04-29
[PTX-6425] Get complete list of nodes while checking for core files

vinayakshnd:
Get complete list of nodes while checking for core files 
not just worker nodes.(On IBM OpenShift setup, nodes are marked
as both worker as well as master.)

Closes #


lsrinivas-pure (approved):
LGTM

mborodin-px (commented):
LGTM

PR #782, opened 2022-04-29
Add test skeleton for PDS

fmilichovsky:
Adds a root spec node for PDS, with no actual logic being tested.
I want to focus on the test being set up correctly and being able to inspect the control plane (pulling the logs on a failed test).

I wanna use this stub to
- Agree on the method of passing configuration to the test (kubeconfigs so far).
- Verify that we're able to talk to the control plane in the jenkins job (getting the logs). Ignoring target cluster for now, I imagine that's gonna be tricky because of the security situation.
- Check the output of a failing job, see if it's usefully formatted.


tests/pds/pds_test.go
```diff
@@ -0,0 +1,78 @@
...
+	RunSpecs(t, "Torpedo: PDS")
+}
+
+var (
+	controlPlane  *pds.ControlPlane
+	targetCluster *pds.TargetCluster
+)
+
+var _ = BeforeSuite(func() {
+	cpKubeconfig, err := tests.GetSourceClusterConfigPath()
```
fmilichovsky, line 27: @aghodke1312 Let me know if you wanna change the way we pass the kubeconfigs to the test.
Reusing this helper assumes there are two comma-separated paths in the `KUBECONFIGS` env variable (first being the control plane, second being the target cluster in our case).

tests/pds/pds_test.go
```diff
@@ -0,0 +1,78 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cpKubeconfig).To(BeAnExistingFile())
+	controlPlane, err = pds.NewControlPlane(cpKubeconfig)
+	Expect(err).NotTo(HaveOccurred())
+
+	targetKubeconfig, err := tests.GetDestinationClusterConfigPath()
```
fmilichovsky, line 33: We're gonna have to think about how we're going to be talking to both clusters when each of them is likely to have completely separate auth.

fmilichovsky:
Closing, transitioning to https://github.com/portworx/pds-integration-test/pull/19.

PR #785, opened 2022-05-03
Added more descriptive error messages to Pure test asserts

Pure-AdamuKaapan:
Adds some more descriptive error messages so that the Torpedo error description makes a bit more sense. Currently we see some failures like "expected (false) to be (true)" which is super opaque. This will make it at least a bit more obvious what went wrong, reducing triage time and allowing some automated triage.


dsupure (approved):
LGTM. Thanks.

PR #788, opened 2022-05-06
:hammer: mongob changes to enable ipv6

lsrinivas-pure:
Enabling IPV6 in mongodb spec

Closes #

Tested both on ipv4 and ipv6 environment and it worked fine


lsrinivas-pure:
> These specs are being used in ipv4 environment as well. We should have a separate copy for ipv6 specs. If there is an environment with ipv4 and ipv6 both enabled and user wants to use ipv4 stack, then this might cause an issue.

this spec is tested for both ipv6 and ipv4 and it worked fine. For remaining specs there is no change required for ipv6/ipv4

jitupawar22 (approved):
LGMT if its working fine for both IPv4 and IPv6.

PR #799, opened 2022-06-01
:hammer: Email reporting  frquency fix for longevity

lsrinivas-pure:
Creating a separate thread for email report so that it is send as per the frequency

Closes #PTX-6880


Rohit-PX (approved):
Looks ok, to me except one question, if the email trigger fails, are we collecting it that info anywhere?

jainsmit (approved):
LGTM !!

PR #807, opened 2022-06-14
:sparkles: Adding kvdb failover test in longevity

lsrinivas-pure:
Added KVDB failove test as part og longevity
Added  logic to deploy Apps hyper-converged

Closes # PTX-5414, PTX-5598

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -985,11 +985,22 @@ func ScheduleApplications(testname string, errChan ...*chan error) []*scheduler.
 	var err error
 
 	Step("schedule applications", func() {
-		taskName := fmt.Sprintf("%s-%v", testname, Inst().InstanceID)
-		contexts, err = Inst().S.Schedule(taskName, scheduler.ScheduleOptions{
+		options := scheduler.ScheduleOptions{
 			AppKeys:            Inst().AppList,
 			StorageProvisioner: Inst().Provisioner,
-		})
+		}
+		//if not hyper converged set up deploy apps only on storageless nodes
+		if !Inst().IsHyperConverged {
+			storagelessNodes := node.GetStorageLessNodes()
+			options = scheduler.ScheduleOptions{
+				AppKeys:            Inst().AppList,
+				StorageProvisioner: Inst().Provisioner,
+				Nodes:              storagelessNodes,
+			}
+
+		}
+		taskName := fmt.Sprintf("%s-%v", testname, Inst().InstanceID)
+		contexts, err = Inst().S.Schedule(taskName, options)
 		processError(err, errChan...)
 		if len(contexts) == 0 {
 			processError(fmt.Errorf("list of contexts is empty for [%s]", taskName), errChan...)
@@ -1076,6 +1087,7 @@ func StartVolDriverAndWait(appNodes []node.Node, errChan ...*chan error) {
 		})
 
 	})
+
 }
 
 // StopVolDriverAndWait stops volume driver on given app nodes and waits till driver is down
@@ -3100,6 +3112,7 @@ type Torpedo struct {
 	AutopilotUpgradeImage               string
 	CsiGenericDriverConfigMap           string
 	HelmValuesConfigMap                 string
+	IsHyperConverged                    bool
 }
 
 // ParseFlags parses command line flags
--- a/tests/common.go
+++ b/tests/common.go
@@ -991,13 +991,30 @@ func ScheduleApplications(testname string, errChan ...*chan error) []*scheduler.
 		}
 		//if not hyper converged set up deploy apps only on storageless nodes
 		if !Inst().IsHyperConverged {
+			logrus.Infof("Scheduling apps only on storageless nodes")
 			storagelessNodes := node.GetStorageLessNodes()
+			if len(storagelessNodes) == 0 {
+				logrus.Info("No storageless nodes available in the PX Cluster. Setting HyperConverges as true")
+			}
+			for _, storagelessNode := range storagelessNodes {
+				if err = Inst().S.AddLabelOnNode(storagelessNode, "storage", "NO"); err != nil {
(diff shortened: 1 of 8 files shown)


tests/common.go
```diff
@@ -985,11 +985,36 @@ func ScheduleApplications(testname string, errChan ...*chan error) []*scheduler.
...
-		contexts, err = Inst().S.Schedule(taskName, scheduler.ScheduleOptions{
+		options := scheduler.ScheduleOptions{
 			AppKeys:            Inst().AppList,
 			StorageProvisioner: Inst().Provisioner,
-		})
+		}
+		//if not hyper converged set up deploy apps only on storageless nodes
+		if !Inst().IsHyperConverged {
+			logrus.Infof("Scheduling apps only on storageless nodes")
+			storagelessNodes := node.GetStorageLessNodes()
```
jainsmit, line 995: Add a check for a cluster where no storage-less node is available. Code should check and take the appropriate action !!

tests/longevity/longevity_test.go
```diff
@@ -117,6 +121,8 @@ var _ = Describe("{Longevity}", func() {
 			Inst().TopologyLabels = labels
 		}
 
+		Inst().IsHyperConverged = hyperConvergedTypeEnabled
```
jainsmit, line 124: Is HyperConverged enable for all type of clusters or distributions ?

jainsmit:
LGTM !! Except few comments ..

tests/common.go
lsrinivas-pure: In current logic if there are no storageless nodes, apps will deployed as if hyperconverged is Yes.
I can add a statement that no storageless nodes found deploying apps accross all nodes.
Is that ok?

lsrinivas-pure: Added the comment

jainsmit (approved):

tests/testTriggers.go
```diff
@@ -4341,6 +4345,176 @@ func createLongevityJiraIssue(event *EventRecord, err error) {
...
+				logrus.Errorf(err.Error())
+				UpdateOutcome(event, err)
+			}
+
+		})
+	})
+}
+
+func validateKVDBMembers(event *EventRecord, kvdbMembers map[string]*volume.MetadataNode, isDestuctive bool) bool {
+	logrus.Infof("Current KVDB members: %v", kvdbMembers)
```
Rohit-PX, line 4490: Remove this line before merging if we don't need it.

tests/testTriggers.go
```diff
@@ -4341,6 +4345,176 @@ func createLongevityJiraIssue(event *EventRecord, err error) {
...
+
+	allHealthy := true
+
+	if len(kvdbMembers) == 0 {
+		return false
+	}
+
+	for id, m := range kvdbMembers {
+		logrus.Infof("validating node : %v, is leader?: %v", id, m.Leader)
```
Rohit-PX, line 4499: Remove line before merging.

tests/testTriggers.go
```diff
@@ -4341,6 +4345,176 @@ func createLongevityJiraIssue(event *EventRecord, err error) {
...
+	})
+}
+
+func validateKVDBMembers(event *EventRecord, kvdbMembers map[string]*volume.MetadataNode, isDestuctive bool) bool {
+	logrus.Infof("Current KVDB members: %v", kvdbMembers)
+
+	allHealthy := true
+
+	if len(kvdbMembers) == 0 {
+		return false
```
Rohit-PX, line 4495: We should update outcome here saying `no kvdb members to validate`

tests/testTriggers.go
```diff
@@ -4341,6 +4345,176 @@ func createLongevityJiraIssue(event *EventRecord, err error) {
...
+	}
+
+	for id, m := range kvdbMembers {
+		logrus.Infof("validating node : %v, is leader?: %v", id, m.Leader)
+
+		if !m.IsHealthy {
+			err := fmt.Errorf("kvdb member node: %v is not healthy", id)
+			allHealthy = allHealthy && false
```
Rohit-PX, line 4503: this is the same as `allHealthy = false` ?

tests/testTriggers.go
```diff
@@ -4341,6 +4345,176 @@ func createLongevityJiraIssue(event *EventRecord, err error) {
...
+			logrus.Warn(err.Error())
+			if isDestuctive {
+				UpdateOutcome(event, err)
+			}
+		} else {
+			allHealthy = allHealthy && true
```
Rohit-PX, line 4509: In this case, allHealthy will not change at all i.e it will remain true if it was already true and will remain false if it was already false?
So you don't need this statement I think.

Rohit-PX (changes requested):
Looks ok after these changes.

tests/testTriggers.go
lsrinivas-pure: Hi Rohit, I have added this to track what are kvdb members before failover and and after failover in case of any issue.
Dev asked me to add this logging while triaging one kvdb issue

tests/testTriggers.go
lsrinivas-pure: no.. here even if one node is not healthy, allhealthy will be false, if we set Healthy = false, there is a chance that second node is not healthy and third is healthy , then allHealthy will be true which is not the actual case

jainsmit (approved):

tests/common.go
```diff
@@ -985,11 +985,39 @@ func ScheduleApplications(testname string, errChan ...*chan error) []*scheduler.
...
+			if len(storagelessNodes) == 0 {
+				logrus.Info("No storageless nodes available in the PX Cluster. Setting HyperConverges as true")
```
Rohit-PX, line 997: After this shouldn't we be setting hyperconverged as true ?

Rohit-PX (approved):
LGTM.

PR #810, opened 2022-06-23
Fixing bug in csisnapshot

jainsmit:
1. Fixing the issue with Csi Snapshot and restore code.
2. Added code for wait for snapshot
3. Added code to check DM path in volume inspect
4. Added code to check pure file volume
5. Added code to not trigger clones for FB volumes

Closes #

http://jen-endurance.pwx.dev.purestorage.com/job/Torpedo/job/tp-fada-longevity/67/console
https://jenkins.pwx.dev.purestorage.com/job/Dev/job/torpedo-dev03/36/console

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -124,6 +124,8 @@ const (
 	DefaultRetryInterval = 10 * time.Second
 	// DefaultTimeout default timeout
 	DefaultTimeout = 3 * time.Minute
+	// SnapshotReadyTimeout timeout for snapshot to be ready
+	SnapshotReadyTimeout = 5 * time.Minute
 
 	autopilotServiceName          = "autopilot"
 	autopilotDefaultNamespace     = "kube-system"
@@ -186,6 +188,9 @@ var (
 	k8sBatch           = batch.Instance()
 	k8sMonitoring      = prometheus.Instance()
 	k8sPolicy          = policy.Instance()
+
+	// k8sExternalsnap is a instance of csisnapshot instance
+	k8sExternalsnap = csisnapshot.Instance()
 	// SnapshotAPIGroup is the group for the resource being referenced.
 	SnapshotAPIGroup = "snapshot.storage.k8s.io"
 )
@@ -639,8 +644,18 @@ func validateSpec(in interface{}) (interface{}, error) {
 // direct access volume, only if PureVolumes is true. If PureVolumes is false, all
 // volumes are valid.
 func (k *K8s) filterPureVolumesIfEnabled(claim *v1.PersistentVolumeClaim) (bool, error) {
+	volTypes := []string{PureFile, PureBlock}
+
+	return k.filterPureTypeVolumeIfEnabled(claim, volTypes)
+}
+
+// filterPureTypeVolumeIfEnabled will return true if matches with given type pure volume
+// If PureVolumes is false, all volumes are valid.
+func (k *K8s) filterPureTypeVolumeIfEnabled(claim *v1.PersistentVolumeClaim, volTypes []string) (bool, error) {
+
 	if !k.PureVolumes {
-		return true, nil // If we aren't filtering for Pure volumes, all are valid
+		// If we aren't filtering for Pure volumes, all are valid
+		return true, nil
 	}
 
 	scForPvc, err := k8sCore.GetStorageClassForPVC(claim)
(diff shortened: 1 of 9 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1992,6 +2006,35 @@ func (d *portworx) IsStorageExpansionEnabled() (bool, error) {
 
 // IsPureVolume return true if volume is pure volume else false
 func (d *portworx) IsPureVolume(volume *torpedovolume.Volume) (bool, error) {
+	var proxySpec *api.ProxySpec
+	var err error
+	if proxySpec, err = d.getProxySpecForAVolume(volume); err != nil {
+		return false, err
+	}
+	if proxySpec == nil {
```
Rohit-PX, line 2014: This check is present twice.

drivers/volume/portworx/portworx.go
```diff
@@ -1992,6 +2006,35 @@ func (d *portworx) IsStorageExpansionEnabled() (bool, error) {
...
+
+	if proxySpec == nil {
+		return false, nil
+	}
+	if proxySpec.ProxyProtocol == api.ProxyProtocol_PROXY_PROTOCOL_PURE_BLOCK {
+		logrus.Debugf("Volume is Pure Block volume: %s", volume.ID)
+		return true, nil
+	}
+
+	if proxySpec.ProxyProtocol == api.ProxyProtocol_PROXY_PROTOCOL_PURE_FILE {
```
Rohit-PX, line 2026: If condition in line 2021 is true, this line will never be executed as the function would return from line 2021.

If we need to check only one of the conditions, a simpler/better way would be:

```
if proxySpec.ProxyProtocol == api.ProxyProtocol_PROXY_PROTOCOL_PURE_BLOCK  || proxySpec.ProxyProtocol == api.ProxyProtocol_PROXY_PROTOCOL_PURE_FILE {
```

tests/testTriggers.go
```diff
@@ -1240,6 +1259,22 @@ func TriggerVolumeClone(contexts *[]*scheduler.Context, recordChan *chan *EventR
 			})
 			for _, vol := range appVolumes {
 				var clonedVolID string
+
+				// Skip clone trigger for Pure FB volumes
+				isPureFileVol, err := Inst().V.IsPureFileVolume(vol)
+				if err != nil {
+					logrus.Errorf("Checking pure file volume is failing with error: %v", err)
```
Rohit-PX, line 1266: It will be helpful to print the volume name/ID along with the error here.

tests/testTriggers.go
```diff
@@ -4355,6 +4414,33 @@ func rangeStructer(args ...interface{}) []interface{} {
...
+		ObjectMeta:        v1obj,
+		Provisioner:       k8s.CsiProvisioner,
+		Parameters:        params,
+		ReclaimPolicy:     &reclaimPolicyDelete,
+		VolumeBindingMode: &bindMode,
+	}
+
+	sc, err := k8sStorage.CreateStorageClass(&scObj)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create CsiSnapshot storage class: %s", name)
```
Rohit-PX, line 4439:
```suggestion
		return nil, fmt.Errorf("failed to create CsiSnapshot storage class: %s. Error: %v", name, err)
```

drivers/scheduler/k8s/k8s.go
```diff
@@ -4898,21 +4918,34 @@ func (k *K8s) RecycleNode(n node.Node) error {
...
 	for _, specObj := range ctx.App.SpecList {
+
 		if obj, ok := specObj.(*corev1.PersistentVolumeClaim); ok {
 			pvc, _ := k8sCore.GetPersistentVolumeClaim(obj.Name, obj.Namespace)
-			snapshotOkay, err := k.filterPureVolumesIfEnabled(pvc)
+			snapshotOkay, err := k.filterPureTypeVolumeIfEnabled(pvc, volTypes)
 			if err != nil {
 				return nil, err
 			}
 			if snapshotOkay {
```
Rohit-PX, line 4934: add and else for what happens if `snapshotOkay` is false

Rohit-PX (changes requested):
Looks good except the minor changes. Also post the link to test torpedo runs.

tests/testTriggers.go
jainsmit: Incorporated !!

tests/testTriggers.go
jainsmit: Incorporated !!

Rohit-PX (approved):

Pure-AdamuKaapan (approved):
LGTM! Cool to see more Pure tests making it in

PR #811, opened 2022-06-28
:hammer: Email fix for IKS and storagepool test

lsrinivas-pure:
This fixes Storagepool test where journal device is not considered.
Also fixes email in longevity for IKS as in IKS all nodes are marked as worker nodes

Closes #

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -26,6 +26,7 @@ import (
 	clusterclient "github.com/libopenstorage/openstorage/api/client/cluster"
 	"github.com/libopenstorage/openstorage/api/spec"
 	"github.com/libopenstorage/openstorage/cluster"
+	v1 "github.com/libopenstorage/operator/pkg/apis/core/v1"
 	optest "github.com/libopenstorage/operator/pkg/util/test"
 	"github.com/pborman/uuid"
 	"github.com/portworx/sched-ops/k8s/apiextensions"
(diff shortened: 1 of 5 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -3612,6 +3627,26 @@ func (d *portworx) UpdateStorageClusterImage(imageName string) error {
 
 }
 
+func (d *portworx) GetPXStorageCluster() (*v1.StorageCluster, error) {
+	pxOps, err := pxOperator.ListStorageClusters(schedops.PXNamespace)
+	if err != nil {
+		er := fmt.Errorf("Error getting Storage Clusters list, Err: %v", err.Error())
```
Rohit-PX, line 3633: you can still use `err` here instead of `er` since it's a local variable because you are using `:=` in line 3633

tests/testTriggers.go
```diff
@@ -1883,15 +1883,17 @@ func TriggerEmailReporter() {
...
-		status, err := Inst().V.GetNodeStatus(n)
-		if err != nil {
-			pxStatus = "ERROR GETTING STATUS"
-		} else {
-			pxStatus = status.String()
-		}
+		if n.StorageNode != nil {
+			status, err := Inst().V.GetNodeStatus(n)
+			if err != nil {
+				pxStatus = "ERROR GETTING STATUS"
```
Rohit-PX, line 1889: Improvement:  define `ERROR GETTING STATUS` as a constant

Rohit-PX (approved):
Looks ok except for the minor changes suggested.

PR #812, opened 2022-06-29
PWX-24550: Enables 'deep' inspection to validate Device Path

Pure-AdamuKaapan:
It seems without this flag we use some sort of cached value. `pxctl v i` worked fine, but the SDK was not.


dsupure (approved):
LGTM!

PR #815, opened 2022-07-01
Add checks for readonly path

dahuang-purestorage:

We were seeing cases where the readonly path was not set correctly, where both the device path and mount path are set to RO. This PR adds checks to make sure we detect it before proceeding further.

Closes #
PWX-24555

Testing note: tested locally with ginkgo and passed:
`ginkgo --focus Sharedv4Functional -v bin/basic.test -- -spec-dir `pwd`/drivers/scheduler/k8s/specs --app-list test-sharedv4`

### DIFF
--- a/tests/basic/sharedv4_test.go
+++ b/tests/basic/sharedv4_test.go
@@ -29,6 +29,8 @@ var _ = Describe("{Sharedv4Functional}", func() {
 	var workers []node.Node
 	var numPods int
 	var namespacePrefix string
+	var volumeMountRW = regexp.MustCompile(`,rw,|,rw|rw,|rw`)
+	var volumeMountRO = regexp.MustCompile(`,ro,|,ro|ro,|ro`)
 
 	JustBeforeEach(func() {
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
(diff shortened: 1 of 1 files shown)


tests/basic/sharedv4_test.go
```diff
@@ -113,6 +115,17 @@ var _ = Describe("{Sharedv4Functional}", func() {
...
+
+					// check only mnt.VfsOpts (fs state) is RO
+					mntList, err := mount.GetMounts()
+					Expect(err).NotTo(HaveOccurred())
+					for _, mnt := range mntList {
+						if mnt.Mountpoint != devicePath {
+							continue
+						}
+						Expect(volumeMountRW.MatchString(mnt.Opts)).To(BeTrue())
+						Expect(volumeMountRO.MatchString(mnt.VfsOpts)).To(BeTrue())
```
pureneelesh, line 127: This looks fine. If the same problem happens again, the test will fail earlier which is good. But the test will still fail, yes? How do we make the test not fail?

pureneelesh (approved):

tests/basic/sharedv4_test.go
dahuang-purestorage: yeah, we are still seeing the same failure in the latest test. I think we might need a more robust way to make sure the mount is set to the expected permissions. I'll add a future task for this.

PR #820, opened 2022-07-13
PWX-24489 Adds vault secrets login test 

dahuang-purestorage:
Adds vault secret login test for testing AppRole authentication

PWX-24489

Test run and passed at: https://jenkins.pwx.dev.purestorage.com/job/Users/job/Daniel/job/vault-test/14/console


tests/secrets/secrets_vault_test.go
```diff
@@ -0,0 +1,106 @@
+package tests
```
harsh-px, line 1: Let's move this to https://github.com/portworx/torpedo/blob/master/tests/basic/basic_test.go. We currently have a problem in torpedo where a new testsuite will increase the size of image and the build time. Until that's fixed, we will try to avoid adding new testsuite folders.

harsh-px (changes requested):

tests/basic/basic_test.go
```diff
@@ -572,6 +574,70 @@ var _ = Describe("{CordonStorageNodesDeployDestroy}", func() {
...
+		secretsProvider       = "vault"
+		portworxContainerName = "portworx"
+	)
+
+	BeforeEach(func() {
+		runID = testrailuttils.AddRunsToMilestone(testrailID)
+		isOpBased, _ := Inst().V.IsOperatorBasedInstall()
+		if !isOpBased {
+			k8sApps := apps.Instance()
+			deployments, err := k8sApps.ListDaemonSets("kube-system", metav1.ListOptions{
```
pureneelesh, line 591: nit: we are listing daemonSets not deployments, the variable name should match that

tests/basic/basic_test.go
```diff
@@ -572,6 +574,70 @@ var _ = Describe("{CordonStorageNodesDeployDestroy}", func() {
...
+
+		It("has to runs secrets login for vault", func() {
+			contexts = make([]*scheduler.Context, 0)
+			n := node.GetWorkerNodes()[0]
+			err := Inst().V.RunSecretsLogin(n, secretsProvider)
+			Expect(err).ToNot(HaveOccurred())
+
+		})
+		JustAfterEach(func() {
+			AfterEachTest(contexts, testrailID, runID)
```
pureneelesh, line 632: AfterEachTest is called in the AfterEach() block as well. Probably we should call it only once per "It" test.

tests/basic/basic_test.go
```diff
@@ -572,6 +574,70 @@ var _ = Describe("{CordonStorageNodesDeployDestroy}", func() {
...
+			}
+		}
+	})
+
+	var _ = Describe("{RunSecretsLogin}", func() {
+		var testrailID = 82774
+		// testrailID corresponds to: https://portworx.testrail.net/index.php?/cases/view/82774
+		var runID int
+
+		It("has to runs secrets login for vault", func() {
```
pureneelesh, line 624: nit: runs -> run

tests/basic/basic_test.go
```diff
@@ -572,6 +574,70 @@ var _ = Describe("{CordonStorageNodesDeployDestroy}", func() {
...
+			contexts = make([]*scheduler.Context, 0)
+			n := node.GetWorkerNodes()[0]
+			err := Inst().V.RunSecretsLogin(n, secretsProvider)
```
pureneelesh, line 627: Relying on just the return code of pxctl command seems a bit inconclusive. Is there a command that we could run that would fail without the secrets login and should pass after the secrets login?

Also, please update the testcase in testrail with the steps, expected result etc. These may not be obvious to someone who has not worked on this feature.

dahuang-purestorage: The command i was looking for was actually setting cluster-wide secrets. However,  libopenstorage  API currently doesn't support it on the server side and using `pxctl` is not feasible if we need to overwrite (requires user input). We do run torpedo tests and apps using encrypted volumes, which essentially requires px to be logged in and read from vault. In that sense, i think it should cover the happy path. 

testrails is updated. thanks for the catch.

tests/basic/basic_test.go
dahuang-purestorage: updated.

tests/basic/basic_test.go
dahuang-purestorage: updated.

tests/basic/basic_test.go
dahuang-purestorage: fixed.

pureneelesh (approved):

PR #831, opened 2022-07-30
PB-2925: BackupInspect, delete and restore create object to include backupUID along with name

diptiranjanpx:
Signed-Off-By: Diptiranjan

As the proto for backupinspect , backupdelete and restorecreate requests included UID as mandatory param, 
the tests also need to be modified accordingily.

Closes # PB-2925

https://jenkins.pwx.dev.purestorage.com/job/PX-Backup/job/px-backup-integration/135/
Backup integration test successful run which vendored torpedo functions.


drivers/backup/portworx/portworx.go
```diff
@@ -1040,6 +1061,31 @@ func (p *portworx) UpdateOwnershipRule(ctx context.Context, req *api.RuleOwnersh
...
 
+func (p *portworx) GetBackupUID(ctx context.Context, orgID, backupName string) (string, error) {
+	var backupUID string
+	var totalBackups int
+	bkpEnumerateReq := &api.BackupEnumerateRequest{OrgId: orgID}
+	bkpEnumerateReq.EnumerateOptions = &api.EnumerateOptions{MaxObjects: uint64(enumerateBatchSize), ObjectIndex: 0}
+	for {
+		enumerateRsp, err := p.EnumerateBackup(ctx, bkpEnumerateReq)
+		if err != nil {
+			return backupUID, fmt.Errorf("Failed to enumerate backups for org %s ctx: [%v]", orgID, err)
```
Rohit-PX, line 1072:
```suggestion
			return "", fmt.Errorf("Failed to enumerate backups for org %s ctx: [%v]", orgID, err)
```

drivers/backup/portworx/portworx.go
```diff
@@ -1040,6 +1061,31 @@ func (p *portworx) UpdateOwnershipRule(ctx context.Context, req *api.RuleOwnersh
...
+		}
+		for _, backup := range enumerateRsp.GetBackups() {
+			if backup.GetName() == backupName {
+				return backup.GetUid(), nil
+			}
+			totalBackups++
+		}
+		if uint64(totalBackups) >= enumerateRsp.GetTotalCount() {
+			break
```
Rohit-PX, line 1081: there could be a case where the condition on line:1075 never gets satisfied in this case, we might return backupUID as blank. Is this desired behaviour?

diptiranjanpx: Yes, it is like a backup is not there and we are trying to find it.

Now if the UID is an empty string, backup inspect will fail with 
"backup object %s UID needs to be provided for inspect".

Previous to the UID fix, when only name was getting passed then also the inspect would have failed but with a different error. 

IMO the function returning empty string is fine, now its upto the caller if it wants to use this empty string or not for further calls.

drivers/backup/portworx/portworx.go
diptiranjanpx: The default string value declared with var is anyway "".

drivers/backup/portworx/portworx.go
```diff
@@ -1040,6 +1061,31 @@ func (p *portworx) UpdateOwnershipRule(ctx context.Context, req *api.RuleOwnersh
...
+		if err != nil {
+			return backupUID, fmt.Errorf("Failed to enumerate backups for org %s ctx: [%v]", orgID, err)
+		}
+		for _, backup := range enumerateRsp.GetBackups() {
+			if backup.GetName() == backupName {
+				return backup.GetUid(), nil
+			}
+			totalBackups++
+		}
+		if uint64(totalBackups) >= enumerateRsp.GetTotalCount() {
```
lalat-das, line 1080: Just curious, 
Can this check in _line no 1080_ be ever greater than the TotalCount? 
I can understand totalBackups being  less than or it can be equal to GetTotalCount()..

lalat-das (approved):
lgtm

drivers/backup/portworx/portworx.go
diptiranjanpx: The idea is to get all backups as long as uint64(totalBackups) < enumerateRsp.GetTotalCount().
Now in ideal case the breaking condition could have been uint64(totalBackups) == enumerateRsp.GetTotalCount().
But there can be a case where enumerate is happening and meanwhile some delete happened , in that case the 
totalBackups count (as this is being handled by client) can be bigger than enumerateRsp.GetTotalCount() (which is always the fresh counts).

Rohit-PX (approved):
LGTM.

PR #841, opened 2022-08-12
Reduce logs in torpedo runs

Rohit-PX:
Signed-off-by: Rohit-PX <rkulkarni@purestorage.com>

Reduce unnecessary logs in torpedo debug mode. Logs include the following:
* List of spec files parsed, this gets printed for every test and for every app
* Printing of the entire PX object at the beginning of the test
* Logs every time an app is registered

All these statements are being moved to the `Tracef` mode so that even if required we can run the job in trace mode if some debugging is required.

Results are as follows when running with just `mysql` app and `SetupTeardown` test:
* Before log file size: 1.3M
* After log file size: 60K 
Attaching both files in the PR

Advantages:
* This will make debugging torpedo easier
* It will reduce the size of the files on the jenkins server

Closes #

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -344,12 +344,12 @@ func (k *K8s) RefreshNodeRegistry() error {
 
 // ParseSpecs parses the application spec file
 func (k *K8s) ParseSpecs(specDir, storageProvisioner string) ([]interface{}, error) {
-	logrus.Infof("ParseSpecs k.CustomConfig = %v", k.customConfig)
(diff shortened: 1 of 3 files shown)


Rohit-PX:
[torpedoNew.log](https://github.com/portworx/torpedo/files/9329936/torpedoNew.log)
[torpedoOld.log](https://github.com/portworx/torpedo/files/9329937/torpedoOld.log)

mborodin-px (approved):
LGTM

jainsmit (approved):
LGTM !!

Santhoshairody:
LGTM

PR #843, opened 2022-08-16
Fixes torpedo checking for FA NVMe paths

Pure-AdamuKaapan:
Also addresses a minor annoyance if the cloudsnap credentials already exist (comes up running torpedo on my own cluster a lot)

### DIFF
--- a/tests/pure_directaccess/pure_test.go
+++ b/tests/pure_directaccess/pure_test.go
@@ -3,6 +3,7 @@ package puredirectaccess
 import (
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/onsi/ginkgo/reporters"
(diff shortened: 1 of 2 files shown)


tests/pure_directaccess/pure_test.go
```diff
@@ -62,6 +63,10 @@ func createCloudsnapCredential() {
 		TimeBeforeRetry: k8s.DefaultRetryInterval,
 		Sudo:            true,
 	})
+	// if the cloudsnap credentials already exist, just leave them there
+	if err != nil && !strings.Contains(err.Error(), "already exist") {
```
dsupure, line 67:
```suggestion
	if err != nil && !strings.Contains(err.Error(), "cloudsnap credentials already exist") {
```

david-enli (approved):
LGTM!

dsupure (approved):
LGTM!

tests/pure_directaccess/pure_test.go
Pure-AdamuKaapan: I wanted to leave the error message broad in case that message happened to change. I figured "already exist" would be a good catch-all

PR #855, opened 2022-08-26
PWX-26528: C58509 Test cloning of FA volume from an existing FA volume 

david-enli:
clone multiple PVCs from a single PVC and validate (C58509).

Closes #

**FADA/FBDA test results**:
* FADA: https://jenkins.pwx.dev.purestorage.com/job/Pure-FAFB/job/CI/job/FADA-FlashArray-DirectAccess/445/
* FBDA: https://jenkins.pwx.dev.purestorage.com/job/Pure-FAFB/job/CI/job/FBDA-FlashBlade-DirectAccess/412/

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -52,6 +52,7 @@ import (
 	"github.com/portworx/torpedo/drivers/volume"
 	"github.com/portworx/torpedo/pkg/aututils"
 	"github.com/portworx/torpedo/pkg/errors"
+	"github.com/portworx/torpedo/pkg/pureutils"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
 	appsapi "k8s.io/api/apps/v1"
@@ -658,6 +659,15 @@ func validateSpec(in interface{}) (interface{}, error) {
 // volumes are valid.
 func (k *K8s) filterPureVolumesIfEnabled(claim *v1.PersistentVolumeClaim) (bool, error) {
 	volTypes := []string{PureFile, PureBlock}
+
+	return k.filterPureTypeVolumeIfEnabled(claim, volTypes)
+}
+
+// filterPureVolumesIfEnabled will return whether or not the given PVC is a Pure
+// direct access volume, only if PureVolumes is true. If PureVolumes is false, all
+// volumes are valid.
+func (k *K8s) filterPureVolumesIfEnabledByPureVolBackend(claim *v1.PersistentVolumeClaim, pureVolBackend string) (bool, error) {
+	volTypes := []string{pureVolBackend}
 
 	return k.filterPureTypeVolumeIfEnabled(claim, volTypes)
 }
@@ -3182,6 +3192,71 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 	return vols, nil
 }
 
+// GetPureVolumes  Get the Pure volumes (if enabled) by type (PureFile or PureBlock)
+//
+func (k *K8s) GetPureVolumes(ctx *scheduler.Context, pureVolType string) ([]*volume.Volume, error) {
+	k8sOps := k8sApps
+	var vols []*volume.Volume
+	for _, specObj := range ctx.App.SpecList {
+		if obj, ok := specObj.(*corev1.PersistentVolumeClaim); ok {
+			pvcObj, err := k8sCore.GetPersistentVolumeClaim(obj.Name, obj.Namespace)
+			if err != nil {
+				return nil, err
+			}
+			shouldAdd, err := k.filterPureVolumesIfEnabledByPureVolBackend(pvcObj, pureVolType)
+			if err != nil {
+				return nil, err
+			}
+			if !shouldAdd {
+				continue
+			}
+
+			pvcSizeObj := pvcObj.Spec.Resources.Requests[corev1.ResourceStorage]
+			pvcSize, _ := pvcSizeObj.AsInt64()
+			vol := &volume.Volume{
+				ID:          string(pvcObj.Spec.VolumeName),
+				Name:        obj.Name,
+				Namespace:   obj.Namespace,
+				Shared:      k.isPVCShared(obj),
+				Annotations: make(map[string]string),
+				Labels:      pvcObj.Labels,
+				Size:        uint64(pvcSize),
+			}
+			for key, val := range obj.Annotations {
+				vol.Annotations[key] = val
+			}
+			vols = append(vols, vol)
+		} else if obj, ok := specObj.(*appsapi.StatefulSet); ok {
+			ss, err := k8sOps.GetStatefulSet(obj.Name, obj.Namespace)
+			if err != nil {
+				return nil, &scheduler.ErrFailedToGetStorage{
+					App:   ctx.App,
+					Cause: fmt.Sprintf("Failed to get StatefulSet: %v. Err: %v", obj.Name, err),
+				}
+			}
+
+			pvcList, err := k8sOps.GetPVCsForStatefulSet(ss)
+			if err != nil || pvcList == nil {
+				return nil, &scheduler.ErrFailedToGetStorage{
+					App:   ctx.App,
+					Cause: fmt.Sprintf("Failed to get PVC from StatefulSet: %v. Err: %v", ss.Name, err),
+				}
+			}
+
+			for _, pvc := range pvcList.Items {
+				vols = append(vols, &volume.Volume{
+					ID:        pvc.Spec.VolumeName,
+					Name:      pvc.Name,
+					Namespace: pvc.Namespace,
+					Shared:    k.isPVCShared(&pvc),
+				})
+			}
+		}
+	}
+
+	return vols, nil
+}
+
 // ResizeVolume  Resize the volume
 func (k *K8s) ResizeVolume(ctx *scheduler.Context, configMapName string) ([]*volume.Volume, error) {
 	var vols []*volume.Volume
@@ -5055,55 +5130,43 @@ func (k *K8s) CreateCsiSnapsForVolumes(ctx *scheduler.Context, snapClass string)
 	return volSnapMap, nil
 }
 
-// CreateCsiSnapsAndValidate create CSI snapshot for Pure FA volumes and restore them to new PVCs, and validate the content
+// CSISnapshotTest create CSI snapshot for volumes and restore them to new PVCs, and validate the content
 // (Testrail cases C58775, 58091)
-func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string) error {
-	// This test will validate the content of the snapshot as opposed to just verify creation of snapshot.
-	timestamp := strings.Split(snapClass, "-")[0]
-	namespace := "fada-basic-snapshot-and-restore-test-" + timestamp
-	clientset, err := getKubeClient("")
+func (k *K8s) CSISnapshotTest(ctx *scheduler.Context, request scheduler.CSISnapshotRequest) error {
+	// This test will validate the content of the volume as opposed to just verify creation of volume.
+
+	pvcObj, err := k8sCore.GetPersistentVolumeClaim(request.OriginalPVCName, request.Namespace)
+	size := pvcObj.Spec.Resources.Requests[corev1.ResourceStorage]
+
 	if err != nil {
-		logrus.Errorf("Failed to get kube client for FADA snapshot and restore test: %s", err)
+		logrus.Errorf("Failed to retrieve PVC %s in namespace: %s : %s", request.OriginalPVCName, request.Namespace, err)
 		return err
 	}
-
-	// create a namespace for snapshot and restore test for FADA
-	nsSpec := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
-	_, err = k8sCore.CreateNamespace(nsSpec)
+	originalStorageClass, err := k8sCore.GetStorageClassForPVC(pvcObj)
 	if err != nil {
-		logrus.Errorf("Failed to create new namespace %s FADA snapshot and restore test: %s", namespace, err)
+		logrus.Errorf("Failed to retrieve SC for PVC %s in namespace: %s : %s", request.OriginalPVCName, request.Namespace, err)
 		return err
 	}
+	storageClassName := originalStorageClass.Name
+	logrus.Infof("Procedding with SC %s", storageClassName)
 
-
-	originalPVCName := "fada-basic-pvc"
-	snapName := "fada-basic-pvc-snapshot" + timestamp
-	restoredPVCName := "fada-basic-pvc-restored" + timestamp
+	podsUsingPVC, err := k8sCore.GetPodsUsingPVC(pvcObj.GetName(), pvcObj.GetNamespace())
+	if err != nil {
+		logrus.Errorf("Failed to retrieve pods using PVC %s/%s", pvcObj.GetName(), pvcObj.GetNamespace())
+		return err
+	}
+	pod := podsUsingPVC[0]
+	mountPath, _ := pureutils.GetAppDataDir(podsUsingPVC[0].Namespace)
 	dirtyData := "thisIsJustSomeRandomText"
-	elasticsearchSC := "elasticsearch-sc"
-
-	logrus.Infof("Create original PVC: %s", originalPVCName)
-	originalPVCSpec := MakePVC(namespace, originalPVCName, elasticsearchSC)
-	originalPVC, err := k8sCore.CreatePersistentVolumeClaim(originalPVCSpec)
-
-	originalPodSpec := MakePod(originalPVCSpec.Namespace, []*v1.PersistentVolumeClaim{originalPVC}, "ls", false)
-
-	originalPod, err := clientset.CoreV1().Pods(namespace).Create(context.TODO(), originalPodSpec, metav1.CreateOptions{})
-	if err != nil {
-		logrus.Errorf("Failed to create original pod: err")
-		return err
-	}
-
-	if err = k.waitForPodToBeReady(originalPod.Name, originalPod.Namespace); err != nil {
-		return &scheduler.ErrFailedToSchedulePod{
-			App: nil,
-			Cause:   fmt.Sprintf("Original pod is not ready. Error: %v", err),
-		}
-	}
-	fmt.Println("Original pod up and running, creating multiple snapshots and verify")
+	snapName := request.SnapName
 	for i :=0; i < 3; i++ {
 		data := fmt.Sprint(dirtyData, strconv.Itoa(int(time.Now().Unix())))
-		err = k.createSnapshotAndVerify(data, fmt.Sprint(snapName, i), namespace, snapClass, fmt.Sprint(restoredPVCName, i), originalPod.Name, originalPVC.Name)
+		err = k.writeDataToPod(data, pod.GetName(), pod.GetNamespace(), mountPath)
+		if err != nil {
+			logrus.Errorf("failed to write data to restored PVC: %s", err)
+			return err
+		}
+		err = k.snapshotAndVerify(size, data, fmt.Sprint(snapName, i), pod.GetNamespace(), storageClassName, request.SnapshotclassName, fmt.Sprint(request.RestoredPVCName, i), request.OriginalPVCName)
 		if err != nil {
 			logrus.Errorf("failed to validate restored PVC content: %s ", err)
 			return err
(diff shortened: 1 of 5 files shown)


drivers/scheduler/scheduler.go
```diff
@@ -424,3 +428,25 @@ func Get(name string) (Driver, error) {
...
+// CSISnapshotRequest contains the necessary info to create a CSI snapshot for validation purpose
+type CSISnapshotRequest struct {
+	Timestamp			string
+	Namespace			string
+	OriginalPVCName 	string
+	SnapName 			string
+	RestoredPVCName 	string
+	StorageclassName	string
+	SnapshotclassName	string
+	FADA				bool
```
Pure-AdamuKaapan, line 441: Can we call this `IsFlashArrayDirectAccess`? `FADA` might not make sense to people outside our team.

Also, the field formatting looks a little wonky here, might just be github but would be worth running `go fmt ./drivers/...`

drivers/scheduler/scheduler.go
```diff
@@ -424,3 +428,25 @@ func Get(name string) (Driver, error) {
...
+}
+
+// CSICloneRequest contains the necessary info to clone from an existing CSI volume
+type CSICloneRequest struct {
+	Timestamp			string
+	Namespace			string
+	OriginalPVCName 	string
+	RestoredPVCName 	string
+	StorageclassName	string
+	FADA				bool
```
Pure-AdamuKaapan, line 451: Same here.

drivers/scheduler/k8s/k8s.go
```diff
@@ -5055,35 +5055,56 @@ func (k *K8s) CreateCsiSnapsForVolumes(ctx *scheduler.Context, snapClass string)
...
+	namespace := request.Namespace + timestamp
 	clientset, err := getKubeClient("")
 	if err != nil {
-		logrus.Errorf("Failed to get kube client for FADA snapshot and restore test: %s", err)
+		logrus.Errorf("Failed to get kube client for snapshot and restore test: %s", err)
 		return err
 	}
+	storageClassName := request.StorageclassName
+	// for FADA test, find a FA SC on our own that has pure_block param, if no FA SC, skip the test
+	if request.FADA {
```
Pure-AdamuKaapan, line 5071: There may be multiple FADA storage classes installed if we run multiple apps: are we not able to use the PVC to get the PV and then use the storage class name from there?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
 	return nil
 }
 
-func (k *K8s) createSnapshotAndVerify(data, snapName, namespace, snapClass, restoredPVCName, originalPod, originalPVC string) error {
-	elasticsearchSC := "elasticsearch-sc"
+// CloneFAVolumeAndValidate create new PVC by cloning an existing PVC and make sure the contents of volume are same
+// (Testrail cases C58509)
+func (k *K8s) CloneFAVolumeAndValidate(ctx *scheduler.Context, request scheduler.CSICloneRequest)  error {
+	// This test will validate the content of the snapshot as opposed to just verify creation of snapshot.
```
Pure-AdamuKaapan, line 5140:
```suggestion
	// This test will validate the content of the volume as opposed to just verify creation of volume.
```

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
...
+	timestamp := request.Timestamp
+	namespace := request.Namespace + timestamp
+	clientset, err := getKubeClient("")
+	if err != nil {
+		logrus.Errorf("Failed to get kube client for clone test: %s", err)
+		return err
+	}
+
+	storageClassName := request.StorageclassName
```
Pure-AdamuKaapan, line 5149: Same here, can we use the PVC to get the PV to get this?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
...
+			logrus.Warn("No FADA SC on the cluster, skipping the test")
+			return nil
+		}
+		storageClassName = fadaStorageClassName
+	}
+	logrus.Infof("Procedding with SC %s", storageClassName)
+
+	// create a namespace for snapshot and restore test
+	nsSpec := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+	_, err = k8sCore.CreateNamespace(nsSpec)
```
Pure-AdamuKaapan, line 5175: Any reason we don't want to use the existing namespace? Only reason I ask is because I know torpedo adds some special labels to the namespace to clean up, so making more might cause issues.

drivers/scheduler/k8s/k8s.go
```diff
@@ -5177,6 +5276,62 @@ func (k *K8s) createSnapshotAndVerify(data, snapName, namespace, snapClass, rest
 	return nil
 }
 
+func (k *K8s) cloneAndVerify(data, namespace, storageClass, clonedPVCName, originalPod, originalPVC string) error {
```
Pure-AdamuKaapan, line 5279: Can you add a method header describing exactly what this does? I can figure it out by reading through but it takes a lot longer.

drivers/scheduler/k8s/k8s.go
```diff
@@ -5177,6 +5276,62 @@ func (k *K8s) createSnapshotAndVerify(data, snapName, namespace, snapClass, rest
...
+		logrus.Errorf("Failed to write data to pod: %s", err)
+		return err
+	}
+
+	// Sync the data, wait 10 secs and then proceed to snapshot the volume
+	cmdArgs3 := []string{"exec", "-it", originalPod, "-n", namespace, "--", "/bin/sync"}
+	command3 := exec.Command("kubectl", cmdArgs3...)
+	_ , err = command3.Output()
+	if err != nil {
+		logrus.Errorf("Failed to sync: err")
```
Pure-AdamuKaapan, line 5298:
```suggestion
		logrus.Errorf("Failed to sync: %s", err)
```

drivers/scheduler/k8s/k8s.go
```diff
@@ -5177,6 +5276,62 @@ func (k *K8s) createSnapshotAndVerify(data, snapName, namespace, snapClass, rest
 	return nil
 }
 
+func (k *K8s) cloneAndVerify(data, namespace, storageClass, clonedPVCName, originalPod, originalPVC string) error {
+	clientset, err := getKubeClient("")
+	if err != nil {
+		logrus.Errorf("Failed to get kube client for clone test: %s", err)
+		return err
+	}
+	cmdArgs2 := []string{"exec", "-it", originalPod, "-n", namespace, "--", "/bin/sh", "-c", fmt.Sprintf("echo -n %s >>  /mnt/volume1/aaaa.txt", data)}
```
Pure-AdamuKaapan, line 5285: Can we move the original data write out into a different method and call it before this? That way this method only does the "cloning and verifying", and the setup is done beforehand.
(This also opens us up for running multiple clones in parallel in this test)

Pure-AdamuKaapan (changes requested):

drivers/scheduler/k8s/k8s.go
```diff
@@ -5055,35 +5055,56 @@ func (k *K8s) CreateCsiSnapsForVolumes(ctx *scheduler.Context, snapClass string)
...
+		}
+		storageClassName = fadaStorageClassName
+	}
+	logrus.Infof("Procedding with SC %s", storageClassName)
+	// create a namespace for snapshot and restore test
 	nsSpec := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
 	_, err = k8sCore.CreateNamespace(nsSpec)
 	if err != nil {
-		logrus.Errorf("Failed to create new namespace %s FADA snapshot and restore test: %s", namespace, err)
+		logrus.Errorf("Failed to create new namespace %s snapshot and restore test: %s", namespace, err)
```
dsupure, line 5096:
```suggestion
		logrus.Errorf("Failed to create new namespace %s for the CSICloneTest: %s", namespace, err)
```

tests/common.go
```diff
@@ -815,18 +827,45 @@ func ValidatePureVolumeStatisticsDynamicUpdate(ctx *scheduler.Context, errChan .
 func ValidatePureVolumeSnapshotAndRestore(ctx *scheduler.Context, errChan ...*chan error) {
```
dsupure, line 827: Can we change this function to `ValidateCSISnapshotAndRestore`?

drivers/scheduler/k8s/k8s.go
dsupure: I don't understand why you need `request.FADA`? Why can't we just use `request.storageClassName` you pass in this function?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
 	return nil
 }
 
-func (k *K8s) createSnapshotAndVerify(data, snapName, namespace, snapClass, restoredPVCName, originalPod, originalPVC string) error {
-	elasticsearchSC := "elasticsearch-sc"
+// CloneFAVolumeAndValidate create new PVC by cloning an existing PVC and make sure the contents of volume are same
+// (Testrail cases C58509)
+func (k *K8s) CloneFAVolumeAndValidate(ctx *scheduler.Context, request scheduler.CSICloneRequest)  error {
```
dsupure, line 5139: Can we change the function name to be `CSICloneTest`?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5055,35 +5055,56 @@ func (k *K8s) CreateCsiSnapsForVolumes(ctx *scheduler.Context, snapClass string)
 	return volSnapMap, nil
 }
 
-// CreateCsiSnapsAndValidate create CSI snapshot for Pure FA volumes and restore them to new PVCs, and validate the content
+// CreateCsiSnapsAndValidate create CSI snapshot for volumes and restore them to new PVCs, and validate the content
 // (Testrail cases C58775, 58091)
-func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string) error {
+func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, request scheduler.CSISnapshotRequest) error {
```
dsupure, line 5060: Can we change the function name to be `CSISnapshotTest`?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
...
+	namespace := request.Namespace + timestamp
+	clientset, err := getKubeClient("")
+	if err != nil {
+		logrus.Errorf("Failed to get kube client for clone test: %s", err)
+		return err
+	}
+
+	storageClassName := request.StorageclassName
+	// for FADA test, find a FA SC on our own that has pure_block param, if no FA SC, skip the test
+	if request.FADA {
```
dsupure, line 5151: If you really need this. Make this code snippet as a function. Prevent to repeat it twice.

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
...
+		if err != nil {
+			logrus.Errorf("failed to validate cloned PVC content: %s ", err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (k *K8s) createSnapshotAndVerify(data, snapName, namespace, storageClass, snapClass, restoredPVCName, originalPod, originalPVC string) error {
```
dsupure, line 5216: To be analogy with cloneAndVerify function, change the function name to be `snapshotAndVerify`

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
...
 	clientset, err := getKubeClient("")
 	if err != nil {
-		logrus.Errorf("Failed to get kube client for FADA snapshot and restore test: %s", err)
+		logrus.Errorf("Failed to get kube client for snapshot and restore test: %s", err)
```
dsupure, line 5219:
```suggestion
		logrus.Errorf("Failed to get kube client: %s", err)
```

drivers/scheduler/k8s/k8s.go
dsupure: Agreed.

drivers/scheduler/scheduler.go
```diff
@@ -424,3 +428,25 @@ func Get(name string) (Driver, error) {
 		Type: "Scheduler",
 	}
 }
+
+// CSISnapshotRequest contains the necessary info to create a CSI snapshot for validation purpose
+type CSISnapshotRequest struct {
```
dsupure, line 433: Please align all the fields.

tests/common.go
```diff
@@ -815,18 +827,45 @@ func ValidatePureVolumeStatisticsDynamicUpdate(ctx *scheduler.Context, errChan .
...
+		}
+
+		err := Inst().S.CreateCsiSnapsAndValidate(ctx, request)
 		processError(err, errChan...)
 
 	})
 }
 
+// ValidatePureVolumeClone is the ginkgo spec for cloning a volume and verifying the content
+func ValidatePureVolumeClone(ctx *scheduler.Context, errChan ...*chan error) {
```
dsupure, line 854: Can we remove the PureVolume terms to avoid confusion. What you are doing here is majorly doing CSI function (clone and snapshot) validation. what storage we are using doesn't matter.

drivers/scheduler/k8s/k8s.go
```diff
@@ -5113,11 +5134,89 @@ func (k *K8s) CreateCsiSnapsAndValidate(ctx *scheduler.Context, snapClass string
 	return nil
 }
 
-func (k *K8s) createSnapshotAndVerify(data, snapName, namespace, snapClass, restoredPVCName, originalPod, originalPVC string) error {
-	elasticsearchSC := "elasticsearch-sc"
+// CloneFAVolumeAndValidate create new PVC by cloning an existing PVC and make sure the contents of volume are same
```
dsupure, line 5137:
```suggestion
// CSICloneTest create new PVC by cloning an existing PVC and make sure the contents of volume are same
```

tests/common.go
```diff
@@ -815,18 +827,45 @@ func ValidatePureVolumeStatisticsDynamicUpdate(ctx *scheduler.Context, errChan .
...
+	context("For validation of an cloning", func() {
+		request := scheduler.CSICloneRequest{
+			Timestamp: strconv.Itoa(int(time.Now().Unix())),
+			Namespace: "fada-basic-clone-test-",
+			OriginalPVCName: "fada-basic-clone-pvc",
+			RestoredPVCName: "fada-basic-pvc-restored",
+			StorageclassName: "elasticsearch-sc",
+			FADA: true,
+		}
+		err := Inst().S.CloneFAVolumeAndValidate(ctx, request)
```
dsupure, line 864:
```suggestion
		err := Inst().S.CSICloneTest(ctx, request)
```

tests/common.go
dsupure, line 854:
```suggestion
func ValidateCSIClone(ctx *scheduler.Context, errChan ...*chan error) {
```

drivers/scheduler/k8s/k8s.go
david-enli: @Pure-AdamuKaapan the test is not associated with any app. Its just like our PSO app, we are using a single pod for this test. Because other apps may have data written to it already and I want this volume to contain only the dirty data for ease of comparison. So it just spins up a simple pod for verification purpose. 
@dsupure since we might be supporting FB snapshot/cloning later but now we aren't. But currently this test is also triggered for FB causing FBDA torpedo failure yesterday. with this parameter, we go in to find a FADA SC, and if it doesn't find it, it will skip the test. But if other people want to just try a CSI clone, they can still call this function, and set FADA to false and still be able to use it

drivers/scheduler/k8s/k8s.go
david-enli: yea, same reason as above, its not using any specific application and I think other namespaces contain the Application name like `elasticsearch-purevolumestest-0-08-29-20h45m01s`

drivers/scheduler/scheduler.go
```diff
@@ -339,8 +342,12 @@ type Driver interface {
 	// CreateCsiSnapshot create csi snapshot for given pvc
 	CreateCsiSnapshot(name string, namespace string, class string, pvc string) (*v1beta1.VolumeSnapshot, error)
 
-	// CreateCsiSnapsAndValidate create csi snapshot and return a pvc using that snapshot
-	CreateCsiSnapsAndValidate(*Context, string) error
+	// CSISnapshotTest create csi snapshot and return a pvc using that snapshot
+	CSISnapshotTest(*Context, CSISnapshotRequest) error
```
dsupure, line 346: I am OK for this PR adding the `CSISnapshotTest` and other functions you recently added to the Driver interface. But think again for the nature of Driver interface operation: `CreateCSISnapshot` is better than `CSISnapshotTest`. 
From the SW architecture point of view, you should remove the validation part in the current function. Driver can create CSI snapshot but it shouldn't do the validation.

pkg/pureutils/pure.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/portworx/torpedo/pkg/units"
+	"strings"
+)
+
+// GetPureAppDataDir checks the pod namespace prefix, and returns a path that we can
+// write data to on a volume. Because the mountpath varies so heavily between applications
+// (some have multiple PVCs, some have configmap vols, etc. etc.) this is the easiest way
+// at the moment. As more apps get added to our test suite, we should add them here.
+func GetPureAppDataDir(namespace string) (string, int) {
```
dsupure, line 12:
```suggestion
func GetAppDataPath(namespace string) (string, int) {
```
Remove Pure. It's not related.

pkg/pureutils/pure.go
```diff
@@ -0,0 +1,26 @@
...
+		return "/var/www/html", units.GiB / 2
+	}
+	if strings.HasPrefix(namespace, "elasticsearch") {
+		return "/usr/share/elasticsearch/data", units.GiB * 2
+	}
+	if strings.HasPrefix(namespace, "mysql-without-enc") {
+		return "/var/lib/mysql", units.GiB
+	}
+	return "", 0
+}
```
dsupure, line 26: nit: newline

drivers/scheduler/k8s/k8s.go
```diff
@@ -3182,6 +3192,71 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 	return vols, nil
 }
 
+// GetPureVolumes  Get the Pure volumes (if enabled) by type (PureFile or PureBlock)
```
dsupure, line 3195: Do we have a better way to get all pure volumes instead of creating a new volume list?

drivers/scheduler/k8s/k8s.go
```diff
@@ -3182,6 +3192,71 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+				Shared:      k.isPVCShared(obj),
+				Annotations: make(map[string]string),
+				Labels:      pvcObj.Labels,
+				Size:        uint64(pvcSize),
+			}
+			for key, val := range obj.Annotations {
+				vol.Annotations[key] = val
+			}
+			vols = append(vols, vol)
+		} else if obj, ok := specObj.(*appsapi.StatefulSet); ok {
```
dsupure, line 3229: Why we need to check it's a statefulset app? What if it's a pod or demonSet deployment?

drivers/scheduler/scheduler.go
david-enli: yea I agree there's probably better place to place the validation. Though functions like `ValidatePureVolumesNoReplicaSets` and `ValidateGetByteUsedForVolume` are also under the drivers package (/torpedo/drivers/volume/portworx). Our test is a little special in that its creating separate objects from the "normal" portworx operation, so its also not exactly portworx operation. I can create a separate ticket for improvement if thats needed.

dsupure: Sounds good. Let's add a TODO here.

drivers/scheduler/k8s/k8s.go
david-enli: sry, what we mean by a new volume list

drivers/scheduler/k8s/k8s.go
david-enli: as for as I know its mostly statefulset if not all? this is the same check we do for all functions of this kind actually.

dsupure: What happens if your app is a deamonset or deployment? Will this function return pure volumes for the app?

drivers/scheduler/k8s/k8s.go
david-enli: the function I added is a variation of an existing getVolumes function, as far as I know this is the only way to get all volumes, what's the concern with the list.

drivers/scheduler/k8s/k8s.go
david-enli: I'm honestly not sure, @Pure-AdamuKaapan do you know more about this?

dsupure (approved):

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: @dsupure this is only getting storage objects: the reason we have to get StatefulSets is because they have the PVC template built into them. Deployments and DaemonSets would use a separate PVC and would be caught by the first if check.

Like @david-enli said, this kind of check is present through most places we list volumes in Torpedo

drivers/scheduler/k8s/k8s.go
```diff
@@ -662,6 +663,15 @@ func (k *K8s) filterPureVolumesIfEnabled(claim *v1.PersistentVolumeClaim) (bool,
 	return k.filterPureTypeVolumeIfEnabled(claim, volTypes)
 }
 
+// filterPureVolumesIfEnabled will return whether or not the given PVC is a Pure
+// direct access volume, only if PureVolumes is true. If PureVolumes is false, all
+// volumes are valid.
+func (k *K8s) filterPureVolumesIfEnabledByType(claim *v1.PersistentVolumeClaim, pureVolType string) (bool, error) {
```
Pure-AdamuKaapan, line 669: Just to clarify, what is a "type" here? Storage class name? `backend` param value? Might want to rename it so it's a bit more clear

Pure-AdamuKaapan: Seems like it's "PureFile" or "PureBlock", can we call it `pureVolBackend` then?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5055,89 +5130,126 @@ func (k *K8s) CreateCsiSnapsForVolumes(ctx *scheduler.Context, snapClass string)
...
-	nsSpec := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
-	_, err = k8sCore.CreateNamespace(nsSpec)
+	originalStorageClass, err := k8sCore.GetStorageClassForPVC(pvcObj)
 	if err != nil {
-		logrus.Errorf("Failed to create new namespace %s FADA snapshot and restore test: %s", namespace, err)
+		logrus.Errorf("Failed to retrieve SC for PVC %s in namespace: %s : %s", request.OriginalPVCName, request.Namespace, err)
 		return err
 	}
+	storageClassName := originalStorageClass.Name
+	logrus.Infof("Procedding with SC %s", storageClassName)
```
Pure-AdamuKaapan, line 5151:
```suggestion
	logrus.Infof("Proceeding with SC %s", storageClassName)
```

drivers/scheduler/k8s/k8s.go
```diff
@@ -5055,89 +5130,126 @@ func (k *K8s) CreateCsiSnapsForVolumes(ctx *scheduler.Context, snapClass string)
...
 	dirtyData := "thisIsJustSomeRandomText"
-	elasticsearchSC := "elasticsearch-sc"
-
-	logrus.Infof("Create original PVC: %s", originalPVCName)
-	originalPVCSpec := MakePVC(namespace, originalPVCName, elasticsearchSC)
-	originalPVC, err := k8sCore.CreatePersistentVolumeClaim(originalPVCSpec)
+	snapName := request.SnapName
+	for i :=0; i < 3; i++ {
+		data := fmt.Sprint(dirtyData, strconv.Itoa(int(time.Now().Unix())))
+		err = k.writeDataToPod(data, pod.GetName(), pod.GetNamespace(), mountPath)
```
Pure-AdamuKaapan, line 5164: Are we intending to do it like:
* Write data
* Snap & verify
* Write data
* Snap & verify
* Write data
* Snap & verify
?
I kinda assumed we'd do:
* Write data
* Snap once
* Make three separate restores

drivers/scheduler/k8s/k8s.go
```diff
@@ -5270,6 +5422,21 @@ func MakePVCFromSnapshot(ns string, name string, sourceSnapshotName string, stor
 	return PVCSpec
 }
 
+// MakePVCByCloning takes namespace, name, source snapshot name and storageclass as parameter and returns a PersistentVolumeClaim spec for PVC creation
+func MakePVCByCloning(size resource.Quantity, ns string, name string, sourcePVCName string, storageClass string) *v1.PersistentVolumeClaim {
+	PVCSpec := MakePVC(size, ns, name, storageClass)
+	PVCSpec.ObjectMeta.Name = name
+	PVCSpec.Namespace = ns
+	PVCSpec.ObjectMeta.Namespace = ns
+	if sourcePVCName != "" {
```
Pure-AdamuKaapan, line 5431: We should probably just throw an error if source PVC name is empty... might help prevent some strangeness where it just ends up making a new volume instead

drivers/scheduler/k8s/k8s.go
```diff
@@ -5270,6 +5422,21 @@ func MakePVCFromSnapshot(ns string, name string, sourceSnapshotName string, stor
 	return PVCSpec
 }
 
+// MakePVCByCloning takes namespace, name, source snapshot name and storageclass as parameter and returns a PersistentVolumeClaim spec for PVC creation
+func MakePVCByCloning(size resource.Quantity, ns string, name string, sourcePVCName string, storageClass string) *v1.PersistentVolumeClaim {
```
Pure-AdamuKaapan, line 5426: Maybe call this `GeneratePVCCloneSpec`? "Make" makes it sound like it's actually running the create.

tests/common.go
```diff
@@ -811,19 +823,64 @@ func ValidatePureVolumeStatisticsDynamicUpdate(ctx *scheduler.Context, errChan .
...
 }
 
-// ValidatePureVolumeSnapshotAndRestore is the ginkgo spec for validating actually creating a FADA snapshot, restoring and verifying the content
-func ValidatePureVolumeSnapshotAndRestore(ctx *scheduler.Context, errChan ...*chan error) {
+// ValidateCSISnapshotAndRestore is the ginkgo spec for validating actually creating a FADA snapshot, restoring and verifying the content
+func ValidateCSISnapshotAndRestore(ctx *scheduler.Context, errChan ...*chan error) {
 	context("For validation of an snapshot and restoring", func() {
+		var err error
 		timestamp := strconv.Itoa(int(time.Now().Unix()))
-		snapShotClassName := PureSnapShotClass + "-" + timestamp
```
Pure-AdamuKaapan, line 818: Any specific reason for changing this? I prefer hyphens personally to be more consistent

Pure-AdamuKaapan (commented):
Left a few comments but overall looks a lot better than last time!

drivers/scheduler/k8s/k8s.go
david-enli: yea so this test is to write data snap and verify, write some more, and then verify

`I kinda assumed we'd do:
Write data
Snap once
Make three separate restores`
this is for NEWSTACK-573 and that will be a separate PR coming out later today( hopefully)

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: Bump on this

tests/common.go
Pure-AdamuKaapan: Discussed w/ En in Slack, this was to make it easier to split and get the timestamp/snapshot class name.

Pure-AdamuKaapan (approved):
LGTM, Thanks for all the hard work! 🚢

PR #865, opened 2022-09-07
:wrench: fixing longevity tests timeout

lsrinivas-pure:
Fixes min run time for longevity test

Closes #PTX-12805

### DIFF
--- a/tests/longevity/longevity_test.go
+++ b/tests/longevity/longevity_test.go
@@ -139,6 +139,9 @@ var _ = Describe("{Longevity}", func() {
 			}
 		})
 		logrus.Infof("Finished registering test triggers")
+		if Inst().MinRunTimeMins != 0 {
+			logrus.Infof("Longevity Tests  timeout set to %d  minutes", Inst().MinRunTimeMins)
+		}
 
 		Step("Register email trigger", func() {
 			for triggerType, triggerFunc := range emailTriggerFunction {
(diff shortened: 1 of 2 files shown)


tests/longevity/longevity_test.go
```diff
@@ -180,6 +183,7 @@ func testTrigger(wg *sync.WaitGroup,
 	for {
 		// if timeout is 0, run indefinitely
 		if timeout != 0 && int(time.Since(start).Seconds()) > timeout {
+			logrus.Infof("Longevity Tests timed out with timeout %d  minutes", minRunTime)
```
jainsmit, line 186: There should be more to be done here. We should come out with timeout, for an example for 1TB pool size it should not be greater than 6 hours and based on this we can calculate the timeout for rebalance. 

This would be useful if pool rebalance operation stuck forever and we should mark test fail. The issue we observed pool rebalance got stuck.

jainsmit: Actually this should monitor pool rebalance progress percentage if percentage doesn't change for long time, we should mark test failed.

lsrinivas-pure: Hi Smit,

this change is for giving user an option to timeout Longevity after running certain time as Longevity in long running test
Ill work on pool rebalnace in another PR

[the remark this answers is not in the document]
lsrinivas-pure: yes , tested in the below run
http://jen-endurance.pwx.dev.purestorage.com/job/Torpedo/job/tp-fa-longevity-pipeline/51/

jainsmit (approved):
LGTM !!

PR #867, opened 2022-09-08
PTX-12815: increase vol validation timeout when there are more volumes

pureneelesh:

We expect the PVCs to be ready within 2 min by default. We increase that timeout when the app has several instances (app scale factor is greater than 1). We need to do the same when an app is using a lot of volumes.

This patch takes into account the number of volumes when setting the timeout.

Signed-off-by: Neelesh Thakur <neelesh.thakur@purestorage.com>

PTX-12815

Verified by running the sharedv4-svc-scale test:
```
11:44:22 INFO[2022-09-08 18:44:22] Using vol scale factor of 10 for app vdbench-sv4-svc-multivol 
```

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -7,11 +7,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 	"io/ioutil"
-	storageapi "k8s.io/api/storage/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


tests/common.go
```diff
@@ -594,8 +596,22 @@ func ValidateVolumes(ctx *scheduler.Context, errChan ...*chan error) {
...
+				logrus.Errorf("Failed to get app %s's volumes", ctx.App.Key)
+				processError(err, errChan...)
+			}
+			volScaleFactor := 1
+			if len(vols) > 10 {
+				// Take into account the number of volumes in the app. More volumes will
+				// take longer to format if the backend storage has limited bandwidth. Even if the
+				// GlobalScaleFactor is 1, high number of volumes in a single app instance
+				// may slow things down.
+				volScaleFactor = len(vols) / 10
```
dahuang-purestorage, line 610: I think golang are treating them as integer by default. Is that what is expected? (e.g. only integer scale factor)
or we are expecting floating scale factor?

pureneelesh: Yes, just integer.

dahuang-purestorage (approved):
LGTM

PR #868, opened 2022-09-12
:wrench: Adding time out for pool expansion validation check

lsrinivas-pure:
Adding time out for pool expansion validation check based on the pool size

Closes #PTX-12785


tests/storage_pool/storage_pool_test.go
```diff
@@ -368,10 +369,12 @@ func poolResizeIsInProgress(poolToBeResized *api.StoragePool) bool {
 
 			updatedPoolToBeResized := pools[poolToBeResized.Uuid]
 			Expect(updatedPoolToBeResized).ShouldNot(BeNil())
+			Expect(waitCount).ShouldNot(BeEquivalentTo(0), fmt.Sprintf("timeed out waiting for pool resize to finish"))
```
jainsmit, line 372: Spelling is wrong for `timeed out`

tests/storage_pool/storage_pool_test.go
```diff
@@ -381,6 +384,21 @@ func poolResizeIsInProgress(poolToBeResized *api.StoragePool) bool {
 	return poolSizeHasBeenChanged
 }
 
+func getWaitCount(poolSize uint64) int {
```
jainsmit, line 387: This methods used in two places, can we put into common lib ?

One more thing instead of waiting for particular period of time. Can we check another way, example if rebalance percentage doesn't change for 15 minutes then we can fail the operation..

lsrinivas-pure: > This methods used in two places, can we put into common lib ?
> 
> One more thing instead of waiting for particular period of time. Can we check another way, example if rebalance percentage doesn't change for 15 minutes then we can fail the operation..

Moved this method common lib.

Reblance percnetange show only when we do for add disk not for resize disk,
As of now added log message to monitor mrssage log. 
Will monitor that for more run and take this check as enhancement for next PR

tests/storage_pool/storage_pool_test.go
lsrinivas-pure: fixed

tests/basic/storage_pool_test.go
```diff
@@ -384,6 +384,7 @@ func roundUpValue(toRound uint64) uint64 {
 
 func poolResizeIsInProgress(poolToBeResized *api.StoragePool) bool {
 	poolSizeHasBeenChanged := false
+	waitCount := 20
 	if poolToBeResized.LastOperation != nil {
 		for {
```
nikolaypopov, line 389: Just wondering why not use already existing `DoRetryWithTimeout()` function we have instead of the doing it via `waitCount`?

tests/basic/storage_pool_test.go
```diff
@@ -384,6 +384,7 @@ func roundUpValue(toRound uint64) uint64 {
 
 func poolResizeIsInProgress(poolToBeResized *api.StoragePool) bool {
 	poolSizeHasBeenChanged := false
+	waitCount := 20
```
jainsmit, line 387: Why to wait 20 minutes for  pool resize operation to be succeed ? It seems this value is high for resize operation..

jainsmit (approved):
Minor comment. LGTM !!

tests/basic/storage_pool_test.go
lsrinivas-pure: yes, it can be used, will take it as enhancemet and will update in next PR as i need time to test it

tests/basic/storage_pool_test.go
lsrinivas-pure: reduced to 5 mins

PR #877, opened 2022-09-23
PWX-27143: Determine the diags_server connection port using the confi…

jrivera-px:
…gured mgmnt port.

On OCP the default diags server port is shifted based on the PX mgmnt port of 17001.   So change code to determine the diags server port using the mgmnt port instead of hard coding to 9014. 
Closes #
PWX-27143


nikolaypopov (approved):
LGTM, please test it on the job with shifted ports before merging.

PR #880, opened 2022-09-23
PWX-27153: Need to make sure to get and use the newly generated profi…

jrivera-px:
…le diags and check s3 for those.

Needed to make sure the profile diags was not only generated before trying to get the profile names for S3 validation check.  Also need to make sure the profiles found were the right ones.   Fix the code to get the existing profiles first so that after the new diags is generated you can compare and make sure you have the latest.   

Closes #
PWX-27153


nikolaypopov (approved):
LGTM

PR #897, opened 2022-10-13
test binary changes for backup

lsrinivas-pure:
Making PX-backup binary changes
We have  segrated all PX enterprise related tests into tests/basic directory and pds tests into tests/pds .
This change will make simialr change to px-backup moving into tests/backup
This commit adheres to rule that one torpedo container will include only one test binary. For example,

portworx/torpedo:master will contain only basic.test
portworx/torpedo-pds:master will contain only pds.test
portworx/torpedo-backup:master will contain only backup.test
This will ensure that the container size will not balloon up again as people add more test directories such as pxbackup, mpxe etc.

Closes #PA-249


.travis.yml
```diff
@@ -25,6 +25,10 @@ after_success:
       sudo docker tag "${DOCKER_HUB_REPO}/torpedo-pds:${DOCKER_HUB_TAG}" "${DOCKER_HUB_REPO}/torpedo-pds:${TRAVIS_BRANCH}";
       sudo docker push "${DOCKER_HUB_REPO}/torpedo-pds:${DOCKER_HUB_TAG}";
       sudo docker push "${DOCKER_HUB_REPO}/torpedo-pds:${TRAVIS_BRANCH}";
+      make container-bkp;
```
pureneelesh, line 28: nit: better to spell out "backup" since "bkp" is not a well known acronym.

pureneelesh (approved):

.travis.yml
lsrinivas-pure: fixed

PR #900, opened 2022-10-17
PTX-13378: Fixing logic to fetch the telemetry status

aghodke1312:
https://jenkins.pwx.dev.purestorage.com/job/Torpedo/view/Operator%20Release%20Cloud%20Dashboard/job/tp-nextpx-gke-op-ms/

### DIFF
--- a/tests/basic/telemetry_test.go
+++ b/tests/basic/telemetry_test.go
@@ -55,9 +55,14 @@ func runPxctlCommand(pxctlCmd string, n node.Node, cmdConnectionOpts *node.Conne
 
 func TelemetryEnabled(currNode node.Node) bool {
 	// This returns true if telemetry is enabled
-	out, err := runPxctlCommand("status -j | jq .telemetrystatus.connection_status.error_code", currNode, nil)
-	Expect(err).NotTo(HaveOccurred(), "Error getting telemetry status for %s", currNode.Name)
-	return strings.TrimSpace(out) == TelemetryEnabledStatus
+	output, err := runPxctlCommand("status | egrep ^Telemetry:", currNode, nil)
+	Expect(err).NotTo(HaveOccurred(), "Failed to get status for nodee %v", currNode.Name)
+	logrus.Infof("node %s: %s", currNode.Name, output)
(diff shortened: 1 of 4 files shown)


tests/basic/telemetry_test.go
```diff
@@ -271,6 +273,8 @@ var _ = Describe("{ProfileOnlyDiags}", func() {
 						return
 					}
 				})
+				// This sleep is required because the heap and stack logs might not have been written at same time.
+				time.Sleep(30* time.Second)
```
nikolaypopov, line 277: I would do retry here instead of hardcoded sleep

tests/basic/telemetry_test.go
```diff
@@ -55,9 +55,14 @@ func runPxctlCommand(pxctlCmd string, n node.Node, cmdConnectionOpts *node.Conne
 
 func TelemetryEnabled(currNode node.Node) bool {
 	// This returns true if telemetry is enabled
-	out, err := runPxctlCommand("status -j | jq .telemetrystatus.connection_status.error_code", currNode, nil)
-	Expect(err).NotTo(HaveOccurred(), "Error getting telemetry status for %s", currNode.Name)
-	return strings.TrimSpace(out) == TelemetryEnabledStatus
+	output, err := runPxctlCommand("status | egrep ^Telemetry:", currNode, nil)
+	Expect(err).NotTo(HaveOccurred(), "Failed to get status for nodee %v", currNode.Name)
```
nikolaypopov, line 59: nit: nodee

tests/basic/telemetry_test.go
jrivera-px: The sleep can actually be removed.  The next step does an "Eventually" which retries the check for the newly generated stack/heap files so no need to have this sleep here at all.

jrivera-px (approved):
LGTM... just remove the sleep.   Thx

CharudathGopal (approved):

aghodke1312:
@jrivera-px Lets say if heap file is generated and .stack is not, in that case the file list will be 1 stack file from previous run and will cause the test to fail, I can keep a 10 sec sleep so that we dont run into this condition?

PR #901, opened 2022-10-17
[PA-229] Upgrade data service image patch versions

bhsrinivasan-px:
This is a PR for upgrading only the image version, keeping the DS_VERSION constant. The env variables DS_VERSION, IMAGE_TO_UPDATE, DS_BUILD will be used for this test. The test first creates a deployment with DS_VERSION/IMAGE_TO_UPDATE and then runs some load against it. Next, it upgrades ONLY the image, from IMAGE_TO_UPGRADE (old image) to the image mentioned in DS_BUILD (new image). 

How to run:
```
export DS_VERSION=14.4
export IMAGE_TO_UPDATE=4416feb
export DS_BUILD=b64d741
ginkgo --focus UpgradeDataServiceImage -v /path/to/torpedo/bin/pds.test -log-location `pwd`
```

Closes #PA-229

This change creates one new spec UpgradeDataServiceImage and modifies the existing spec UpgradeDataServiceVersion and ensures that both of these related testcases use a common runner function. The environment variables needed for the new testcase have been mentioned above.

Test Execution Logs: https://gist.github.com/bhsrinivasan-px/69b83bb6127ee098ae9cb6380613e029

### DIFF
--- a/tests/pds/dataservice_test.go
+++ b/tests/pds/dataservice_test.go
@@ -2,7 +2,6 @@ package tests
 
 import (
 	"net/http"
-	"os"
 	"strconv"
 	"testing"
 
@@ -61,8 +60,6 @@ var (
 	DeployAllImages                         bool
 	dsVersion                               string
 	dsBuild                                 string
-	oldDSImage                              string
-	oldDSVersion                            string
 	deployments                             map[string][]*pds.ModelsDeployment
 	dataServiceVersionBuildMap              map[string][]string
 	dataServiceImageMap                     map[string][]string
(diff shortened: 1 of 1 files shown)


tests/pds/dataservice_test.go
```diff
@@ -272,29 +181,154 @@ var _ = Describe("{UpgradeDataServiceVersion}", func() {
...
+			Step("Get the resource and app config template for supported dataservice", func() {
+				dataServiceDefaultResourceTemplateIDMap, dataServiceNameIDMap, err = pdslib.GetResourceTemplate(tenantID, supportedDataServices)
+				Expect(err).NotTo(HaveOccurred())
+
+				dataServiceNameDefaultAppConfigMap, err = pdslib.GetAppConfTemplate(tenantID, dataServiceNameIDMap)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(dataServiceNameDefaultAppConfigMap).NotTo(BeEmpty())
+			})
+
+		} else {
```
lsrinivas-pure, line 294: Is the validation required here? if it else condition then value must be true only

tests/pds/dataservice_test.go
```diff
@@ -136,128 +133,40 @@ var _ = BeforeSuite(func() {
...
-					}
-				}
-			}
-		})
-
-		Step("Update the data service patch versions", func() {
+func UpgradeDataService(oldVersion string, oldImage string, dsVersion string, dsBuild string) {
+
+	Step("Deploy Data Services", func() {
+		logrus.Infof("Version/Build: %v %v", oldVersion, oldImage)
```
lsrinivas-pure, line 139: can we move the deployment and deletion of services to justBeforeEach and JustAfterEach

lsrinivas-pure (changes requested):

tests/pds/dataservice_test.go
bhsrinivasan-px: there are other specs which use the same function with different variables, will check with Madan if it can be done

bhsrinivasan-px: As mentioned, the deployment ids, names etc will not be accessible in JustAfterEach for cleanup, and there will be a bit of code repetition if we move the deploy and teardown steps to JustBeforeEach, JustAfterEach

tests/pds/dataservice_test.go
```diff
@@ -470,8 +500,8 @@ var _ = Describe("{DeployAllDataServices}", func() {
 	})
 })
 
-func TestMain(m *testing.M) {
-	// call flag.Parse() here if TestMain uses flags
-	ParseFlags()
-	os.Exit(m.Run())
-}
+// func TestMain(m *testing.M) {
+// 	// call flag.Parse() here if TestMain uses flags
```
lsrinivas-pure, line 504: uncomment this

lsrinivas-pure (changes requested):

tests/pds/dataservice_test.go
bhsrinivasan-px: done

lsrinivas-pure (approved):

PR #902, opened 2022-10-17
PWX-27436: Improved Pure command logging to find errors

Pure-AdamuKaapan:
Some Pure tests were failing with just a mysterious "exit status 1", but when I went onto the node I could read and write just fine. This will grab both stdout and stderr and print them in case of an error, which should help with debugging.

**This does not fix the issue**, just makes it easier to debug.


dsupure (approved):
LGTM!

PR #907, opened 2022-10-19
Fix for nil pointer in updateoutcome when err is nil

lsrinivas-pure:
UpdateOutcome is givinng nil pointer when there is no error- This PR fixes it

Closes #


Madanagopal19:
LGTM

Madanagopal19 (approved):
change looks good for me

PR #919, opened 2022-10-27
[PA-213] Register multiple namespaces, deploy dataservices to multiple namespaces

bhsrinivasan-px:
Part of the PDS system test automation suite

Closes # PA-213

### DIFF
--- a/drivers/pds/lib/utils.go
+++ b/drivers/pds/lib/utils.go
@@ -1015,3 +1015,283 @@ func ValidateDataServiceVolumes(deployment *pds.ModelsDeployment, dataService st
 	return resourceTemp, storageOp, config, nil
 
 }
+
+func CreateK8sPDSNamespace(nname string) (*corev1.Namespace, error) {
+	ns, err := k8sCore.CreateNamespace(&corev1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   nname,
+			Labels: map[string]string{"pds.portworx.com/available": "true"},
+		},
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("could not create ns %v", nname)
+	}
+
+	return ns, nil
+
+}
+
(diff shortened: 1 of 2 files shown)


drivers/pds/lib/utils.go
```diff
@@ -245,19 +280,88 @@ func SetupPDSTest(ControlPlaneURL, ClusterType, TargetClusterName, AccountName s
...
+func ReadParams(filename string) (*Parameter, error) {
+	var jsonPara Parameter
+
+	if filename == "" {
+		filename, err = filepath.Abs(defaultParams)
+		logrus.Infof("filename %v", filename)
+		if err != nil {
+			logrus.Errorf("FilePath err %v", err)
+			return nil, err
+		}
```
nikolaypopov, line 332: Can we please not have this, if you want to have verbose error, just return verbose error
```
		if err != nil {
			return nil, fmt.Errorf("FilePath err %v", err)
		}
```
There are too many places to mention this, can you please fix it everywhere you have this

drivers/pds/lib/utils.go
```diff
@@ -245,19 +280,88 @@ func SetupPDSTest(ControlPlaneURL, ClusterType, TargetClusterName, AccountName s
...
+		err = json.Unmarshal(file, &jsonPara)
+		if err != nil {
+			logrus.Errorf("Error while unmarshalling json: %v\n", err)
+			return nil, err
+		}
+	} else {
+		cm, err := core.Instance().GetConfigMap(pdsParamsConfigmap, configmapNamespace)
+		if err != nil {
+			logrus.Errorf("Error reading config map: %v", err)
+		}
```
nikolaypopov, line 350: Do not just log error, I am pretty sure you want to return this
```
		if err != nil {
			return fmt.Errorf("Failed to get configmap [%s] in namespace [%s], Err: %v", pdsParamsConfigmap, configmapNamespace, err)
		}
```

nikolaypopov (changes requested):
Please return errors instead/or of printing them and try to add verbosity
**NOTE:** If you giving error about any object not being found or whatever, please at least add objet name and namespace so its easier to debug when this fails

PR #929, opened 2022-11-07
PTX-13848: Improve tagging logic for reporting to dashboard

CharudathGopal:
- Changed tagging logic
- Added code to push TestRail ID to Aetos
- Minor nit fixes

### DIFF
--- a/tests/basic/storage_pool_test.go
+++ b/tests/basic/storage_pool_test.go
@@ -19,7 +19,7 @@ const poolResizeTimeout = time.Minute * 90
 const retryTimeout = time.Minute
 
 var _ = Describe("{StoragePoolExpandDiskResize}", func() {
-	StartTorpedoTest("StoragePoolExpandDiskResize", "Validate storage pool expansion expansion using resize-disk option", nil)
+	StartTorpedoTest("StoragePoolExpandDiskResize", "Validate storage pool expansion expansion using resize-disk option", nil, 0)
 	var contexts []*scheduler.Context
 	stepLog := "has to schedule apps, and expand it by resizing a disk"
 	It(stepLog, func() {
@@ -103,7 +103,7 @@ var _ = Describe("{StoragePoolExpandDiskResize}", func() {
 })
 
 var _ = Describe("{StoragePoolExpandDiskAdd}", func() {
-	StartTorpedoTest("StoragePoolExpandDiskAdd", "Validate storage pool expansion expansion using add-disk option", nil)
+	StartTorpedoTest("StoragePoolExpandDiskAdd", "Validate storage pool expansion expansion using add-disk option", nil, 0)
 	var contexts []*scheduler.Context
 
 	stepLog := "should get the existing pool and expand it by adding a disk"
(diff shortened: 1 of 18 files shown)


tests/basic/storage_pool_test.go
```diff
@@ -403,18 +395,11 @@ func poolResizeIsInProgress(poolToBeResized *api.StoragePool) bool {
...
 			updatedPoolToBeResized := pools[poolToBeResized.Uuid]
-
 			dash.VerifyFatal(updatedPoolToBeResized != nil, true, "Validate pool to be resized exist")
-			dash.VerifyFatal(waitCount, 0, "timed out waiting for pool resize to finish")
 			if updatedPoolToBeResized.LastOperation.Status != api.SdkStoragePool_OPERATION_SUCCESSFUL {
 				dash.VerifyFatal(updatedPoolToBeResized.LastOperation.Status != api.SdkStoragePool_OPERATION_FAILED, true, fmt.Sprintf("PoolResize has failed. Error: %s", updatedPoolToBeResized.LastOperation))
-				err = ValidatePoolRebalance()
-				if err != nil {
-					dash.VerifyFatal(err, nil, "Error while pool rebalance")
-				}
```
mborodin-px, line 414: looks like something wrong here, as you remove some steps from the test, rebase issue?

lavong599 (approved):
Fairly simple changes.  Don't see anything that needs attention.

tests/basic/storage_pool_test.go
```diff
@@ -193,10 +191,7 @@ var _ = Describe("{StoragePoolExpandDiskAdd}", func() {
 })
 
 var _ = Describe("{PoolResizeDiskReboot}", func() {
-	JustBeforeEach(func() {
-		StartTorpedoTest("PoolResizeDiskReboot", "Initiate pool expansion using resize-disk and reboot node", nil)
-	})
-
+	StartTorpedoTest("PoolResizeDiskReboot", "Initiate pool expansion using resize-disk and reboot node", nil)
```
mborodin-px, line 194: StartTorpedoTest mehod takes 4 arguments, but you pass only 3

tests/common.go
```diff
@@ -4854,10 +4869,10 @@ func updatePxRuntimeOpts() error {
 }
 
 //StartTorpedoTest starts the logging for torpedo test
-func StartTorpedoTest(testName, testDescription string, tags []string) {
+func StartTorpedoTest(testName, testDescription string, tags map[string]string, testRepoID int) {
```
mborodin-px, line 4872: what is the testRepoId? When you invoke this method you pass `testrailID`, so maybe it is better to name it as `testrailID`?

tests/common.go
```diff
@@ -4854,10 +4869,10 @@ func updatePxRuntimeOpts() error {
...
 	TestLogger = CreateLogger(fmt.Sprintf("%s.log", testName))
 	SetTorpedoFileOutput(log, TestLogger)
-	dash.TestCaseBegin(testName, testDescription, "", tags)
+	dash.TestCaseBegin(testName, testDescription, strconv.Itoa(testRepoID), tags)
```
mborodin-px, line 4875: why do we need to make the conversion here, maybe we can pass that variable as a string?

CharudathGopal: testRailID is declared and used as int all our the place, which is correct. But Aetos API expects this to be a string. So making the conversion.

CharudathGopal:
@mborodin-px Here are the successful run with image built from the PR.

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Aniket/job/tp-basic-sharedv4/lastCompletedBuild/

lsrinivas-pure (approved):
LGTM

PR #930, opened 2022-11-07
PWX-27035: sharedv4-svc as the default for sharedv4 volumes

pureneelesh:
StorageClass needs to specify sharedv4_svc_type: "" when sharedv4 volume is desired otherwise sharedv4 service volume gets created.

Changed sharedv4 specs to specify that parameter.

Changed sharedv4-svc specs to omit that parameter (except for one to test "ClusterIP" value).

PWX-25794


lsrinivas-pure (approved):
LGTM

PR #935, opened 2022-11-10
PTX-13915:Provide logic to use the same Aetos TestSetID across Torpedo/Spawn run

CharudathGopal:

Closes #

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -8,8 +8,8 @@ if [ -z "${ENABLE_DASH}" ]; then
     ENABLE_DASH=true
 fi
 
-if [ -z "${TESTSET_ID}" ]; then
-    TESTSET_ID="0"
+if [ -z "${DASH_UID}" ]; then
+    DASH_UID="0"
 fi
 
 if [ -z "${SCALE_FACTOR}" ]; then
@@ -479,7 +479,7 @@ spec:
             "--test-desc=$TEST_DESCRIPTION",
             "--test-type=$TEST_TYPE",
             "--test-tags=$TEST_TAGS",
-            "--testset-id=$TESTSET_ID",
+            "--testset-id=$DASH_UID",
             "--branch=$BRANCH",
             "--product=$PRODUCT",
             "--torpedo-job-name=$TORPEDO_JOB_NAME",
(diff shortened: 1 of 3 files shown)


lsrinivas-pure (approved):
LGTM

deployments/deploy-ssh.sh
```diff
@@ -479,7 +479,7 @@ spec:
             "--test-desc=$TEST_DESCRIPTION",
             "--test-type=$TEST_TYPE",
             "--test-tags=$TEST_TAGS",
-            "--testset-id=$TESTSET_ID",
+            "--testset-id=$DASH_UID",
```
mborodin-px, line 482: use one variable as `aetos-uuid` and remove `enable-dash`

pkg/aetosutil/dashboardutil.go
```diff
@@ -154,7 +154,7 @@ func (d *Dashboard) TestSetBegin(testSet *TestSet) {
 				d.Log.Errorf("TestSetId creation failed. Cause : %v", err)
 			}
 			d.Log.Infof("Dashboard URL : %s", fmt.Sprintf("http://aetos.pwx.purestorage.com/resultSet/testSetID/%d", d.TestSetID))
```
mborodin-px, line 156: the Aetos host should be parameterized here

tests/common.go
```diff
@@ -4044,7 +4044,7 @@ func ParseFlags() {
 		}
 		if testsetID != 0 {
 			dash.TestSetID = testsetID
-			os.Setenv("TESTSET-ID", fmt.Sprint(testsetID))
+			os.Setenv("DASH_UID", fmt.Sprint(testsetID))
```
mborodin-px, line 4047: why do we need to set environment variable? The dash object should use it if pass `aetos-uuid` as a aparameter

lsrinivas-pure: This is added when we are using separate binaries , where it is creating different testset id for each test, this can be removed now as we are using only one binary for all the tests. 
@CharudathGopal you can raise a ticket in my name for this clean up.Will make the changes and test it

deployments/deploy-ssh.sh
CharudathGopal: we are using this as env variable and cannot use env variable with `-`.

tests/common.go
CharudathGopal: I dont see a use of it at this point, but this is added to reuse the UID when trying multiple runs in the same context.

@lsrinivas-pure are we using this anywhere as of now ?

pkg/aetosutil/dashboardutil.go
```diff
@@ -30,6 +29,7 @@ var (
 const (
 	//DashBoardBaseURL for posting logs
 	DashBoardBaseURL = "http://aetos.pwx.purestorage.com/dashboard" //"http://aetos-dm.pwx.purestorage.com:3939/dashboard"
+	AetosBaseURL     = "http://aetos.pwx.purestorage.com"
```
lsrinivas-pure, line 32: Is the API url changed now?

PR #949, opened 2022-11-16
Adding tags for pwx deployments to aetos

aghodke1312:
http://aetos.pwx.purestorage.com/resultSet/testSetID/15368

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -9,8 +9,13 @@ if [ -z "${ENABLE_DASH}" ]; then
 fi
 
 if [ -z "${DASH_UID}" ]; then
-    DASH_UID="0"
+    if [ -e /build.properties ]; then
+      DASH_UID=`cat /build.properties | grep -oE '[0-9]+'`
+    else
+      DASH_UID="0"
+    fi
 fi
+
 
 if [ -z "${SCALE_FACTOR}" ]; then
     SCALE_FACTOR="10"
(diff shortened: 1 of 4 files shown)


deployments/deploy-ssh.sh
```diff
@@ -9,9 +9,14 @@ if [ -z "${ENABLE_DASH}" ]; then
 fi
 
 if [ -z "${DASH_UID}" ]; then
-    DASH_UID="0"
+    if [ -e /build.properties ]; then
+      DASH_UID=`cat /build.properties | grep -oE '[0-9]+'`
```
CharudathGopal, line 13: This logic will break if there are multiple env variables in the build.prop.. recently I added a logic for this in spawn.groovy. Take a look at that

tests/common.go
```diff
@@ -4909,6 +4909,10 @@ func GetCloudDriveDeviceSpecs() ([]string, error) {
 func StartTorpedoTest(testName, testDescription string, tags map[string]string, testRepoID int) {
 	TestLogger = CreateLogger(fmt.Sprintf("%s.log", testName))
 	SetTorpedoFileOutput(log, TestLogger)
+	tags["apps"] = strings.Join(Inst().AppList, ",")
```
CharudathGopal, line 4912: What does provisioner means here? Spawn also pushes similar key.. cross check once it might get overwritten

CharudathGopal (approved):

PR #965, opened 2022-11-23
Added initial test case for User/Group management with role-mappings

mkoppal-px:
This PR contains a sample test case which is essential for User and Group management using keycloak APIs. Also, we will be assigned Px-Backup specific roles to groups and validating it.

Changes include:

- Sample test case for keycloak user/group management and role-mappings
- Includes environment variables for backup api and keycloak endpoint
- Logic to utilise backup and keycloak endpoint when running locally or from a VM outside of the k8s cluster
- Added functions like GetRolesForUser, ValidateUserRole, AddRoleToGroup, DeleteRoleFromGroup, DeleteUser, DeleteGroup
- Formatting a few comments

[Testrun link](http://aetos.pwx.purestorage.com/resultSet/testSetID/21167) 

Closes # PA-308

- The sample test case is just a building block for the real system tests that will be added in the coming days. These functions will be used mainly in the Share Backup feature where we need to test with multiple users and groups with different levels of access to the shared backup
- Please go easy since this is my first PR 😄

### DIFF
--- a/drivers/backup/auth.go
+++ b/drivers/backup/auth.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -287,7 +286,7 @@ func GetRolesForUser(userName string) ([]KeycloakRoleRepresentation, error) {
 	fn := "GetRolesForUser"
 	headers, err := GetCommonHTTPHeaders(PxCentralAdminUser, PxCentralAdminPwd)
 	if err != nil {
-		logrus.Errorf("%s: %v", fn, err)
+		log.Errorf("%s: %v", fn, err)
 		return nil, err
 	}
 	keycloakEndPoint, err := getKeycloakEndPoint(true)
@@ -302,35 +301,16 @@ func GetRolesForUser(userName string) ([]KeycloakRoleRepresentation, error) {
 	method := "GET"
 	response, err := processHTTPRequest(method, reqURL, headers, nil)
 	if err != nil {
-		logrus.Errorf("%s: %v", fn, err)
+		log.Errorf("%s: %v", fn, err)
 		return nil, err
 	}
 	var roles []KeycloakRoleRepresentation
 	err = json.Unmarshal(response, &roles)
 	if err != nil {
-		logrus.Errorf("%s: %v", fn, err)
+		log.Errorf("%s: %v", fn, err)
 		return nil, err
 	}
 	return roles, nil
-}
-
-// ValidateUserRole will validate if a given user has the provided PxBackupRole mapped to it
-func ValidateUserRole(userName string, role PxBackupRole) bool {
-	roleMapping, err := GetRolesForUser(userName)
-	if err != nil {
-		return false
-	}
-	roleID, err := GetRoleID(role)
-	if err != nil {
-		return false
-	}
-	for _, r := range roleMapping {
-		if r.ID == roleID {
-			return true
-		}
-	}
-
-	return false
 }
 
 type PxBackupRole string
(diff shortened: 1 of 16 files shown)


drivers/backup/auth.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/sirupsen/logrus"
```
lsrinivas-pure, line 7: rebase with latest changes and use pkg/log

drivers/backup/auth.go
```diff
@@ -264,8 +282,68 @@ func GetAllRoles() ([]KeycloakRoleRepresentation, error) {
...
+		return nil, err
+	}
+	return roles, nil
+}
+
+// ValidateUserRole will validate if a given user has the provided PxBackupRole mapped to it
+func ValidateUserRole(userName string, role PxBackupRole) bool {
+	roleMapping, err := GetRolesForUser(userName)
+	if err != nil {
+		return false
```
lsrinivas-pure, line 321: retruning just false we will miss to log the error, fail it here logging the error or return the error

drivers/backup/auth.go
```diff
@@ -264,8 +282,68 @@ func GetAllRoles() ([]KeycloakRoleRepresentation, error) {
...
+	err = json.Unmarshal(response, &roles)
+	if err != nil {
+		logrus.Errorf("%s: %v", fn, err)
+		return nil, err
+	}
+	return roles, nil
+}
+
+// ValidateUserRole will validate if a given user has the provided PxBackupRole mapped to it
+func ValidateUserRole(userName string, role PxBackupRole) bool {
```
lsrinivas-pure, line 318: Validate can we move it to some common files , rather than in drivers

tests/backup/backup_test.go
```diff
@@ -71,7 +70,72 @@ var _ = Describe("{BackupClusterVerification}", func() {
...
+// This is a sample test case to verify User/Group Management and role mapping
+var _ = Describe("{UserGroupManagement}", func() {
+	JustBeforeEach(func() {
+		log.Infof("No pre-setup required for this testcase")
+		StartTorpedoTest("Backup: UserGroupManagement", "Creating users and adding them to groups", nil, 0)
+	})
+	It("User and group role mappings", func() {
+		Step("Create Users", func() {
+			err := backup.AddUser("testuser1", "test", "user1", "testuser1@localhost.com", "Password1")
+			if err != nil {
```
lsrinivas-pure, line 82: use log.FailOnError ,so that we can avoid err !=nil check

lsrinivas-pure (changes requested):

drivers/backup/auth.go
mkoppal-px: Addressed

drivers/backup/auth.go
mkoppal-px: Done. Moved it to common.go

drivers/backup/auth.go
mkoppal-px: Now returning bool and err. Then handling error and failing it.

tests/backup/backup_test.go
mkoppal-px: Done. Makes sense.

lsrinivas-pure (approved):
LGTM

kshithijiyer-px (commented):
A few things we can do is:

- [ ] Add group to group mapping function which cane be used. 
- [ ] Add group to default role mapping function. 
- [ ] Add user to default role mapping function.

This will be helpful for us to test scenarios like OIDC to keycloak mapping scenarios.

drivers/backup/auth.go
```diff
@@ -54,6 +54,8 @@ const (
 	pxBackupNamespace = "PX_BACKUP_NAMESPACE"
 	// OidcSecretName where secrets for OIDC auth cred info resides
 	oidcSecretName = "SECRET_NAME"
+	// PxCentralUI URL Eg: http://10.13.114.233:32482
```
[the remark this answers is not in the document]
mkoppal-px: Done

kshithijiyer-px (approved):
LGTM

mkoppal-px:
> A few things we can do is:
> 
> * [ ]  Add group to group mapping function which cane be used.
> * [ ]  Add group to default role mapping function.
> * [ ]  Add user to default role mapping function.
> 
> This will be helpful for us to test scenarios like OIDC to keycloak mapping scenarios.

@kshithijiyer-px Created PA tickets for these - PA-404 and PA-405

PR #970, opened 2022-12-01
Reduced namespace length, removed extra date param and reduced task p…

gejain:
It will fix namespace length issue (< 63 chars)

https://portworx.atlassian.net/browse/PTX-14206

Ran it locally, not its removing duplicate dates and reduced task prefix size as well

### DIFF
--- a/tests/testTriggers.go
+++ b/tests/testTriggers.go
@@ -5836,7 +5836,7 @@ func TriggerAsyncDRVolumeOnly(contexts *[]*scheduler.Context, recordChan *chan *
 	chaosLevel := ChaosMap[AsyncDRVolumeOnly]
 	var (
 		migrationNamespaces   []string
-		taskNamePrefix        = "async-dr-vol-only-mig"
+		taskNamePrefix        = "adr-vonly"
 		allMigrations         []*storkapi.Migration
 		includeResourcesFlag  = false
 		startApplicationsFlag = false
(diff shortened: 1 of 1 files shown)


tests/testTriggers.go
```diff
@@ -5856,7 +5856,7 @@ func TriggerAsyncDRVolumeOnly(contexts *[]*scheduler.Context, recordChan *chan *
 		}
 		UpdateOutcome(event, err)
 		for i := 0; i < Inst().GlobalScaleFactor; i++ {
-			taskName := fmt.Sprintf("%s-%d-%s", taskNamePrefix, i, time.Now().Format("15h03m05s"))
+			taskName := fmt.Sprintf("%s-%d", taskNamePrefix, i)
```
lsrinivas-pure, line 5859: leave the time, it will be consistent with other apps

lsrinivas-pure (changes requested):

tests/testTriggers.go
gejain: Time is getting append at the end of the string, please refer below its coming twice -

fio-fa-cloud-drives-async-dr-vol-only-mig-0-12h12m49s-11-23-10h21m50s

So when I removed this, I see something like below -

with date -
mysql-adr-vonly-0-22h10m55s-11-30-22h09m44s   Active   11h

without date -
mysql-adr-vonly-0-11-30-22h34m14s             Active   11h

lsrinivas-pure (approved):
LGTM

PR #990, opened 2022-12-20
PWX-27845 : Fix for Torpedo CSICloneTest test intermittent failure

ashinde-px:
Signed-off-by: Amol Shinde <ashinde@purestorage.com>

Torpedo CSICloneTest test intermittent failure, changed k8sObjectCreateTimeout to 4 minutes

Closes #PWX-27845

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -125,7 +125,7 @@ const (
 const (
 	k8sNodeReadyTimeout    = 5 * time.Minute
 	volDirCleanupTimeout   = 5 * time.Minute
-	k8sObjectCreateTimeout = 2 * time.Minute
+	k8sObjectCreateTimeout = 4 * time.Minute
 	k8sDestroyTimeout      = 5 * time.Minute
 	// FindFilesOnWorkerTimeout timeout for find files on worker
 	FindFilesOnWorkerTimeout = 1 * time.Minute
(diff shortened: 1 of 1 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -125,7 +125,7 @@ const (
 const (
 	k8sNodeReadyTimeout    = 5 * time.Minute
 	volDirCleanupTimeout   = 5 * time.Minute
-	k8sObjectCreateTimeout = 2 * time.Minute
+	k8sObjectCreateTimeout = 4 * time.Minute
```
Pure-AdamuKaapan, line 128: Is this just "how long to wait for pods to be ready", that kind of thing?

Are we sure that we need to increase the timeout and we aren't covering up some other issue? Or is it just like... statefulsets starting 1 at a time and that takes too long?

ashinde-px: Yes Its about statefulsets only. About creation of k8s object.

ashinde-px: Thanks @Pure-AdamuKaapan I have updated code accordingly, Added new timeout variable for CSI clone and snapshot functionality.

drivers/scheduler/k8s/k8s.go
```diff
@@ -126,6 +126,7 @@ const (
 	k8sNodeReadyTimeout    = 5 * time.Minute
 	volDirCleanupTimeout   = 5 * time.Minute
 	k8sObjectCreateTimeout = 2 * time.Minute
+	k8sObjectCloneSnapshotTimeout = 4 * time.Minute
```
smaanika-px, line 129: Amol - Since this timeout is applied for pod creation, should we call it as k8sPodCreateTimeout?

ashinde-px: Done

nikolaypopov (approved):
LGTM

PR #994, opened 2022-12-21
Fix error handling

Madanagopal19:
Fix error handling

### DIFF
--- a/tests/pds/dataservice_test.go
+++ b/tests/pds/dataservice_test.go
@@ -1,6 +1,7 @@
 package tests
 
 import (
+	"errors"
 	"net/http"
 	"strconv"
 	"sync"
@@ -24,6 +25,29 @@ const (
 	defaultTestConnectionTimeout = 15 * time.Minute
 	defaultRebootTimeRange       = 5 * time.Minute
 )
+
+var _ = Describe("{CollectLogs}", func() {
+	JustBeforeEach(func() {
+		StartTorpedoTest("DeletePDSPods", "delete pds pods and validate if its coming back online and dataserices are not affected", nil, 0)
+	})
+
+	It("Collect logs", func() {
+		Step("collect container logs", func() {
+			isDeploymentsDeleted = false
+			err = pdslib.CollectPodLogsandValidateWorkloads()
+			err = errors.New("an error message")
+			Expect(err).ShouldNot(HaveOccurred())
+			//log.FailOnError(err, "Failed while collecting pod logs")
+		})
+	})
+	JustAfterEach(func() {
+		if !isDeploymentsDeleted {
+			Step("Delete created deployments")
+		}
+
+		defer EndTorpedoTest()
+	})
+})
 
 var _ = Describe("{DeletePDSPods}", func() {
 	JustBeforeEach(func() {
@@ -348,7 +372,8 @@ var _ = Describe("{ScaleUPDataServices}", func() {
 						dataServiceDefaultAppConfigID, deployment.GetImageId(),
 						int32(ds.ScaleReplicas), dataServiceDefaultResourceTemplateID, namespace)
 
-					log.FailOnError(err, "Error while updating dataservices")
+					// log.FailOnError(err, "Error while updating dataservices")
+					Expect(err).ShouldNot(HaveOccurred())
 					log.InfoD("Scaling up DataService %v ", ds.Name)
 
 					resourceTemp, storageOp, config, err := pdslib.ValidateDataServiceVolumes(updatedDeployment, ds.Name, dataServiceDefaultResourceTemplateID, storageTemplateID, namespace)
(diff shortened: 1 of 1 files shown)


drivers/pds/lib/utils.go
```diff
@@ -914,6 +914,30 @@ func CreatepostgresqlWorkload(dnsEndpoint string, pdsPassword string, scalefacto
 	return deployment, err
 }
 
+func CollectPodLogsandValidateWorkloads() error {
+	pods, err := k8sCore.GetPods("automation", nil)
+	if err != nil {
+		log.Errorf("An Error occured while getting pods")
```
nikolaypopov, line 920: Why are we logging this error instead of returning?
`return fmt.Errorf("Failed to get pods, Err; %v", err)

drivers/pds/lib/utils.go
```diff
@@ -914,6 +914,30 @@ func CreatepostgresqlWorkload(dnsEndpoint string, pdsPassword string, scalefacto
...
+	}
+	failure_messages := []string{"Name or service not known", "could not translate host name", "Connection refused"}
+	for _, pod := range pods.Items {
+		if pod.Name == "pgload-slcww-64b4b7ffdd-bjt7w" {
+			podlogs, err := k8sCore.GetPodLog(pod.Name, pod.Namespace, &corev1.PodLogOptions{})
+			if err != nil {
+				log.Errorf("An Error occured while getting pod logs")
```
nikolaypopov, line 927: Same here, can we return instead of logging please
`return fmt.Errorf("Failed to get logs from pod [%s] in namespace [%s], Err: %v", pod.Name, pod.Namespace, err)`

drivers/pds/lib/utils.go
```diff
@@ -914,6 +914,30 @@ func CreatepostgresqlWorkload(dnsEndpoint string, pdsPassword string, scalefacto
...
+			log.Infof(podlogs)
+			for _, messages := range failure_messages {
+				if strings.Contains(podlogs, messages) {
+					log.Infof("Unable to connect to the database")
+				}
+			}
+			break
+		}
+	}
+	return err
```
nikolaypopov, line 938: Should return nil here, since we should fail on every previous error above

drivers/pds/lib/utils.go
```diff
@@ -914,6 +914,30 @@ func CreatepostgresqlWorkload(dnsEndpoint string, pdsPassword string, scalefacto
...
+	for _, pod := range pods.Items {
+		if pod.Name == "pgload-slcww-64b4b7ffdd-bjt7w" {
+			podlogs, err := k8sCore.GetPodLog(pod.Name, pod.Namespace, &corev1.PodLogOptions{})
+			if err != nil {
+				log.Errorf("An Error occured while getting pod logs")
+			}
+			log.Infof(podlogs)
+			for _, messages := range failure_messages {
+				if strings.Contains(podlogs, messages) {
+					log.Infof("Unable to connect to the database")
```
nikolaypopov, line 932: If we match the expected error message to the logs in pod, shouldn't we fail here? If we do fail here we don't need that `break` since we will just return if we find expected error

tests/pds/dataservice_test.go
```diff
@@ -25,6 +26,29 @@ const (
...
+		StartTorpedoTest("DeletePDSPods", "delete pds pods and validate if its coming back online and dataserices are not affected", nil, 0)
+	})
+
+	It("Collect logs", func() {
+		Step("collect container logs", func() {
+			isDeploymentsDeleted = false
+			err = pdslib.CollectPodLogsandValidateWorkloads()
+			err = errors.New("an error message")
+			Expect(err).ShouldNot(HaveOccurred())
+			//log.FailOnError(err, "Failed while collecting pod logs")
```
nikolaypopov, line 40: Remove this line if not needed

tests/pds/dataservice_test.go
```diff
@@ -25,6 +26,29 @@ const (
...
+var _ = Describe("{CollectLogs}", func() {
+	JustBeforeEach(func() {
+		StartTorpedoTest("DeletePDSPods", "delete pds pods and validate if its coming back online and dataserices are not affected", nil, 0)
+	})
+
+	It("Collect logs", func() {
+		Step("collect container logs", func() {
+			isDeploymentsDeleted = false
+			err = pdslib.CollectPodLogsandValidateWorkloads()
+			err = errors.New("an error message")
```
nikolaypopov, line 38: Why are you overwriting error from `pdslib.CollectPodLogsandValidateWorkloads()` with a simple `an error message`?

tests/pds/dataservice_test.go
```diff
@@ -348,7 +372,8 @@ var _ = Describe("{ScaleUPDataServices}", func() {
 						dataServiceDefaultAppConfigID, deployment.GetImageId(),
 						int32(ds.ScaleReplicas), dataServiceDefaultResourceTemplateID, namespace)
 
-					log.FailOnError(err, "Error while updating dataservices")
+					// log.FailOnError(err, "Error while updating dataservices")
```
nikolaypopov, line 375: Remove this if not needed

tests/pds/dataservice_test.go
```diff
@@ -25,6 +26,29 @@ const (
...
+			isDeploymentsDeleted = false
+			err = pdslib.CollectPodLogsandValidateWorkloads()
+			err = errors.New("an error message")
+			Expect(err).ShouldNot(HaveOccurred())
+			//log.FailOnError(err, "Failed while collecting pod logs")
+		})
+	})
+	JustAfterEach(func() {
+		if !isDeploymentsDeleted {
+			Step("Delete created deployments")
```
nikolaypopov, line 45: Looks like we saying we going to delete created deployment, but not doing anything here, doesn't seem right

drivers/pds/lib/utils.go
```diff
@@ -914,6 +914,30 @@ func CreatepostgresqlWorkload(dnsEndpoint string, pdsPassword string, scalefacto
...
 }
 
+func CollectPodLogsandValidateWorkloads() error {
+	pods, err := k8sCore.GetPods("automation", nil)
+	if err != nil {
+		log.Errorf("An Error occured while getting pods")
+	}
+	failure_messages := []string{"Name or service not known", "could not translate host name", "Connection refused"}
+	for _, pod := range pods.Items {
+		if pod.Name == "pgload-slcww-64b4b7ffdd-bjt7w" {
```
nikolaypopov, line 924: Are you sure this `pgload-slcww-64b4b7ffdd-bjt7w` always going to be a name for this pod?

nikolaypopov (changes requested):

PR #1003, opened 2023-01-13
Making Bucket name env variable

kshithijiyer-px:
Currently if multiple users run torpedo tests with same object store all the instances try to create the same bucket and it fails with the below trance: 
```
• Failure [93.714 seconds]
{BasicBackupCreation}
/root/torpedo/tests/backup/backup_test.go:125
  Basic Backup Creation [It]
  /root/torpedo/tests/backup/backup_test.go:174

  Failed to create bucket [bucket-aws]. Error: [BucketAlreadyOwnedByYou: Your previous request to create the named bucket succeeded and you already own it.
  	status code: 409, request id: 17397DE2EFAF9A3E, host id: ]
  Unexpected error:
      <*s3err.RequestFailure | 0xc001035ca0>: {
          RequestFailure: <*awserr.requestError | 0xc001096500>{
              awsError: <*awserr.baseError | 0xc0010964c0>{
                  code: "BucketAlreadyOwnedByYou",
                  message: "Your previous request to create the named bucket succeeded and you already own it.",
                  errs: nil,
              },
              statusCode: 409,
              requestID: "17397DE2EFAF9A3E",
              bytes: nil,
          },
          hostID: "",
      }
      BucketAlreadyOwnedByYou: Your previous request to create the named bucket succeeded and you already own it.
      	status code: 409, request id: 17397DE2EFAF9A3E, host id: 
```

**Changes done:**
- [x] Add `getBucketName()` to fetch bucket name from env. 
- [x] Make changes to` BasicBackupCreation` to pick bucket name from env. 
- [x] Moving `CreateBucket()` call to `BeforeSuite()` where one bucket is created which can be used in any testcase. 
- [x] Moving `DeleteBucket()` call  to `AfterSuite()` and adding logic to delete all buckets present.  

**Changes needed while running the test**:
You'll need to export backup name list as shown below:
```
export BUCKET_NAME="kiyer-test-bucket"
```

Closes # PA-491

The test might not run seamlessly @snigdha-px is working on the fixes to make it run.

### DIFF
--- a/tests/backup/backup_basic_test.go
+++ b/tests/backup/backup_basic_test.go
@@ -4,9 +4,10 @@ import (
 	"os"
 	"testing"
 	"time"
+	"strings"
+	"fmt"
 
 	"github.com/portworx/torpedo/pkg/log"
-
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/reporters"
 	. "github.com/onsi/gomega"
@@ -47,6 +48,11 @@ var (
 	}
 )
 
+func getBucketName() []string{
+	bucketName := os.Getenv("BUCKET_NAME")
+	return strings.Split(bucketName, ",")
+}
+
 func TestBasic(t *testing.T) {
 	RegisterFailHandler(Fail)
 
@@ -62,9 +68,25 @@ var _ = BeforeSuite(func() {
 	log.Infof("Init instance")
 	InitInstance()
 	dash.TestSetBegin(dash.TestSet)
+	// Create the first bucket from the list to be used as generic bucket
+	providers := getProviders()
+	bucket_name := getBucketName()
+	for _, provider := range providers {
+		bucketName := fmt.Sprintf("%s-%s", provider, bucket_name[0])
+		CreateBucket(provider, bucketName)
+		}
 })
 
 var _ = AfterSuite(func() {
+	// Cleanup all bucket after suite
+	providers := getProviders()
+	bucket_name := getBucketName()
+	for _, provider := range providers{
+		for _, BucketName := range bucket_name{
+				bucketName := fmt.Sprintf("%s-%s", provider, BucketName)
+				DeleteBucket(provider, bucketName)
+		}
+	}
 	defer dash.TestSetEnd()
 })
 
--- a/tests/backup/backup_basic_test.go
+++ b/tests/backup/backup_basic_test.go
@@ -68,6 +68,7 @@ var _ = BeforeSuite(func() {
 	log.Infof("Init instance")
 	InitInstance()
 	dash.TestSetBegin(dash.TestSet)
+	StartTorpedoTest("Setup buckets:", "Creating one generic bucket to be used in all cases", nil, 0)
 	// Create the first bucket from the list to be used as generic bucket
 	providers := getProviders()
 	bucket_name := getBucketName()
@@ -87,6 +88,7 @@ var _ = AfterSuite(func() {
 				DeleteBucket(provider, bucketName)
 		}
 	}
+	defer EndTorpedoTest()
 	defer dash.TestSetEnd()
 })
 
--- a/tests/backup/backup_basic_test.go
+++ b/tests/backup/backup_basic_test.go
@@ -76,9 +76,11 @@ var _ = BeforeSuite(func() {
 		bucketName := fmt.Sprintf("%s-%s", provider, bucket_name[0])
 		CreateBucket(provider, bucketName)
 		}
+	defer EndTorpedoTest()
 })
 
 var _ = AfterSuite(func() {
+	StartTorpedoTest("Cleanup buckets:", "Removing buckets", nil, 0)
 	// Cleanup all bucket after suite
 	providers := getProviders()
 	bucket_name := getBucketName()
--- a/tests/backup/backup_basic_test.go
+++ b/tests/backup/backup_basic_test.go
@@ -69,6 +69,7 @@ var _ = BeforeSuite(func() {
 	InitInstance()
 	dash.TestSetBegin(dash.TestSet)
 	StartTorpedoTest("Setup buckets:", "Creating one generic bucket to be used in all cases", nil, 0)
+	defer EndTorpedoTest()
 	// Create the first bucket from the list to be used as generic bucket
 	providers := getProviders()
 	bucket_name := getBucketName()
@@ -76,11 +77,12 @@ var _ = BeforeSuite(func() {
 		bucketName := fmt.Sprintf("%s-%s", provider, bucket_name[0])
 		CreateBucket(provider, bucketName)
 		}
-	defer EndTorpedoTest()
 })
 
 var _ = AfterSuite(func() {
 	StartTorpedoTest("Cleanup buckets:", "Removing buckets", nil, 0)
+	defer EndTorpedoTest()
+	defer dash.TestSetEnd()
 	// Cleanup all bucket after suite
 	providers := getProviders()
 	bucket_name := getBucketName()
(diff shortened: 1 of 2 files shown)


skonda-px (approved):
LGTM

mkoppal-px:
We are planning to create bucket names by appending timestamp. Eg: `autogenerated-bucket-aws-1673581777`
This will eliminate the need for setting environment variable.

What we can do is add a condition


```
if bucket name is set in ENV
 <make use of it>
else generate a bucket name
 bucketName := fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
```

to check if the env variable is set we can use

```
bucketName, present := os.LookupEnv(BUCKET_NAME)
	if !(present) {
		return fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
	}
```

Let me know your thoughts.

lsrinivas-pure:
> We are planning to create bucket names by appending timestamp. Eg: `autogenerated-bucket-aws-1673581777` This will eliminate the need for setting environment variable.
> 
> What we can do is add a condition
> 
> ```
> if bucket name is set in ENV
>  <make use of it>
> else generate a bucket name
>  bucketName := fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
> ```
> 
> to check if the env variable is set we can use
> 
> ```
> bucketName, present := os.LookupEnv(BUCKET_NAME)
> 	if !(present) {
> 		return fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
> 	}
> ```
> 
> Let me know your thoughts. The objective is to not fail the test just because a BUCKET_NAME was not provided. It should create a new bucket and move on.

yes, we can use time stamp. also are we deleting the bucket once test is done?

mkoppal-px:
Yes. @snigdha-px is working on cleaning up the bucket after the test is executed.

snigdha-px:
> > We are planning to create bucket names by appending timestamp. Eg: `autogenerated-bucket-aws-1673581777` This will eliminate the need for setting environment variable.
> > What we can do is add a condition
> > ```
> > if bucket name is set in ENV
> >  <make use of it>
> > else generate a bucket name
> >  bucketName := fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to check if the env variable is set we can use
> > ```
> > bucketName, present := os.LookupEnv(BUCKET_NAME)
> > 	if !(present) {
> > 		return fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
> > 	}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let me know your thoughts. The objective is to not fail the test just because a BUCKET_NAME was not provided. It should create a new bucket and move on.
> 
> yes, we can use time stamp. also are we deleting the bucket once test is done?

Yes I am taking care of creating bucket name using time stamp. 
Deleting buckets at the end of testcase is also what is my next task

snigdha-px:
> We are planning to create bucket names by appending timestamp. Eg: `autogenerated-bucket-aws-1673581777` This will eliminate the need for setting environment variable.
> 
> What we can do is add a condition
> 
> ```
> if bucket name is set in ENV
>  <make use of it>
> else generate a bucket name
>  bucketName := fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
> ```
> 
> to check if the env variable is set we can use
> 
> ```
> bucketName, present := os.LookupEnv(BUCKET_NAME)
> 	if !(present) {
> 		return fmt.Sprintf("%s-%s-%v", "autogenerated-bucket", provider, time.Now().Unix())
> 	}
> ```
> 
> Let me know your thoughts. The objective is to not fail the test just because a BUCKET_NAME was not provided. It should create a new bucket and move on.

Yes correct. I am trying to creating all the variables like bucket name, rule names and all possibles values by appending timestamp so that previous already created objects does not fail the current run.

kshithijiyer-px:
@lsrinivas-pure @mkoppal-px @snigdha-px There are a few problems with creating timestamp based buckets:

1. Let's say 2 users trigger the torpedo at same time it will hit the same issue as one of them will get the bucket already exists issue. This can happen with multiple jobs running at the same time as well. 
2. With `autogenerated-bucket-aws-1673581777` if the automation fails how does one figure out which bucket is for which job.
3. How does one figure out who owns the buckets when we have such bucket names `autogenerated-bucket-aws-1673581777`?

That's why I am considering to move the code to take env which can be populated based on Jenkins job names.

mkoppal-px:
> @lsrinivas-pure @mkoppal-px @snigdha-px There are a few problems with creating timestamp based buckets:
> 
> 1. Let's say 2 users trigger the torpedo at same time it will hit the same issue as one of them will get the bucket already exists issue. This can happen with multiple jobs running at the same time as well.
> 2. With `autogenerated-bucket-aws-1673581777` if the automation fails how does one figure out which bucket is for which job.
> 3. How does one figure out who owns the buckets when we have such bucket names `autogenerated-bucket-aws-1673581777`?
> 
> That's why I am considering to move the code to take env which can be populated based on Jenkins job names.

1. Two users triggering the same job at the very same second is highly unlikely.  We can drill down the timestamp to milliseconds if that happens frequently and causes problems.
2. When automation fails we would know the jenkins job and in the console of that job we would get the bucket name.
3. The bucket owner will be the user who calls the API right?

Moreover, we can have the env variable. The only extra logic that I am suggesting is creating a new bucket if it is not present. This will eliminate failures in case the cleanup fails and duplicate buckets are left over. Which is exactly what we are facing now.

kshithijiyer-px:
A summary of offline discussion we had: 
1. The issue with having `autogenerated-bucket-aws-1673581777` is that you'll have a bunch of buckets which can't be tied to any user or job with which we will end up having a large number of orphan buckets.
2. Creating one bucket per case would create multiple buckets which need to be cleaned up post run hence adding generic code to create a single bucket and cleanup post run in `BeforeSuite()` and `AfterSuite()`

tests/backup/backup_basic_test.go
```diff
@@ -62,9 +68,25 @@ var _ = BeforeSuite(func() {
 	log.Infof("Init instance")
 	InitInstance()
 	dash.TestSetBegin(dash.TestSet)
+	// Create the first bucket from the list to be used as generic bucket
+	providers := getProviders()
+	bucket_name := getBucketName()
+	for _, provider := range providers {
```
lsrinivas-pure, line 74: can we add testcasebegin and testcaseend for create and delete buckets, so that these are logged in dashboard

lsrinivas-pure (changes requested):

tests/backup/backup_test.go
```diff
@@ -201,14 +199,14 @@ var _ = Describe("{BasicBackupCreation}", func() {
 
 		Step("Creating bucket,backup location and cloud setting", func() {
 			log.InfoD("Creating bucket,backup location and cloud setting")
+			bucket_name := getBucketName()
 			for _, provider := range providers {
-				bucketName := fmt.Sprintf("%s-%s", "bucket", provider)
+				bucketName := fmt.Sprintf("%s-%s", provider, bucket_name[0])
```
snigdha-px, line 204: As per our last discussion, for this test case we need to add timestamp only for bucket name and backup location as this is just UT and will not run in jenkin.

I have added the timestamp part in my PR.
For the real test cases we will use the method  getBucketName

snigdha-px (changes requested):

snigdha-px (approved):
LGTM

tests/backup/backup_basic_test.go
kshithijiyer-px: Done

lsrinivas-pure (approved):
LGTM

PR #1022, opened 2023-01-20
PTX-15067:Torpedo job doesn't fail if core was found on the node

CharudathGopal:
- Made change to fail the test upon core found

### DIFF
--- a/tests/basic/ipv6_test.go
+++ b/tests/basic/ipv6_test.go
@@ -26,7 +26,9 @@ var _ = Describe("{IPv6PxctlFunctional}", func() {
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
 		nodes = node.GetWorkerNodes()
 		numNodes = len(nodes)
-		StartTorpedoTest("IPv6PxctlCommands", "Pxctl commands on IPv6 setups test", nil, testrailID)
+		tags := make(map[string]string, 0)
+		tags["ipv6"] = "true"
+		StartTorpedoTest("IPv6PxctlCommands", "Pxctl commands on IPv6 setups test", tags, testrailID)
 	})
 
 	Context("{IPv6PxctlCommands}", func() {
--- a/tests/basic/ipv6_test.go
+++ b/tests/basic/ipv6_test.go
@@ -26,8 +26,9 @@ var _ = Describe("{IPv6PxctlFunctional}", func() {
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
 		nodes = node.GetWorkerNodes()
 		numNodes = len(nodes)
-		tags := make(map[string]string, 0)
-		tags["ipv6"] = "true"
+		tags := map[string]string{
+			"ipv6": "true",
+		}
 		StartTorpedoTest("IPv6PxctlCommands", "Pxctl commands on IPv6 setups test", tags, testrailID)
 	})
 
--- a/tests/basic/longevity_test.go
+++ b/tests/basic/longevity_test.go
@@ -108,7 +108,9 @@ var _ = Describe("{Longevity}", func() {
 
 	BeforeEach(func() {
 		if !populateDone {
-			StartTorpedoTest("PX-Longevity", "Validate PX longevity workflow", nil, 0)
+			tags := make(map[string]string, 0)
+			tags["longevity"] = "true"
+			StartTorpedoTest("PX-Longevity", "Validate PX longevity workflow", tags, 0)
 
 			populateIntervals()
 			populateDisruptiveTriggers()
--- a/tests/basic/longevity_test.go
+++ b/tests/basic/longevity_test.go
@@ -108,8 +108,9 @@ var _ = Describe("{Longevity}", func() {
 
 	BeforeEach(func() {
 		if !populateDone {
-			tags := make(map[string]string, 0)
-			tags["longevity"] = "true"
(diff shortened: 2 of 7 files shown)


tests/basic/ipv6_test.go
```diff
@@ -26,7 +26,9 @@ var _ = Describe("{IPv6PxctlFunctional}", func() {
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
 		nodes = node.GetWorkerNodes()
 		numNodes = len(nodes)
-		StartTorpedoTest("IPv6PxctlCommands", "Pxctl commands on IPv6 setups test", nil, testrailID)
+		tags := make(map[string]string, 0)
+		tags["ipv6"] = "true"
```
mborodin-px, line 30: you can initialize map[string]string with values like this:
```
tags := map[string]string{
		"ipv6": true,
}
```

tests/basic/longevity_test.go
```diff
@@ -108,7 +108,9 @@ var _ = Describe("{Longevity}", func() {
 
 	BeforeEach(func() {
 		if !populateDone {
-			StartTorpedoTest("PX-Longevity", "Validate PX longevity workflow", nil, 0)
+			tags := make(map[string]string, 0)
+			tags["longevity"] = "true"
```
mborodin-px, line 112: the same as above

tests/basic/unhappypath_test.go
```diff
@@ -41,7 +41,9 @@ var _ = Describe("{NetworkErrorInjection}", func() {
 	// testrailID corresponds to: https://portworx.testrail.net/index.php?/cases/view/35264
 	var runID int
 	JustBeforeEach(func() {
-		StartTorpedoTest("NetworkErrorInjection", "Verify stability of the system when there is a network error", nil, testrailID)
+		tags := make(map[string]string, 0)
+		tags["networkErrorInjection"] = "true"
```
mborodin-px, line 45: the same as above

tests/basic/upgrade_cluster_test.go
```diff
@@ -15,7 +15,9 @@ var _ = Describe("{UpgradeCluster}", func() {
 	var contexts []*scheduler.Context
 
 	JustBeforeEach(func() {
-		StartTorpedoTest("UpgradeCluster", "Upgrade cluster test", nil, 0)
+		tags := make(map[string]string, 0)
+		tags["upgradeCluster"] = "true"
```
mborodin-px, line 19: the same as above

tests/common.go
```diff
@@ -1827,12 +1828,18 @@ func PerformSystemCheck() {
 					TimeBeforeRetry: 10 * time.Second,
 				})
 				if len(file) != 0 || err != nil {
-					log.Errorf("an error occurred, collecting bundle")
-					CollectSupport()
-					dash.VerifySafely(file == "", true, fmt.Sprintf("Core generated on node %s, Core Path: %s", n.Name, file))
+					dash.Errorf("Core generated on node %s, Core Path: %s", n.Name, file)
```
mborodin-px, line 1831: I would re-phrase the error message to `Core file was found on node %n`

tests/common.go
```diff
@@ -1827,12 +1828,18 @@ func PerformSystemCheck() {
...
+					// Collect Support Bundle only once
+					if !foundCore {
+						CollectSupport()
+					}
+					foundCore = true
```
mborodin-px, line 1836: you don't need to continue if core file was found. Just `break` from the loop as the `CollectSupport()` method has already collected all needed logs and diags from all nodes

tests/common.go
```diff
@@ -1827,12 +1828,18 @@ func PerformSystemCheck() {
...
 				}
 				log.FailOnError(err, "Error occurred while checking for core on node %s", n.Name)
 			}
+			if foundCore {
+				log.Fatalf("Core generated, please check logs for more details")
+			}
```
mborodin-px, line 1842: don't need this as above comment

tests/common.go
CharudathGopal: I am continuing the loop to logs message if there is core in other nodes too. it wont collect support bundle but just logs the info if core found. Let me know if this doesn't make sense

CharudathGopal:
Thanks for review Maks. Taken care of comments.

tests/common.go
```diff
@@ -1827,11 +1827,11 @@ func PerformSystemCheck() {
 					TimeBeforeRetry: 10 * time.Second,
 				})
 				if len(file) != 0 || err != nil {
-					log.Errorf("an error occurred, collecting bundle")
+					dash.Errorf("Core file was found on node %s, Core Path: %s", n.Name, file)
+					// Collect Support Bundle only once
 					CollectSupport()
-					dash.VerifySafely(file == "", true, fmt.Sprintf("Core generated on node %s, Core Path: %s", n.Name, file))
+					log.Fatalf("Core generated, please check logs for more details")
```
mborodin-px, line 1833: you have to add `break` here to get out from loop

mborodin-px (approved):
LGTM

tests/common.go
CharudathGopal: no need @mborodin-px , once the log.Fatal is called it will fail the test and return

kshithijiyer-px (approved):

PR #1028, opened 2023-01-22
Added testcase :Share backups to users with different roles:82947

snigdha-px:
https://portworx.atlassian.net/browse/PA-328
Closes #
Aetoe link : http://aetos.pwx.purestorage.com/resultSet/testSetID/76597

### DIFF
--- a/tests/backup/backup_test.go
+++ b/tests/backup/backup_test.go
@@ -1073,6 +1073,224 @@ var _ = Describe("{BackupAlternatingBetweenLockedAndUnlockedBuckets}", func() {
 	})
 })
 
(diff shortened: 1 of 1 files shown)


mkoppal-px:
I think the test is not verifying the objective of the test case.
The flow should be like this:
Create users
Create backups
Add roles to users
Here roles are - px-backup-infra.admin, px-backup-app.admin and px-backup-app.user
Then you need to provide all levels of access to each user with these roles and validate that irrespective of the user roles, backup share is working

snigdha-px:
Raised a new PR : https://github.com/portworx/torpedo/pull/1038

PR #1033, opened 2023-01-23
Add: New App `postgres-backup` for PX-Backup System Tests

sudas-px:
PX-Backup System Tests were failing as the number of replicas for Postgres was 2.
- Created a new app named `postgres-backup` and updated the replicas to 1.

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -61,9 +61,11 @@ if [ -n "$SKIP_TESTS" ]; then
 fi
 
 FOCUS_ARG=""
+GINKGO_FOCUS=""
 if [ -n "$FOCUS_TESTS" ]; then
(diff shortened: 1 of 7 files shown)


mkoppal-px:
There are a lot of diffs in `deployments/deploy-ssh.sh`
Is it just formatting? I would suggest we keep it as it is.

@lsrinivas-pure Please comment.

Rest of it LGTM 👍

deployments/deploy-ssh.sh
```diff
@@ -1,230 +1,228 @@
 #!/bin/bash -x
 
 if [ -n "${VERBOSE}" ]; then
-    VERBOSE="--v"
+  VERBOSE="--v"
```
kshithijiyer-px, line 4: Do you really need these file changes in for the PR?

sudas-px: Reverted the formatting

kshithijiyer-px (approved):

PR #1048, opened 2023-01-26
Changes to read DASH_UID from build.properties file

CharudathGopal:
Testing Results: https://jenkins.pwx.dev.purestorage.com/job/Torpedo/job/tp-basic-sharedv4-svc/1094/consoleFull

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -4064,11 +4064,40 @@ func ParseFlags() {
 			}
 		}
 
+		/*
+			Get TestSetID based on below precedence
+			1. Check if user has passed in command line and use
+			2. Check if user has set it has an env variable and use
+			3. Check if build.properties available with TestSetID
+			4. If not present create a new one
+		*/
 		val, ok := os.LookupEnv("DASH_UID")
-		if ok {
+		if testsetID != 0 {
+			dash.TestSetID = testsetID
+		} else if ok && (val != "" && val != "0") {
 			testsetID, err = strconv.Atoi(val)
 			if err != nil {
 				log.Warnf("Failed to convert environment testset id  %v to int, err: %v", val, err)
+			}
+			log.Infof(fmt.Sprintf("Using TestSetID: %s set as enviornment variable", val))
+		} else {
+			fileName := "/build.properties"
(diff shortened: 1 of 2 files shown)


tests/common.go
```diff
@@ -4064,12 +4064,41 @@ func ParseFlags() {
...
+			Get TestSetID based on below precedence
+			1. Check if user has passed in command line and use
+			2. Check if user has set it has an env variable and use
+			3. Check if build.properties available with TestSetID
+			4. If not present create a new one
+		*/
 		val, ok := os.LookupEnv("DASH_UID")
-		if ok {
+		if testsetID != 0 {
+			dash.TestSetID = testsetID
```
lsrinivas-pure, line 4076: Since this check is also happening at line 4073, you can remove it at one place

tests/common.go
```diff
@@ -4064,12 +4064,41 @@ func ParseFlags() {
...
+		} else if ok && (val != "" && val != "0") {
 			testsetID, err = strconv.Atoi(val)
 			if err != nil {
 				log.Warnf("Failed to convert environment testset id  %v to int, err: %v", val, err)
 			}
+			log.Infof(fmt.Sprintf("Using TestSetID: %s set as enviornment variable", val))
```
lsrinivas-pure, line 4082: can you make this first line after else if condition , if there is an error this statement might be  invalid I think

tests/common.go
```diff
@@ -4064,12 +4064,41 @@ func ParseFlags() {
...
+		} else {
+			fileName := "/build.properties"
```
lsrinivas-pure, line 4084: how are we copying build.properties inside torpedo contanier?

tests/common.go
CharudathGopal: Have already removed this, its just one time now and if testsetID is not zero we set it as env variable

tests/common.go
CharudathGopal: makes sense

tests/common.go
CharudathGopal: This has to be mounted with the docker command

PR #1055, opened 2023-01-26
Reduced size of fio-fa-clouddrive volume

Pure-AdamuKaapan:
This addresses the issue where HA-update tests take too long to complete as the volume was much larger than it needed to be.

### DIFF
--- a/drivers/scheduler/k8s/specs/fio-fa-cloud-drives/fio.yaml
+++ b/drivers/scheduler/k8s/specs/fio-fa-cloud-drives/fio.yaml
@@ -61,7 +61,7 @@ spec:
       - ReadWriteOnce
       resources:
         requests:
-          storage: 200Gi
+          storage: 10Gi
   - metadata:
       name: fio-log
     spec:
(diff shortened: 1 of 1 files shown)


drivers/scheduler/k8s/specs/fio-fa-cloud-drives/fio.yaml
```diff
@@ -61,7 +61,7 @@ spec:
       - ReadWriteOnce
       resources:
         requests:
-          storage: 200Gi
+          storage: 10Gi
```
jainsmit, line 64: Better to use yaml template to set the size at run time. Basically when using the same spec in system test, it would be good to have bigger size ..

`  {{ if .VolumeSize }}
  storage: "{{ .VolumeSize }}"
  {{ else }}
  storage: "10Gi"{{ end }}`

Pure-AdamuKaapan: Sounds good! Thank you, will fix

jainsmit (approved):
LGTM !!

PR #1058, opened 2023-01-27
Cloud credential names will now be suffixed with timestamp to avoid failures

mkoppal-px:
We are seeing failures like `"failed to retrieve cloud credential [cred-aws]: object not found"` because the cloud credentials are not getting cleanup some times. It is a known issue and will be addressed later. For now, we are working around it by suffixing time stamp

Closes #


snigdha-px (approved):
LGTM

tests/backup/backup_test.go
```diff
@@ -325,7 +325,7 @@ var _ = Describe("{BasicBackupCreation}", func() {
 			bucketNames := getBucketName()
 			for _, provider := range providers {
 				bucketName := fmt.Sprintf("%s-%s", provider, bucketNames[0])
-				cloudCredName = fmt.Sprintf("%s-%s", "cred", provider)
+				cloudCredName = fmt.Sprintf("%s-%s-%v", "cred", provider, time.Now().Unix())
```
kshithijiyer-px, line 328: Can we declare it as a variable and use it this way it will fail in delete?

kshithijiyer-px (changes requested):

kshithijiyer-px (approved):

PR #1061, opened 2023-01-27
Fix scale up scenario for new data services

Madanagopal19:
logs - https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-system-test/job/pds-qa01-system-test/


nikolaypopov (commented):
I am not familiar with PDS tests at all but it looks good to me from a code perspective

nikolaypopov (approved):

PR #1068, opened 2023-01-30
adding CBT pipeline

lsrinivas-pure:

Closes #


Jenkinsfile
```diff
@@ -0,0 +1,15 @@
...
+
+pipeline {
+	agent {
+		label "porx-builder"
+	}
+
+	stages {
+		stage("Run Torpedo CBT") {
+			steps {
+				build(job: "CBT/torpedo-cbt", parameters: [string(name: "TP_GIT_COMMIT", value: GIT_COMMIT)])
```
mborodin-px, line 11: make the job name as a global Jenkins variable, as we can change easily from Jenkins, rather that in the code.
Also use `TORPEDO_GIT_COMMIT` as a variable name

mborodin-px (changes requested):

mborodin-px:
@lsrinivas-pure actually it should be two jobs in this pipeline:
1. build Torpedo which produces `portworx/torpedo:<git_sha>` image
2. run Torpedo test using `portworx/torpedo:<git_sha>`

Jenkinsfile
lsrinivas-pure: > @lsrinivas-pure actually it should be two jobs in this pipeline:
> 
> 1. build Torpedo which produces `portworx/torpedo:<git_sha>` image
> 2. run Torpedo test using `portworx/torpedo:<git_sha>`

@mborodin-px will work on the changes. Just wanted to kow any particular reason it should be 2 jobs?

lsrinivas-pure: fixed

mborodin-px (approved):
LGTM

PR #1085, opened 2023-02-03
Post Execution safe cleanup

mkoppal-px:
Sometimes because of failures during execution, the cleanup tasks are not completed. To be safe we will be deleting all non admin users/groups, backup locations, cloud credentials, backups and restores in the AfterSuite.

Closes #

[Jenkins Run](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test/81/)

### DIFF
--- a/drivers/backup/auth.go
+++ b/drivers/backup/auth.go
@@ -459,7 +459,7 @@ func FetchIDOfUser(userName string) (string, error) {
 			break
 		}
 	}
-
+	log.Infof("Fetching ID of user %s - %s", userName, clientID)
 	return clientID, nil
 }
 
@@ -690,7 +690,8 @@ func AddUser(userName, firstName, lastName, email, password string) error {
 		log.Errorf("%s: %v", fn, err)
 		return err
 	}
-	_, err = processHTTPRequest(method, reqURL, headers, strings.NewReader(string(userBytes)))
+	response, err := processHTTPRequest(method, reqURL, headers, strings.NewReader(string(userBytes)))
+	log.Infof("Response for user [%s] creation - %s", userName, string(response))
 	if err != nil {
 		log.Errorf("%s: %v", fn, err)
 		return err
@@ -716,7 +717,8 @@ func DeleteUser(userName string) error {
 		return err
 	}
 
-	_, err = processHTTPRequest(method, reqURL, headers, nil)
+	response, err := processHTTPRequest(method, reqURL, headers, nil)
+	log.Infof("Response for user [%s] deletion - %s", userName, string(response))
 	if err != nil {
 		return err
 	}
@@ -939,7 +941,8 @@ func AddGroup(group string) error {
 		log.Errorf("%s: %v", fn, err)
 		return err
 	}
-	_, err = processHTTPRequest(method, reqURL, headers, strings.NewReader(string(userBytes)))
+	response, err := processHTTPRequest(method, reqURL, headers, strings.NewReader(string(userBytes)))
+	log.Infof("Creating group response - %s", string(response))
 	if err != nil {
 		log.Errorf("%s: %v", fn, err)
 		return err
@@ -1045,7 +1048,8 @@ func AddGroupToUser(user, group string) error {
 		log.Errorf("%s: %v", fn, err)
 		return err
 	}
-	_, err = processHTTPRequest(method, reqURL, headers, strings.NewReader(string(dataBytes)))
+	response, err := processHTTPRequest(method, reqURL, headers, strings.NewReader(string(dataBytes)))
+	log.Infof("Adding user to group response - %s", string(response))
 	if err != nil {
 		log.Errorf("%s: %v", fn, err)
 		return err
@@ -1068,7 +1072,7 @@ func FetchIDOfGroup(name string) (string, error) {
 			break
 		}
 	}
-
+	log.Infof("Fetching ID of group %s - %s", name, groupID)
 	return groupID, nil
 }
 
--- a/tests/backup/backup_basic_test.go
+++ b/tests/backup/backup_basic_test.go
@@ -174,7 +174,7 @@ var _ = AfterSuite(func() {
 	dash.VerifySafely(err, nil, "Verifying cleaning up of all groups from keycloak")
 	for _, group := range allGroups {
 		if !strings.Contains(group.Name, "admin") && !strings.Contains(group.Name, "app") {
-			backup.DeleteGroup(group.Name)
+			err = backup.DeleteGroup(group.Name)
 			dash.VerifySafely(err, nil, fmt.Sprintf("Verifying group [%s] deletion", group.Name))
 		} else {
 			log.Infof("Group %s was not deleted", group.Name)
@@ -212,7 +212,8 @@ var _ = AfterSuite(func() {
 			return "", false, nil
 		}
 	}
-	task.DoRetryWithTimeout(backupLocationDeletionSuccess, 5*time.Minute, 30*time.Second)
+	_, err = task.DoRetryWithTimeout(backupLocationDeletionSuccess, 5*time.Minute, 30*time.Second)
+	dash.VerifySafely(err, nil, "Verifying backup location deletion success")
 
 	// Cleanup all cloud credentials
 	allCloudCredentials, err := getAllCloudCredentials(ctx)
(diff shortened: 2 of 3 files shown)


tests/backup/backup_basic_test.go
```diff
@@ -146,10 +148,89 @@ var _ = BeforeSuite(func() {
...
+	StartTorpedoTest("Environment cleanup", "Removing Px-Backup entities created during the test execution", nil, 0)
 	defer dash.TestSetEnd()
 	defer EndTorpedoTest()
-	// Cleanup all bucket after suite
+
+	// Cleanup all non admin users
+	ctx, err := backup.GetAdminCtxFromSecret()
+	log.FailOnError(err, "Fetching px-central-admin ctx")
+	allUsers, err := backup.GetAllUsers()
+	dash.VerifySafely(err, nil, "Verifying cleaning up of all users from keycloak")
```
lsrinivas-pure, line 159: Can you check here if these are logging to the dashboard. 
As there no testcase created here, these dash logs wont be posted to dashboard

tests/backup/backup_test.go
```diff
@@ -7830,6 +7830,24 @@ func GetAllBackupsAdmin() ([]string, error) {
 	return backupNames, err
 }
 
+// GetAllBackupsAdmin returns all the backups that px-central-admin has access to
```
lsrinivas-pure, line 7833: correct the function name

lsrinivas-pure (changes requested):

tests/backup/backup_basic_test.go
mkoppal-px: Yes. It is being pushed to [dashboard](http://aetos.pwx.purestorage.com/resultSet/testSetID/88569/testCaseID/348628)

Since it part of `Environment cleanup` test which is part of AfterSuite. Let me know if this is what you are looking for.

tests/backup/backup_test.go
mkoppal-px: Done

drivers/backup/auth.go
```diff
@@ -690,7 +690,8 @@ func AddUser(userName, firstName, lastName, email, password string) error {
 		log.Errorf("%s: %v", fn, err)
 		return err
 	}
-	_, err = processHTTPRequest(method, reqURL, headers, strings.NewReader(string(userBytes)))
+	response, err := processHTTPRequest(method, reqURL, headers, strings.NewReader(string(userBytes)))
+	log.Infof("Response for user [%s] creation - %s", userName, string(response))
```
nikolaypopov, line 694: Do we need to know response for all these commands when they are successful? Should we maybe just add these responses for the failures?

mkoppal-px: Keycloak API never throws an error in case of failure. It returns a 2xx response even if the operation does succeed but returns the error as a string in the response. For us to know whether the API call was successful is to print out the response string and make further decisions based on that. For example, when you try to add a user to a group and it doesn't succeed, we do not get any error, but we get this error message in the response:

```
12:15:55  2023-02-02 06:45:55 +0000:[INFO log.go:#237]  Adding user to group response - {"error":"RESTEASY003210: Could not find resource for full path: [http://pxcentral-keycloak-http.px-backup.svc.cluster.local:80/auth/admin/realms/master/users//groups/7351abb4-2729-4e80-93b8-8ceca68e3259"}](http://pxcentral-keycloak-http.px-backup.svc.cluster.local/auth/admin/realms/master/users//groups/7351abb4-2729-4e80-93b8-8ceca68e3259%22%7D)
```

lsrinivas-pure: how are you handling this failed responses?

mkoppal-px: We are not sure what are the different type of error strings that we may get in the response. The Keycloak API documentation also does not mention about it.
As of now we are just going to print it out.
In the subsequent PR, we will check the error string and do 
`strings.Contains(string(response), "error message")`
If it contains error we will throw an error.

mkoppal-px: We have to do this for all keycloak APIs that we are calling

nikolaypopov: So if it doesn't return error, but you can see that the response has error why don't you do add a wrapper for API response to check if it has error or not and then handle it accordingly? I don't see a point to print response every time for no reason, if you not even getting anything out of that response, these responses are only useful to see when something is not working correctly or throwing an error, otherwise lets not print them, too much of not useful log.

mkoppal-px: As I said, we do not have an official documentation which can let us know what are the different types of error strings that we can expect when there are failures. I agree that it will add a lot of unwanted log entries, but this is the first step in identifying what are the error messages that we can expect. In the subsequent PRs we will handle the error based on the response.
Some of these are legacy functions written long back, we are yet to figure out and apply the error handling logic to all keycloak related APIs which is in our TODO list.

mkoppal-px: If you think we should not have these log entries at all, then I can remove them and add them later when we have everything figured out.
Anyway, adding log entries was not the main object of this PR. It was to add the cleanup functions.

tests/backup/backup_test.go
```diff
@@ -8097,6 +8097,24 @@ func GetAllBackupsAdmin() ([]string, error) {
 	return backupNames, err
 }
 
+// GetAllRestoresAdmin returns all the backups that px-central-admin has access to
+func GetAllRestoresAdmin() ([]string, error) {
+	restoreNames := make([]string, 0)
+	backupDriver := Inst().Backup
+	ctx, err := backup.GetAdminCtxFromSecret()
+	log.FailOnError(err, "Fetching ctx")
```
nikolaypopov, line 8105: Lets not `log.FailOnError()` here, this function returns error, so lets just return it here

tests/backup/backup_test.go
```diff
@@ -8097,6 +8097,24 @@ func GetAllBackupsAdmin() ([]string, error) {
...
+
+	restoreEnumerateRequest := &api.RestoreEnumerateRequest{
+		OrgId: orgID,
+	}
+	restoreResponse, err := backupDriver.EnumerateRestore(ctx, restoreEnumerateRequest)
+	log.FailOnError(err, "Failed to enumerate on restore objects")
```
nikolaypopov, line 8111: Same here, just return error and then handle it with `log.FailOnError()` where you call this func

tests/backup/backup_test.go
```diff
@@ -8177,3 +8195,47 @@ func removeIntItemFromSlice(itemList []int, item []int) []int {
...
+func getAllCloudCredentials(ctx context.Context) (map[string]string, error) {
+	cloudCredentialMap := make(map[string]string, 0)
+	backupDriver := Inst().Backup
+	cloudCredentialEnumerateRequest := &api.CloudCredentialEnumerateRequest{
+		OrgId: orgID,
+	}
+	response, err := backupDriver.EnumerateCloudCredential(ctx, cloudCredentialEnumerateRequest)
+	if err != nil {
+		return nil, nil
+	}
```
nikolaypopov, line 8230: Is there a reason you are not returning error here? Can we at least have some sort of message here like no cloud credentials found or something, if the error means that there were no credentials found in the org

tests/backup/backup_test.go
```diff
@@ -8177,3 +8195,47 @@ func removeIntItemFromSlice(itemList []int, item []int) []int {
...
+func getAllBackupLocations(ctx context.Context) (map[string]string, error) {
+	backupLocationMap := make(map[string]string, 0)
+	backupDriver := Inst().Backup
+	backupLocationEnumerateRequest := &api.BackupLocationEnumerateRequest{
+		OrgId: orgID,
+	}
+	response, err := backupDriver.EnumerateBackupLocation(ctx, backupLocationEnumerateRequest)
+	if err != nil {
+		return nil, nil
+	}
```
nikolaypopov, line 8208: Same here, I think we would need a message here about no backup locations found for the org or something, don't you think?

tests/backup/backup_test.go
mkoppal-px: Done. Fixed it.

tests/backup/backup_test.go
mkoppal-px: Done. Fixed it.

tests/backup/backup_test.go
mkoppal-px: Good catch. Not sure what was I even thinking here. 😄

tests/backup/backup_test.go
mkoppal-px: Done

tests/backup/backup_test.go
mkoppal-px: I will be returning an empty map and the error and handle the error from where we call this func

lsrinivas-pure (approved):
LGTM

PR #1098, opened 2023-02-08
fix:delete backup validation

apimpalgaonkar:
-Added fix for following functions 

**CreateMultipleUsersAndGroups**:
 - Printing username and group name instead of keycloak response
 
**DeleteSharedBackup** :
- deleted backup validation change

**ClusterBackupShareToggle**:
- Validating cluster share with user

Jenkins link - 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/PX-Backup/job/system-tests/job/px-backup-system-test/86/consoleFull

### DIFF
--- a/tests/backup/backup_test.go
+++ b/tests/backup/backup_test.go
@@ -2791,6 +2791,7 @@ var _ = Describe("{ClusterBackupShareToggle}", func() {
 				}
 				_, err = task.DoRetryWithTimeout(clusterShareCheck, 2*time.Minute, 10*time.Second)
 				log.FailOnError(err, "Unable to fetch backup from shared cluster for user %s", username)
+				log.Infof("fetched user backups %v", userBackups)
 
 				restoreName := fmt.Sprintf("%s-%v", RestoreNamePrefix, time.Now().Unix())
 				ValidateSharedBackupWithUsers(username, accessLevel, userBackups[len(userBackups)-1], restoreName)
--- a/tests/backup/backup_test.go
+++ b/tests/backup/backup_test.go
@@ -2783,8 +2783,10 @@ var _ = Describe("{ClusterBackupShareToggle}", func() {
 				log.FailOnError(err, "Fail on Fetching backups for %s", username)
 				if len(userBackups) == 0 {
 					time.Sleep(2 * time.Minute)
+					userBackups, err = GetAllBackupsForUser(username, password)
+					log.FailOnError(err, "Fail on Fetching backups for %s", username)
 				}
-				dash.VerifyFatal(len(userBackups), len(userBackups) > 0, fmt.Sprintf("Verifying backups are shared with user %s", username))
+				dash.VerifyFatal(true, len(userBackups) > 0, fmt.Sprintf("Verifying backups are shared with user %s", username))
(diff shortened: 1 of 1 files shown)


tests/backup/backup_test.go
```diff
@@ -2777,7 +2781,12 @@ var _ = Describe("{ClusterBackupShareToggle}", func() {
 
 				userBackups, err := GetAllBackupsForUser(username, password)
 				log.FailOnError(err, "Fail on Fetching backups for %s", username)
-
+				if len(userBackups) == 0 {
+					time.Sleep(2 * time.Minute)
```
mkoppal-px, line 2785: We need not use `time.Sleep` here. Create a function which gets all the backups for the user and add it to `task.DoRetryWithTimeout`
Set the timeout to 2 min and polling interval to 10 sec.

tests/backup/backup_test.go
```diff
@@ -140,6 +140,8 @@ var _ = Describe("{CreateMultipleUsersAndGroups}", func() {
 	numberOfGroups := 10
 	users := make([]string, 0)
 	groups := make([]string, 0)
+	userValidation := make([]string, 0)
+	groupValidation := make([]string, 0)
```
mkoppal-px, line 144: We don't have to make it so complicated. All you need is one slice each for users and groups. No need for maps.

mkoppal-px:
Use timestamps for user and email strings while creating users. Also append the group name with timestamp.

tests/backup/backup_test.go
mkoppal-px: Resolved

apimpalgaonkar:
Jenkins Run - 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/PX-Backup/job/system-tests/job/px-backup-system-test/92/console

tests/backup/backup_test.go
```diff
@@ -3355,9 +3371,19 @@ var _ = Describe("{DeleteSharedBackup}", func() {
...
+			adminBackups, _ := GetAllBackupsAdmin()
+			log.Infof("%v", adminBackups)
+			adminBackupsMap := make(map[string]bool)
+			for _, backup := range adminBackups {
+				adminBackupsMap[backup] = true
+			}
+			for _, name := range backupNames {
+				if adminBackupsMap[name] {
+					backupNotDeleted = name
+					break
```
mkoppal-px, line 3383: Log the names of the backups not deleted before breaking

mkoppal-px (approved):
LGTM except for a minor comment regarding logging

tests/backup/backup_test.go
apimpalgaonkar: While executing the dash.VerifyFatal process, I am evaluating whether the backupNotDeleted variable contains any data. Regardless of the outcome, the backupNotDeleted information will be displayed there if I'm not wrong , and in the loop as soon I as found first backup which is not deleted I'm breaking the loop.

apimpalgaonkar: In the dash.VerifyFatal procedure, I am checking if the backupNotDeleted variable is empty. In any case, the contents of backupNotDeleted will be printed there.

snigdha-px (approved):
LGTM

PR #1100, opened 2023-02-09
Increase the wait time for the profile upload to take place.  Change …

jrivera-px:
…it from 5 to 20 mins.

The actual time it takes for the upload to happen took longer than 5 mins.   Making it large to avoid this.  If it fails after 20mins there is a real problem.
Closes #


nikolaypopov (approved):
LGTM, let see if it stop failing

PR #1110, opened 2023-02-14
Making changes to time out values to address longevity fails PTX-15633

rvenkatakrishnan1:
For adddressing timeout issues wrt TP-Longevity pipeline job

Closes #PTX-15663


tests/basic/sharedv4_test.go
```diff
@@ -1,5 +1,6 @@
 package tests
 
+/*
```
lsrinivas-pure, line 3: revert this change

lsrinivas-pure (changes requested):

tests/basic/sharedv4_test.go
rvenkatakrishnan1: will do

rvenkatakrishnan1: done

PR #1114, opened 2023-02-15
Enabling bash in Torpedo container

dbhatnagar-px:
This PR enables bash access in Torpedo Container


nikolaypopov (approved):
Please run job that actually uses it first just to make sure this works, otherwise lgtm

lsrinivas-pure (approved):
LGTM

PR #1116, opened 2023-02-16
PTX-15783 - for local PX loadbalancer in EKS use local k8s endpoint

Rohit-PX:

Closes # https://portworx.atlassian.net/browse/PTX-15783

This is a workaround for a known issue with EKS which we have currently not decided to tackle but has a placeholder: https://portworx.atlassian.net/browse/PWX-29184

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -94,6 +94,7 @@ const (
 	pureBlockValue                            = "pure_block"
 	clusterIDFile                             = "/etc/pwx/cluster_uuid"
 	pxReleaseManifestURLEnvVarName            = "PX_RELEASE_MANIFEST_URL"
+	pxServiceLocalEndpoint                    = "portworx-service.kube-system.svc.cluster.local"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -2860,15 +2861,21 @@ func (d *portworx) GetClusterPairingInfo(kubeConfigPath, token string, isPxLBSer
...
 		endpoint, err := d.schedOps.GetServiceEndpoint()
 		if err != nil {
 			return nil, fmt.Errorf("failed to get px service endpoint: %v", err)
 		}
-		address = endpoint
+		if reversePair == true {
+			address = pxServiceLocalEndpoint
+		} else {
+			address = endpoint
+		}
```
mborodin-px, line 2876: you can rewrite this block like this:
```
address = endpoint
if reversePair == true {
	address = pxServiceLocalEndpoint
}
```

Rohit-PX:
```
--- PASS: TestMigration (578.50s)
    --- PASS: TestMigration/testMigrationFailoverFailback (575.47s)
        --- PASS: TestMigration/testMigrationFailoverFailback/vanillaFailoverAndFailbackMigrationTest (575.13s)
PASS
```

PR #1142, opened 2023-02-22
PWX-28766: Pod Usage Telemetry loggly integration testing

ggriffiths:

Closes #

### DIFF
--- a/tests/basic/podmetric_test.go
+++ b/tests/basic/podmetric_test.go
@@ -1,15 +1,20 @@
 package tests
 
 import (
+	"encoding/json"
 	"fmt"
+	"math"
 	"os"
+	"time"
 
 	. "github.com/onsi/ginkgo"
+	"github.com/portworx/torpedo/drivers/node"
 	"github.com/portworx/torpedo/drivers/scheduler"
 	"github.com/portworx/torpedo/pkg/log"
 	rest "github.com/portworx/torpedo/pkg/restutil"
 	"github.com/portworx/torpedo/pkg/testrailuttils"
 	. "github.com/portworx/torpedo/tests"
+	v1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


tests/basic/podmetric_test.go
```diff
@@ -79,10 +120,129 @@ var _ = Describe("{PodMetricFunctional}", func() {
...
+
+	// Count one minute per pod using a PX volume
+	return len(totalPods) * int(meteringInterval.Minutes())
+}
+
+func getLatestPodHours(meteringData []*CallhomeData) float64 {
+	return meteringData[0].PodHour
+}
+
+func verifyPodHourWithError(actualPodHours, expectedPodHours, reasonableErrorPercent float64) {
```
nikolaypopov, line 243: Same here, can we return errors and do all the error checking in the test? Otherwise it looks very weird looking at test code, where half the functions you are using return error and you handle it in test code and the other half do not return anything and are being handled inside the actual function. We should pick one approach and stick to it. In my opinion all functions should return errors and then handled by the test itself, what do you think?

ggriffiths: That sounds good to me, I wasn't aware of any existing practices here. Fine with that.

tests/basic/podmetric_test.go
```diff
@@ -41,12 +38,58 @@ var _ = Describe("{PodMetricFunctional}", func() {
...
+				for _, md := range meteringData {
+					dash.VerifyFatal(md.ClusterUUID, clusterUUID, "this cluster should have data now")
+				}
+
+				log.InfoD("Check pod hours is correct")
+				expectedAppPodHours, err := getExpectedPodHours(contexts, meteringInterval)
+				log.FailOnError(err, "Failed to get expectedAppPodHours")
+				expectedPodHours := (float64(expectedAppPodHours) / 60) + initialPodHours
+				log.InfoD("Estimated pod hours for this app is %v", expectedPodHours)
+				log.InfoD("Estimated total pod hours is %v", expectedPodHours)
```
nikolaypopov, line 87: Remove one of these prints, its same thing, unless you mean to print something different here

nikolaypopov (approved):
LGTM

PR #1145, opened 2023-02-23
Enabling Consul Workload (Approach taken with static images) 

dbhatnagar-px:
This PR enables Consul Workload through pre-cooked Consul agent and Consul bench images

Jenkins Run: https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/124/


drivers/pds/lib/utils.go
```diff
@@ -1074,6 +1076,131 @@ func RunTpccWorkload(dbUser string, pdsPassword string, dnsEndpoint string, dbNa
...
+	}
+	log.InfoD("Going to Trigger Consul Bench Workload for the Deployment")
+	deployment, err := k8sApps.CreateDeployment(deploymentSpec, metav1.CreateOptions{})
+	if err != nil {
+		log.Errorf("An Error Occured while creating deployment %v", err)
+		return nil, err
+	}
+	err = k8sApps.ValidateDeployment(deployment, timeOut, timeInterval)
+	if err != nil {
+		log.Errorf("An Error Occured while validating the pod %v", err)
```
lsrinivas-pure, line 1195: no need to log error  here as you are anyway logging in the function calling this

drivers/pds/lib/utils.go
```diff
@@ -1074,6 +1076,131 @@ func RunTpccWorkload(dbUser string, pdsPassword string, dnsEndpoint string, dbNa
...
+		return nil, err
+	}
+
+	time.Sleep(1 * time.Minute)
```
lsrinivas-pure, line 1199: any specific reason for this wait

lsrinivas-pure (changes requested):

drivers/pds/lib/utils.go
dbhatnagar-px: Hi Leela, actually it's required to log here so that we know if this failed in Creation of deployment (line: 1190) or validating the deployment (line:1195). The calling function would just give out an error saying this function failed - these micro errors would tell us the exact place where it failed.

drivers/pds/lib/utils.go
dbhatnagar-px: Here we are simply running this workload for 1 minute

drivers/pds/lib/utils.go
lsrinivas-pure: since error message is unique, will that not help?
Im asking this because we want to maintain uniformity across all the tests

drivers/pds/lib/utils.go
lsrinivas-pure: can you please add a comment, the reason for waiting

lsrinivas-pure (changes requested):

drivers/pds/lib/utils.go
dbhatnagar-px: Done

drivers/pds/lib/utils.go
dbhatnagar-px: Done

lsrinivas-pure (approved):
LGTM

Madanagopal19 (approved):
LGTM

PR #1151, opened 2023-02-23
PWX-28696: Adds test validating that FACD topology drives are on the correct nodes

Pure-AdamuKaapan:
For the FlashArray Cloud Drive topology feature, we need to validate that cloud drives are actually provisioned on the proper arrays that match their zone. The test looks approximately like this:
1. Parse `px-pure-secret` to get the labels for each array
2. Get all PX nodes
3. For each PX node:
  a) Get drive set for node
  b) For each drive config in drive set:
    i) Check "FA management endpoint" label and use this to get the array zone (using the map from step 1)
    ii) Compare FA zone label to node zone label: if they differ, fail test


tests/pure_types.go
```diff
@@ -2,9 +2,18 @@ package tests
 
 import (
 	"encoding/json"
-	"github.com/portworx/torpedo/pkg/log"
 
 	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/portworx/torpedo/pkg/log"
+)
```
lsrinivas-pure, line 8: can we move this file to pkg or drivers as this is not a test file?

tests/common.go
```diff
@@ -552,6 +552,74 @@ func ValidateContext(ctx *scheduler.Context, errChan ...*chan error) {
...
 }
 
+func ValidatePureCloudDriveTopologies(errChan ...*chan error) {
+	defer func() {
+		if len(errChan) > 0 {
+			close(*errChan[0])
+		}
+	}()
+
+	ginkgo.Describe("For validation of FlashArray Cloud Drive topologies", func() {
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: Nope, I was just copying from one of the tests above. If I can go without then I'll gladly just make it into one continuous function.

tests/basic/license_test.go
```diff
@@ -23,7 +23,7 @@ type Label string
 const (
 	defaultReadynessTimeout = 2 * time.Minute
 
-	pureSecretNamespace = "kube-system"
+	PureSecretNamespace = "kube-system"
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: It will not, it will be in the same namespace as the StorageCluster/Portworx.

tests/pure_types.go
Pure-AdamuKaapan: Good call, yes

tests/common.go
```diff
@@ -552,6 +552,74 @@ func ValidateContext(ctx *scheduler.Context, errChan ...*chan error) {
 	})
 }
 
+func ValidatePureCloudDriveTopologies(errChan ...*chan error) {
```
nikolaypopov, line 555: Let's have a return for functions so we can then later on decide how we want to treat it in the test

tests/pure_types.go
```diff
@@ -57,8 +83,8 @@ func GetSecret(output interface{}, name, namespace string) error {
 	var pureConnectionJSON []byte
 	var ok bool
 
-	if pureConnectionJSON, ok = secret.Data[pureJSONKey]; !ok {
-		log.Errorf("Secret '%s' is missing field '%s'", name, pureJSONKey)
+	if pureConnectionJSON, ok = secret.Data[PureJSONKey]; !ok {
+		log.Errorf("Secret '%s' is missing field '%s'", name, PureJSONKey)
 		return err
```
nikolaypopov, line 88: Let's not `log.Errorf()`, let's just return verbose error

tests/common.go
```diff
@@ -552,6 +552,74 @@ func ValidateContext(ctx *scheduler.Context, errChan ...*chan error) {
...
+		processError(err, errChan...)
+		nodesMap := node.GetNodesByName()
+
+		driverNamespace, err := Inst().V.GetVolumeDriverNamespace()
+		processError(err, errChan...)
+
+		pxPureSecret, err := GetPXPureSecret(driverNamespace)
+		processError(err, errChan...)
+
+		endpointToZoneMap := pxPureSecret.GetArrayToZoneMap()
```
nikolaypopov, line 573: Can this possible be empty?

pkg/pureutils/pure_types.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (p *PXPureSecret) GetArrayToZoneMap() map[string]string {
+	arrayEndpointToZoneMap := map[string]string{}
+	for _, array := range p.Arrays {
+		if array.Labels == nil || len(array.Labels) == 0 {
+			arrayEndpointToZoneMap[array.MgmtEndPoint] = ""
+			continue
+		}
+		zoneLabel, ok := array.Labels[PureSecretArrayZoneLabel]
```
pure-jliao, line 55: nit: this should work for a nil map as well, so condition above can be removed.

pkg/pureutils/pure_types.go
```diff
@@ -0,0 +1,96 @@
+package pureutils
```
siyingjin, line 1: Why did you move this file from test path to pureutils. Are we going to use it anywhere outside of tests?

tests/common.go
```diff
@@ -552,6 +552,71 @@ func ValidateContext(ctx *scheduler.Context, errChan ...*chan error) {
...
+
+		driveSet, err := Inst().V.GetDriveSet(&nodeFromMap)
+		if err != nil { return err }
+
+		for configID, driveConfig := range driveSet.Configs {
+			err = nil
+			if driveConfig.Labels == nil || len(driveConfig.Labels) == 0 {
+				return fmt.Errorf("drive config %s has no labels: validate that you're running on PX master or 3.0+ and using FlashArray cloud drives with topology enabled", configID)
+			}
+
```
pure-jliao, line 600: similar here, can be removed.

siyingjin (commented):
Makes sense to me overall. I'm interested to see an example test result.

pure-jliao (commented):
overall LGTM

tests/common.go
```diff
@@ -552,6 +552,71 @@ func ValidateContext(ctx *scheduler.Context, errChan ...*chan error) {
...
+			if arrayEndpoint, ok = driveConfig.Labels[pureutils.CloudDriveFAMgmtLabel]; !ok {
+				return fmt.Errorf("drive config %s is missing the '%s' label: validate that you're running PX master or 3.0+ and using FlashArray cloud drives with topology enabled", configID, pureutils.CloudDriveFAMgmtLabel)
+			}
+
+			var driveZone string
+			if driveZone, ok = endpointToZoneMap[arrayEndpoint]; !ok {
+				return fmt.Errorf("drive config %s is on array with endpoint '%s', which is not listed in px-pure-secret", configID, arrayEndpoint)
+			}
+
+			if driveZone != nodeZone {
```
dsupure, line 611: It looks to me we can only run positive tests today and we can only use Jenkins to test negative tests. In our test plan confluence wiki, please add one section to describe how we do the negative tests, so we know how to add/modify negative tests when needed.

pkg/pureutils/pure_types.go
Pure-AdamuKaapan: It was a code review comment from above. I think the goal is to keep more "utility function" code in `pkg` while keeping the `tests` folder for actual test code/logic?

tests/common.go
Pure-AdamuKaapan: Added an "additional testing notes" section here: https://portworx.atlassian.net/wiki/spaces/PE/pages/2596339877/FACD+CSI+Topology+Test+Plan

pkg/pureutils/pure_types.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/portworx/sched-ops/k8s/core"
+)
+
+const (
+	PureSecretName   = "px-pure-secret"
+	PureS3SecretName = "px-pure-secret-fbs3"
+	PureJSONKey      = "pure.json"
+
+	PureSecretArrayZoneLabel = "topology.portworx.io/zone"
```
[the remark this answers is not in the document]
Pure-AdamuKaapan: We don't seem to have a PD ticket for FACD CSI topology yet. Should I create that now, or wait until we get closer? Either way I'll be sure to note it, it's a pretty big requirement in my mind so I won't forget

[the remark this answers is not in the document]
dsupure: Let's create a PD for the FACD CSI topology and add note saying it only support zone.

dsupure (approved):
LGTM!

pure-jliao (approved):
seems not changed a lot since last time, so LGTM

Pure-AdamuKaapan:
Closing in lieu of #1189

PR #1159, opened 2023-02-28
PWX-29380 Add configurable metering interval

dahuang-purestorage:

PWX-29380

### DIFF
--- a/tests/basic/podmetric_test.go
+++ b/tests/basic/podmetric_test.go
@@ -5,9 +5,12 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"strconv"
 	"time"
 
+	optest "github.com/libopenstorage/operator/pkg/util/test"
 	. "github.com/onsi/ginkgo"
+	"github.com/portworx/sched-ops/k8s/operator"
 	"github.com/portworx/torpedo/drivers/node"
 	"github.com/portworx/torpedo/drivers/scheduler"
 	"github.com/portworx/torpedo/pkg/log"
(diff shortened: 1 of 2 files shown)


tests/basic/podmetric_test.go
```diff
@@ -24,11 +27,14 @@ var _ = Describe("{PodMetricFunctional}", func() {
 	var testrailID, runID int
 	var contexts []*scheduler.Context
 	var namespacePrefix string
+	var meteringIntervalString = os.Getenv("PODMETRIC_METERING_INTERVAL_MINUTES")
+	var callHomeIntervalString = os.Getenv("PODMETRIC_METERING_INTERVAL_MINUTES")
```
ggriffiths, line 31: is it intentional for these to be the same env var?

dahuang-purestorage: yes. i guess in that case, i can load once and assign the var to the other?

ggriffiths: No this is good, maybe just include a comment on why we're setting the metering/callhome interval to be the same. Others looking at this later on might see it as a bug.

ggriffiths (approved):

tests/basic/podmetric_test.go
```diff
@@ -263,3 +278,36 @@ func verifyPodHourWithError(actualPodHours, expectedPodHours, reasonableErrorPer
 
 	return nil
 }
+
+// updateStorageSpecRuntimeOpts updates the storageSpec's loggly callhome interval and
+// metering interval. Finally, restarts all PX pods and checks its condition.
+func updateStorageSpecRuntimeOpts(callhomeInterval string, meteringInterval string) {
```
nikolaypopov, line 284: I am always voting for funcs to return things and then handle return in the test, but it seems there are many things that are going away from it, it just looks better when you go through test to see that we handling returned errors/ojects properly, rather than looking at every function that that doesn't return

ggriffiths: Thanks for noticing this @nikolaypopov, I've addressed this for Daniel since he's out.

tests/basic/podmetric_test.go
```diff
@@ -70,9 +79,16 @@ var _ = Describe("{PodMetricFunctional}", func() {
 				time.Sleep(waitDuration)
 
 				log.InfoD("Check metering data is accurate")
-				meteringData, err = getMeteringData(clusterUUID, meteringInterval)
-				log.FailOnError(err, "Failed to get metering data")
-				existsData = len(meteringData) > 0
+
+				// try to get non-empty metering data for 3 minutes
+				Eventually(func() bool {
+					meteringData, err = getMeteringData(clusterUUID, meteringInterval)
```
lsrinivas-pure, line 85: What happens if this is not ture eventually, will it still execute next lines.
If not can you use our internal method `task.DoRetryWithTimeout`

ggriffiths: Eventually is a ginkgo built-in function that will retry until a function is true.

The test will fail if that is not the case, which is the goal here

nikolaypopov (approved):
LGTM

PR #1169, opened 2023-03-03
PTX-15161 -cloud drive add of same size/types should not create seperate pools

gpal1:

During addition of cloud drive (of same size and type) to the existing pools with PX restart - no new pool should get created.

Added a check to verify total pool count after addition of cloud drive 

Closes #
PTX-16216


lsrinivas-pure (approved):
LGTM

vprabhakar-px (approved):

PR #1170, opened 2023-03-03
Enabling Random name of PV/PVC/Independent App and also pre-check to delete all resources if they are present

dbhatnagar-px:
This PR takes care of any duplicate PV/PVC/Independent app that may already been running in the system. Also just in case they are running, this PR makes sure to delete them before proceeding ahead with the TC


gauraa2 (approved):
LGTM

drivers/pds/lib/utils.go
```diff
@@ -1310,7 +1362,7 @@ func CreateIndependentPV(name string) (*corev1.PersistentVolume, error) {
 			},
 		},
 	}
-	pv, err := k8sCore.CreatePersistentVolume(pv)
+	pv, err = k8sCore.CreatePersistentVolume(pv)
```
lsrinivas-pure, line 1365: can u delete the below error logging?

dbhatnagar-px: Done

drivers/pds/lib/utils.go
```diff
@@ -1286,8 +1292,54 @@ func RegisterClusterToControlPlane(controlPlaneUrl, tenantId, clusterType string
...
+// Check if PV and associated PVC is still present. If yes then delete both of them
+func CheckAndDeleteIndependentPV(name string) error {
+	pv_check, err := k8sCore.GetPersistentVolume(name)
+	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
+			return nil
+		}
+		return err
+	}
+	log.InfoD("Stranded PV Found. Going ahead to delete this PV and associated entities")
```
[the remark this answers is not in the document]
dbhatnagar-px: Done

PR #1174, opened 2023-03-07
PWX-28724 : Update OKE automation to use disk encryption

nikita-bhatia:
Add missing env variable "INSTANCE_GROUP" required for torpedo automation.

Tested with the change done, value is being set on torpedo pod and jenkins build passed.


vinayakshnd (approved):
LGTM

PR #1181, opened 2023-03-08
New Test scripts for ResyncFailedPoolOutOfRebalance

vprabhakar-px:

Closes #

### DIFF
--- a/drivers/node/node_registry.go
+++ b/drivers/node/node_registry.go
@@ -91,13 +91,14 @@ func GetStorageDriverNodes() []Node {
 
 // IsStorageNode returns true if the node is a storage node, false otherwise
 func IsStorageNode(n Node) bool {
-	return len(n.StoragePools) > 0
+	return len(n.Pools) > 0
 }
 
 // GetStorageNodes gets all the nodes with non-empty StoragePools
 func GetStorageNodes() []Node {
 	var nodeList []Node
-	for _, n := range nodeRegistry {
+	storageDriverNodes := GetStorageDriverNodes()
+	for _, n := range storageDriverNodes {
 		if IsStorageNode(n) {
 			nodeList = append(nodeList, n)
 		}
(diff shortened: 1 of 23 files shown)


tests/basic/storage_pool_test.go
```diff
@@ -7923,3 +7923,94 @@ var _ = Describe("{DiffPoolExpansionFromMaintenanceNode}", func() {
...
+			dash.VerifyFatal(resizeErr, nil,
+				fmt.Sprintf("Verify pool %s on expansion using auto option", poolUUID))
+		}
+		log.InfoD("Validate pool rebalance after drive add")
+		err = ValidatePoolRebalance(*nodeDetail, poolToBeResized.ID)
+
+		// Validate Volume resync if any volume got in to resync mode
+		for _, eachContext := range contexts {
+			vols, err := Inst().S.GetVolumes(eachContext)
+			log.FailOnError(err, "Failed to get volumes from context")
```
lsrinivas-pure, line 7990: can you add check if vol has already repl3 , if so you can make it 2 and make it 3 again.
Also you can make this operation only on volume which has "poolUUID" as one of its replica

tests/common.go
```diff
@@ -1726,14 +1726,15 @@ func GetStorageNodes() ([]node.Node, error) {
...
-
-	for _, node := range nodes {
-		devices, err := Inst().V.GetStorageDevices(node)
-		if err != nil {
-			return nil, err
-		}
-		if len(devices) > 0 {
-			storageNodes = append(storageNodes, node)
+	for _, eachNode := range nodes {
+		if node.IsStorageNode(eachNode) {
```
lsrinivas-pure, line 1730: this seems like a redundant method, can you delete this and replace it with node.GetStorageNodes

vprabhakar-px: This is not redundant , I am modifying already available function , which is GetStorageNodes() , 
I am basically checking if the node is storageNode , and then proceeding

tests/basic/storage_pool_test.go
vprabhakar-px: done

drivers/node/node_registry.go
```diff
@@ -217,3 +217,13 @@ func GetNodeDetailsByNodeName(nodeName string) (Node, error) {
 	}
 	return Node{}, fmt.Errorf("failed to get Node Details by Node Name [%s] ", nodeName)
 }
+
+// GetNodeUuidByName get node uuid for a given node name
+func GetNodeUuidByName(nodeName string) (string, error) {
+	for _, n := range nodeRegistry {
+		if n.Name == nodeName {
+			return n.uuid, nil
```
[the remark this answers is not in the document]
vprabhakar-px: already deleted

lsrinivas-pure (approved):
lgtm

PR #1187, opened 2023-03-09
PA-603: Add test for BackupSync

kshithijiyer-px:
- [x] Add test for Backupsync
- [x] Change `DeleteBackupLocation()` to take parameter for delete backups 

Closes # PA-603

Run results:
Dashboard URL : http://aetos.pwx.purestorage.com/resultSet/testSetID/121202 (Only the testcase added in this PR.)

A bunch of commits in the PR don't worry will squash and merge.


kshithijiyer-px:
While running the job I saw that there were only 45 of 51 backups synced hence increasing the timeout and trying.

kshithijiyer-px:
Still waiting for one run of all the cases, reset is ready for review please feel free to comment.

tests/backup/backup_restore_basic_test.go
```diff
@@ -895,3 +896,197 @@ var _ = Describe("{CustomResourceRestore}", func() {
...
+			appContexts = ScheduleApplications(taskName)
+			contexts = append(contexts, appContexts...)
+			for _, ctx := range appContexts {
+				ctx.ReadinessTimeout = appReadinessTimeout
+				namespace := GetAppNamespace(ctx, taskName)
+				bkpNamespaces = append(bkpNamespaces, namespace)
+			}
+		}
+	})
+	It("validate that the backup sync syncs all the backups present in bucket", func() {
```
mkoppal-px, line 937: nit: Can you make `v` in `validate` upper case?

tests/backup/backup_restore_basic_test.go
```diff
@@ -895,3 +896,197 @@ var _ = Describe("{CustomResourceRestore}", func() {
...
+			// Wait for backups to get synced
+			for i := 0; i < 1000; i++ {
+				fetchedBackupNames, _ := GetAllBackupsAdmin()
+				log.Infof(fmt.Sprintf("Fetched backups %d", len(fetchedBackupNames)))
+				if len(fetchedBackupNames) == len(backupNames) {
+					break
+				} else {
+					time.Sleep(30)
+				}
+			}
```
mkoppal-px, line 1039: We can use the ready-made `task.DoRetryWithTimeout` function here

kshithijiyer-px: I tried using it when I try to using this but the problem here is there is no way to pass external variables when you take this approach. Here I am mostly using backupNames which is picked up earlier and then using it  to compare with the newer fetched value. 
```
backupLocationDeleteStatusCheck := func() (interface{}, bool, error) {
```

tests/backup/backup_restore_basic_test.go
```diff
@@ -895,3 +896,197 @@ var _ = Describe("{CustomResourceRestore}", func() {
...
+		dash.VerifySafely(err, nil, fmt.Sprintf("Deleting cluster %s", destinationClusterName))
+
+		log.Infof("Cleaning up backup location - %s", customBackupLocationName)
+		err = DeleteBackupLocation(customBackupLocationName, backupLocationUID, orgID, true)
+		dash.VerifySafely(err, nil, fmt.Sprintf("Deleting backup location %s", customBackupLocationName))
+		log.Infof("Cleaning cloud credential")
+		//TODO: Eliminate time.Sleep
+		time.Sleep(time.Minute * 5)
+		err = DeleteCloudCredential(credName, orgID, cloudCredUID)
+		dash.VerifySafely(err, nil, fmt.Sprintf("Deleting cloud cred %s", credName))
```
mkoppal-px, line 1090: Please use `CleanupCloudSettingsAndClusters` function which does all the required cleanup

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
```diff
@@ -895,3 +896,197 @@ var _ = Describe("{CustomResourceRestore}", func() {
...
+			}
+			log.Infof("List of backups - %v", backupNames)
+		})
+
+		Step("Remove the backup location where backups were taken", func() {
+			log.InfoD("Remove backup location where backups were taken")
+			err := DeleteBackupLocation(customBackupLocationName, backupLocationUID, orgID, false)
+			dash.VerifySafely(err, nil, fmt.Sprintf("Deleting backup location %s", customBackupLocationName))
+			//TODO: Eliminate time.Sleep
+			time.Sleep(time.Minute * 3)
```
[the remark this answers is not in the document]
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
mkoppal-px: Here you can keep enumerating the backups and check until the length is what is expected. After that you can come out and proceed. Something like

```
backupCountCheck = func() (interface{}, bool, error) {
    fetchedBackupNames, _ := GetAllBackupsAdmin()
    og.Infof(fmt.Sprintf("Fetched backups %d", len(fetchedBackupNames)))
				if len(fetchedBackupNames) == len(backupNames) {
					return "", false, nil
				} else {
					return "", true, fmt.Errorf("Expected length was %d but got %d", len(backupNames), len(fetchedBackupNames))
				}
}

_, err := task.DoRetryWithTimeout(backupCountCheck, Timeout, RetryTime)

```

tests/backup/backup_restore_basic_test.go
```diff
@@ -895,3 +896,197 @@ var _ = Describe("{CustomResourceRestore}", func() {
...
+				log.Infof(fmt.Sprintf("Fetched backups %d", len(fetchedBackupNames)))
+				if len(fetchedBackupNames) == len(backupNames) {
+					break
+				} else {
+					time.Sleep(30)
+				}
+			}
+			fetchedBackupNames, err := GetAllBackupsAdmin()
+			log.FailOnError(err, "Getting a list of all backups")
+			dash.VerifyFatal(len(fetchedBackupNames), len(backupNames), "Comparing the expected and actual number of backups")
```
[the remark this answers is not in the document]
kshithijiyer-px: Ideally It would be better if all the other cases cleanup their backups, the problem here which can happen is that we would be able to validate if backupsync is able to pull in all the backups or not. And this thing might not happen as well because by default  `DeleteBackupLocation()` would run with true and remove the backups present in the bucket.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
```diff
@@ -895,3 +897,206 @@ var _ = Describe("{CustomResourceRestore}", func() {
...
+				backupNames = append(backupNames, backupName)
+			}
+		})
+
+		Step("Check if all backups are synced or not", func() {
+			log.InfoD("Check if backups created before are synced or not")
+
+			// Wait for backups to get synced
+			for i := 0; i < 1000; i++ {
+				fetchedBackupNames, _ := GetAllBackupsAdmin()
```
mkoppal-px, line 1049: Even here we should not be getting all admin backups because there may be remnant backups from previous tests which will case failure.

tests/backup/backup_restore_basic_test.go
mkoppal-px: I tried to see if anonymous functions can access variables not in its scope and it should work - https://go.dev/play/p/_GIdpQ89Lsp

Here you can access `backupNames` even without passing it to the function.

kshithijiyer-px: I get your point that it's a better way and initially in my first commit was trying to do the same but it still fails. 

Dashboard URL : http://aetos.pwx.purestorage.com/resultSet/testSetID/121988

kshithijiyer-px: Thank @mkoppal-px done and removed the loop

tests/backup/backup_locked_bucket_test.go
```diff
@@ -178,7 +178,7 @@ var _ = Describe("{BackupAlternatingBetweenLockedAndUnlockedBuckets}", func() {
 
 		log.InfoD("Deleting backup location and cloud setting")
 		for backupLocationUID, backupLocationName := range BackupLocationMap {
-			err := DeleteBackupLocation(backupLocationName, backupLocationUID, orgID)
+			err := DeleteBackupLocation(backupLocationName, backupLocationUID, orgID, false)
```
snigdha-px, line 181: Here should it be err := DeleteBackupLocation(backupLocationName, backupLocationUID, orgID, true)?

kshithijiyer-px: @snigdha-px It's a locked bucket you can't remove backups before 2 days.

snigdha-px: ok

snigdha-px (approved):
LGTM

tests/backup/backup_restore_basic_test.go
```diff
@@ -1068,3 +1069,209 @@ var _ = Describe("{AllNSBackupWithIncludeNewNSOption}", func() {
...
+				}
+				resp, err := backupDriver.InspectBackup(ctx, backupInspectRequest)
+				log.FailOnError(err, "Inspect each backup from list")
+				actual := resp.GetBackup().GetStatus().Status
+				expected := api.BackupInfo_StatusInfo_Success
+				dash.VerifyFatal(actual, expected, "Check each backup for success status")
+			}
+		})
+	})
+	JustAfterEach(func() {
```
kphalgun-px, line 1256: Are the backups we created being deleted?

tests/backup/backup_restore_basic_test.go
```diff
@@ -1068,3 +1069,209 @@ var _ = Describe("{AllNSBackupWithIncludeNewNSOption}", func() {
...
+			log.Infof("List of backups - %v", backupNames)
+		})
+
+		Step("Remove the backup location where backups were taken", func() {
+			log.InfoD("Remove backup location where backups were taken")
+			// Issue a remove backup location call
+			err := DeleteBackupLocation(customBackupLocationName, backupLocationUID, orgID, false)
+			dash.VerifySafely(err, nil, fmt.Sprintf("Deleting backup location %s", customBackupLocationName))
+
+			// Wait until backup location is remove or not
```
mkoppal-px, line 1179: nit: It can be `Wait until backup location is removed`

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Yes. If you give `DeleteBackupLocation()` with true it will remove all the backups.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
```diff
@@ -1068,3 +1069,209 @@ var _ = Describe("{AllNSBackupWithIncludeNewNSOption}", func() {
...
+			backupLocationDeleteStatusCheck := func() (interface{}, bool, error) {
+				ctx, err := backup.GetAdminCtxFromSecret()
+				status, err := IsBackupLocationPresent(customBackupLocationName, ctx, orgID)
+				if err != nil {
+					return "", true, fmt.Errorf("backup location %s still present with error %v", customBackupLocationName, err)
+				}
```
mkoppal-px, line 1185: We have to add `log.FailOnError` here

tests/backup/backup_restore_basic_test.go
```diff
@@ -1068,3 +1069,209 @@ var _ = Describe("{AllNSBackupWithIncludeNewNSOption}", func() {
...
+			}
+		})
+
+		Step("Check if all backups are synced or not", func() {
+			log.InfoD("Check if backups created before are synced or not")
+
+			// Wait for backups to get synced
+			checkBackupSync := func() (interface{}, bool, error) {
+				fetchedBackupNames, err := GetAllBackupsAdmin()
+				if err != nil {
```
mkoppal-px, line 1222: Here again we need `log.FailOnError`

tests/backup/backup_restore_basic_test.go
```diff
@@ -1068,3 +1069,209 @@ var _ = Describe("{AllNSBackupWithIncludeNewNSOption}", func() {
...
+			dash.VerifySafely(err, nil, fmt.Sprintf("Verify destroying app %s, Err: %v", taskName, err))
+		}
+		ctx, err := backup.GetAdminCtxFromSecret()
+		log.FailOnError(err, "Fetching px-central-admin ctx")
+
+		log.Infof("Deleting registered clusters for admin context")
+		err = DeleteCluster(SourceClusterName, orgID, ctx)
+		dash.VerifySafely(err, nil, fmt.Sprintf("Deleting cluster %s", SourceClusterName))
+		err = DeleteCluster(destinationClusterName, orgID, ctx)
+		dash.VerifySafely(err, nil, fmt.Sprintf("Deleting cluster %s", destinationClusterName))
```
mkoppal-px, line 1273: This is also taken care in `CleanupCloudSettingsAndClusters`

mkoppal-px:
I think we need to cleanup the backups created during this test

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Why so it's inside `backupLocationDeleteStatusCheck := func()`? Ideally it's a function and the instead of returning we are going back to directly erroring out. I would suggest to return instead of using `log.FailOnError` to stay consistent.

tests/common.go
```diff
@@ -2713,13 +2713,13 @@ func DeleteCluster(name string, orgID string, ctx context1.Context) error {
 }
 
 // DeleteBackupLocation deletes backup location
-func DeleteBackupLocation(name string, backupLocationUID string, orgID string) error {
+func DeleteBackupLocation(name string, backupLocationUID string, orgID string, DeleteExistingBackups bool) error {
```
kphalgun-px, line 2716: To ensure consistency, perhaps we could consider renaming it to "backupLocationName"? Although it may not be of significant importance.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Why so it's inside backupLocationDeleteStatusCheck := func()? Ideally it's a function and the instead of returning we are going back to directly erroring out. I would suggest to return instead of using log.FailOnError to stay consistent.

tests/backup/backup_restore_basic_test.go
```diff
@@ -984,7 +985,7 @@ var _ = Describe("{AllNSBackupWithIncludeNewNSOption}", func() {
 					ctx.ReadinessTimeout = appReadinessTimeout
 					namespace := GetAppNamespace(ctx, taskName)
 					log.InfoD("Scheduled application with namespace [%s]", namespace)
-					// appNamespaces in this scenario is a of newly created namespaces
+					// appNamespaces in this scenario is of newly created namespaces
```
kphalgun-px, line 988: Thanks for correcting it.

tests/backup/backup_restore_basic_test.go
mkoppal-px: But this is not an error from the anonymous function. This error is coming from `IsBackupLocationPresent`. If we are getting an error from `IsBackupLocationPresent` it means the API call has failed and there is some other issue which is coming from `EnumerateBackupLocation` call. We should fail immediately. But here if we say 

```
if err != nil {
					return "", true, fmt.Errorf("backup location %s still present with error %v", customBackupLocationName, err)
				}
```
We are basically saying that we know there is an error in the `EnumerateBackupLocation` and yet we want to keep retrying which would not be correct.

tests/common.go
kshithijiyer-px: This is outside the scope of this PR we can do it in the next PR.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: What if the next retry works in this case we will be erroring out in the first instance itself.

tests/backup/backup_restore_basic_test.go
kphalgun-px: Since "DeleteExistingBackups" was already set to true in "DeleteBackupLocation," deleting backups beforehand in JustAfterEach of certain test cases may be somewhat redundant.

kshithijiyer-px: Done

kshithijiyer-px:
> I think we need to cleanup the backups created during this test

It's done as a part of BackupLocation deletion as we are giving the API with true.

kphalgun-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #1188, opened 2023-03-09
PWX-29644: Reduced size of FA CD fio app

Pure-AdamuKaapan:
During nightly CI, repl update is taking just a tiiiiiny bit too long. Reducing the size and config of the fio-fa-cloud-drive app so that it will finish sooner.


dsupure (approved):
LGTM!

PR #1189, opened 2023-03-09
PWX-28696: Adds test validating that FACD topology drives are on the correct nodes

Pure-AdamuKaapan:
Re-open of #1151 but with a branch from origin.

For the FlashArray Cloud Drive topology feature, we need to validate that cloud drives are actually provisioned on the proper arrays that match their zone. The test looks approximately like this:
1. Parse `px-pure-secret` to get the labels for each array
2. Get all PX nodes
3. For each PX node:
  a) Get drive set for node
  b) For each drive config in drive set:
    i) Check "FA management endpoint" label and use this to get the array zone (using the map from step 1)
    ii) Compare FA zone label to node zone label: if they differ, fail test


nikolaypopov (approved):
LGTM

PR #1190, opened 2023-03-10
Fix for Upgrade DataService Version/Image

Madanagopal19:
aetos logs : 
http://aetos.pwx.purestorage.com/resultSet/testSetID/118763 - ImageUpgrade
http://aetos.pwx.purestorage.com/resultSet/testSetID/118127 - VersionUpgrade

### DIFF
--- a/drivers/pds/lib/utils.go
+++ b/drivers/pds/lib/utils.go
@@ -630,7 +630,9 @@ func GetVersionsImage(dsVersion string, dsBuild string, dataServiceID string) (s
 	isVersionAvailable = false
 	isBuildAvailable = false
 	for i := 0; i < len(versions); i++ {
-		if (*versions[i].Enabled) && (*versions[i].Name == dsVersion) {
+		log.Debugf("version name %s and is enabled=%t", *versions[i].Name, *versions[i].Enabled)
+		if *versions[i].Name == dsVersion {
+			log.Debugf("DS Version %s is enabled in the control plane", dsVersion)
 			images, _ = components.Image.ListImages(versions[i].GetId())
 			for j := 0; j < len(images); j++ {
 				if *images[j].Build == dsBuild {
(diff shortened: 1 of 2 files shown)


drivers/pds/lib/utils.go
```diff
@@ -630,7 +630,9 @@ func GetVersionsImage(dsVersion string, dsBuild string, dataServiceID string) (s
 	isVersionAvailable = false
 	isBuildAvailable = false
 	for i := 0; i < len(versions); i++ {
-		if (*versions[i].Enabled) && (*versions[i].Name == dsVersion) {
+		log.Debugf("version name %s and is enabled=%t", *versions[i].Name, *versions[i].Enabled)
+		if *versions[i].Name == dsVersion {
+			log.Debugf("DS Version %s is enabled in the control plane", dsVersion)
```
lsrinivas-pure, line 635: here enable check is not happening . is it intentional?

tests/pds/dataservice_test.go
```diff
@@ -911,6 +911,19 @@ func UpgradeDataService(dataservice, oldVersion, oldImage, dsVersion, dsBuild st
...
+		defer func() {
+			Step("Delete the workload generating deployments", func() {
+				if Contains(dataServiceDeploymentWorkloads, ds.Name) {
+					log.InfoD("Deleting Workload Generating pods %v ", dep.Name)
+					err = pdslib.DeleteK8sDeployments(dep.Name, namespace)
+				} else if Contains(dataServicePodWorkloads, ds.Name) {
+					log.InfoD("Deleting Workload Generating pods %v ", pod.Name)
+					err = pdslib.DeleteK8sPods(pod.Name, namespace)
+				}
+				log.FailOnError(err, "error deleting workload generating pods")
```
lsrinivas-pure, line 923: add ds.Name in the error log

drivers/pds/lib/utils.go
Madanagopal19: yes, we don't need that check as suggested by developers.

tests/pds/dataservice_test.go
Madanagopal19: Done

lsrinivas-pure (approved):
lgtm

PR #1192, opened 2023-03-10
Added Testrail result update capability for Px Backup system tests

mkoppal-px:
Px Backup system tests did not include the testrail test result update feature. Used the existing mechanism and modified it to make it relevant for Px Backup.
This PR includes

- Created a common function `SetupTestRail` to setup testrail
- Added `PxBackupVersion` attribute to the testrail object
- Replaced all `EndTorpedoTest()` function with `EndPxBackupTorpedoTest(contexts)` because we did not want to call `AfterEachTest` at the end of every test and it felt logical to be part of `EndPxBackupTorpedoTest`
- Modified `StartTorpedoTest` to add Milestone in testrail if the testrail setup goes through fine. Verified that this will not affect existing Px tests which are calling `AddRunsToMilestone` outside the `StartTorpedoTest` function

Closes #PA-651

Since our jenkins jobs do not support the testrail paramaters, I tested it locally by running from my laptop. Here is a sample test run - https://portworx.testrail.net/index.php?/runs/view/5660&group_by=cases:custom_automated&group_order=asc&group_id=1

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -271,10 +271,12 @@ const (
 )
 
 var pxRuntimeOpts string
+var PxBackupVersion string
 
-const (
-	taskNamePrefix = "backupcreaterestore"
-	orgID          = "default"
+var (
+	RunIdForSuite             int
+	TestRailSetupSuccessful   bool
+	CurrentTestRailTestCaseId int
 )
 
 var (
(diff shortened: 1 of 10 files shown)


lsrinivas-pure (dismissed):
lgtm

tests/common.go
```diff
@@ -4867,6 +4860,10 @@ func StartTorpedoTest(testName, testDescription string, tags map[string]string,
 	tags["pureVolume"] = fmt.Sprintf("%t", Inst().PureVolumes)
 	tags["pureSANType"] = Inst().PureSANType
 	dash.TestCaseBegin(testName, testDescription, strconv.Itoa(testRepoID), tags)
+	if TestRailSetupSuccessful && testRepoID != 0 {
+		RunIdForSuite = testrailuttils.AddRunsToMilestone(testRepoID)
+		CurrentTestRailTestCaseId = testRepoID
```
lsrinivas-pure, line 4865: will this not be prob for PX cases. please check once

kshithijiyer-px (approved):
LGTM

tests/common.go
mkoppal-px: It is not disruptive operation. It will not fail for PX tests

mkoppal-px: `AddRunsToMilestone` will just check if a run needs to be added or if it already exists. If it already exists it doesn't do anything.

lsrinivas-pure: it will not fail but Im thinking it might add duplicate entries in testrail

lsrinivas-pure (approved):
lgtm

kphalgun-px (approved):
LGTM

PR #1202, opened 2023-03-14
enabling all DS for Independent App TC

dbhatnagar-px:
Enabling all Data Services for Independent App Test Scenario


Madanagopal19 (approved):
LGTM

lsrinivas-pure (approved):
lgtm

PR #1207, opened 2023-03-15
minor fix for resync function

vprabhakar-px:
minor fix to return when the mode is in clean state after reboot,


tests/basic/storage_pool_test.go
```diff
@@ -5251,15 +5253,53 @@ var _ = Describe("{PoolResizeVolumesResync}", func() {
...
+					err = Inst().V.SetReplicationFactor(vol, newRepl, nil, nil, true)
+					if err != nil {
+						return err
+					}
+				}
+
+				// Change Replica sets of each volumes created to 3
+				var maxReplicaFactor int64
+				var nodesToBeUpdated []string
+				var poolsToBeUpdated []string
```
mkoppal-px, line 5272: nit: Would be cleaner if these vars are declared in parenthesis like this 

```
var (
    maxReplicaFactor int64
    nodesToBeUpdated []string
    poolsToBeUpdated []string
)
```

mkoppal-px (approved):
LGTM except for one minor comment

PR #1209, opened 2023-03-16
Add container status check when calcuating pod hour

dahuang-purestorage:
Running postgres job was failing because one pod is stuck on purpose and we counting all pods using the PVC. This commit adds the check for containers that are in ready state.

PWX-29379

Local run passes locally.


ggriffiths (approved):

lsrinivas-pure (approved):
lgtm

PR #1214, opened 2023-03-16
Fixed a cleanup related bug in ShareBackupsAndClusterWithUser

mkoppal-px:
During the cleanup of `ShareBackupsAndClusterWithUser` there was a backup created by testuser which was getting deleted in the `JustAfterEach`, but even before the deletion could complete we are deleting the backup location as `px-central-admin`. Since `px-central-admin` user did not have access to the backup, backup location delete request was throwing error even though we have set the param `DeleteBackups: true`.
To mitigate this, I have added a helper function `DeleteBackupAndWait` which will wait till the backup created by the testuser is deleted and then proceed with backup location deletion.

- Created `DeleteBackupAndWait` helper function
- `DeleteBackupLocation` function used in `CleanupCloudSettingsAndClusters` was not handling the error, so fixed that as well

Closes #PA-719

[Jenkins run](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/442/consoleFull)


tests/backup/backup_basic_test.go
```diff
@@ -2,11 +2,6 @@ package tests
 
 import (
 	"fmt"
-	"os"
-	"strings"
```
kshithijiyer-px, line 6: Why are we moving it down?

mkoppal-px: This is Goland's way for formatting.

kshithijiyer-px: But ideally standard libs should be on top right?

mkoppal-px: I ran the go formatter again and it did not show any change in this file

snigdha-px (approved):
LGTM

tests/backup/backup_basic_test.go
kshithijiyer-px: Okay but this is a concern, we would need to find out more here. I see that in the PRs it keeps moving it up and down.

mkoppal-px: Yes. I agree. Even I am seeing this frequently.

lsrinivas-pure (approved):
lgtm

PR #1216, opened 2023-03-17
Bug fix in ShareBackupWithUsersAndGroups

mkoppal-px:
`ShareBackupWithUsersAndGroups` was failing intermittently because when I was trying to select a user from View Only access group to verify View Only access on a backup, it would choose a user who did not have a backup location to retrieve. Now I am choosing a user who is able to able to retrieve backup location.
Now instead of expecting an error message with `failed to retrieve backup location` it will receive the message `doesn't have permission to restore backup` which is what is expected for a View Only user who is trying to restore.

Closes #PA-727

[Successful Jenkins Run for the test](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/484/console)

### DIFF
--- a/tests/backup/backup_share_test.go
+++ b/tests/backup/backup_share_test.go
@@ -446,6 +446,7 @@ var _ = Describe("{ShareBackupWithUsersAndGroups}", func() {
 	var credName string
 	bkpNamespaces = make([]string, 0)
 	backupLocationMap := make(map[string]string)
+	var chosenUser string
 
 	JustBeforeEach(func() {
 		StartTorpedoTest("ShareBackupWithUsersAndGroups",
@@ -608,9 +609,10 @@ var _ = Describe("{ShareBackupWithUsersAndGroups}", func() {
 		Step("Share Backup with Full access to a user of View Only access group and Validate", func() {
 			log.InfoD("Share Backup with Full access to a user of View Only access group and Validate")
 			// Get user from the view access group
-			username, err := backup.GetRandomUserFromGroup(groups[0])
+			var err error
+			chosenUser, err = backup.GetRandomUserFromGroup(groups[0])
 			log.FailOnError(err, "Failed to get a random user from group [%s]", groups[0])
-			log.Infof("Sharing backup with user - %s", username)
+			log.Infof("Sharing backup with user - %s", chosenUser)
 
 			// Get Admin Context - needed to share backup and get backup UID
 			ctx, err := backup.GetAdminCtxFromSecret()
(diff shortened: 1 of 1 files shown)


vprabhakar-px (approved):

kshithijiyer-px (approved):

tests/backup/backup_share_test.go
```diff
@@ -608,22 +609,23 @@ var _ = Describe("{ShareBackupWithUsersAndGroups}", func() {
 		Step("Share Backup with Full access to a user of View Only access group and Validate", func() {
 			log.InfoD("Share Backup with Full access to a user of View Only access group and Validate")
 			// Get user from the view access group
-			username, err := backup.GetRandomUserFromGroup(groups[0])
+			var err error
+			chosenUser, err = backup.GetRandomUserFromGroup(groups[0])
```
kphalgun-px, line 613: How about chosenUser, err := backup.GetRandomUserFromGroup(groups[0])?

tests/backup/backup_share_test.go
```diff
@@ -608,22 +609,23 @@ var _ = Describe("{ShareBackupWithUsersAndGroups}", func() {
...
 			// Share backup with the user
 			backupName := backupNames[0]
-			err = ShareBackup(backupName, nil, []string{username}, FullAccess, ctx)
+			err = ShareBackup(backupName, nil, []string{chosenUser}, FullAccess, ctx)
 			log.FailOnError(err, "Failed to share backup %s", backupName)
 
 			// Get user context
-			ctxNonAdmin, err := backup.GetNonAdminCtx(username, "Password1")
-			log.FailOnError(err, "Fetching %s ctx", username)
+			ctxNonAdmin, err := backup.GetNonAdminCtx(chosenUser, "Password1")
```
kphalgun-px, line 627: Would it be a better approach to use a global variable "password" instead of hard-coding it within the code?

tests/backup/backup_share_test.go
kphalgun-px, line 1: Do we have to delete backups explicitly in JustAfterEach?

tests/backup/backup_share_test.go
mkoppal-px: `chosenUser` is declared at the test case level not step level. We are using it in multiple steps.

tests/backup/backup_share_test.go
mkoppal-px: We need to do it for all tests. We can do it in a separate PR.

mkoppal-px:
> I will approve this PR, but if future, we cannot use clear passwords, users, accounts of any sort, keys, emails in code, regardless of how fake or harmless they are. We need to implement a method to pass those via configmap, env vars, params, which ever way you guys decide to go with. Might want to first have a discussion for the approach so all can agree on one.

Created a ticket for it - #PA-731

PR #1228, opened 2023-03-20
[PA-733] Fix CreateCloudCredential

kphalgun-px:
**What this PR does / why we need it**:
This PR makes the CreateCloudCredential function return an error instead of handling it internally, aiming to enhance consistency with other common functions. This is in line with best practices for error handling in Torpedo and enables better error management.

**Which issue(s) this PR fixes** (optional)
Closes #PA-733

**Special notes for your reviewer**:
Please review the Jenkins build for the "BasicBackupCreation" test case using the link below

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/567

Aetos Dashboard: http://aetos.pwx.purestorage.com/resultSet/testSetID/129710

PR #1195, opened 2023-03-10
[PA-672] Fix and refactor ClusterBackupShareToggle

kphalgun-px:
[PA-681] Fix GetAllBackupsForUser

[PA-682] Fix ClusterUpdateBackupShare

**What this PR does / why we need it**:
This PR refactors the "ClusterBackupShareToggle" function and fixes an issue with getting the UID of the schedule backup. This PR also fixes issues with GetAllBackupsForUser and ClusterUpdateBackupShare. This PR improves the reliability and functionality of the "ClusterBackupShareToggle" test case, ensuring that it functions as intended.

**Which issue(s) this PR fixes** (optional)
Closes #PA672, #PA681, #PA682

**Special notes for your reviewer**:
If the "ClusterBackupShareToggle" test case fails like below, please ensure that you have the latest version of Px-Backup installed. This PR relies on a fix for a bug that was present in Px-Backup v2.3.3, which has been resolved in a later version.

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/265

Aetos Dashboard: http://aetos.pwx.purestorage.com/resultSet/testSetID/119233

Thanks to the efforts of @snigdha-px  Et. al, we can now ensure that this test case passes as expected with the latest version of Px-Backup installed. Please review the Jenkins build for the "ClusterBackupShareToggle" test case using the link below

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/266

Aetos Dashboard: http://aetos.pwx.purestorage.com/resultSet/testSetID/119240

PR #1230, opened 2023-03-20
[PA-735] Remove RegisterBackupCluster and add GetClusterStatus

kphalgun-px:
The PR eliminates an non significant function, RegisterBackupCluster, and adds GetClusterStatus for obtaining the cluster status.  This may improve overall code maintainability.

Closes #PA-735

Please review the Jenkins build for the "BasicBackupCreation" test case using the link below

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/579

Aetos Dashboard: http://aetos.pwx.purestorage.com/resultSet/testSetID/129891


tthurlapati-px:
This is probably out of scope for this PR, but in all of the test cases, we're checking for the status of the Source Cluster, but not the destination cluster; why is that? Shouldn't we be testing the status of _both_, given that both are being created (in most cases) @mkoppal-px ?

drivers/backup/portworx/portworx.go
```diff
@@ -334,6 +334,22 @@ func (p *portworx) GetClusterName(ctx context.Context, orgID string, clusterUid
...
+		return api.ClusterInfo_StatusInfo_Invalid, err
+	}
+	for _, cluster := range enumerateRsp.GetClusters() {
+		if cluster.GetName() == clusterName {
+			return cluster.Status.Status, nil
+		}
+	}
+	return api.ClusterInfo_StatusInfo_Invalid, fmt.Errorf("cluster with name '%s' not found for org '%s'", clusterName, orgID)
+}
+
```
tthurlapati-px, line 352: You can simplify the function by using a different api endpoint:

```
func (p *portworx) GetClusterStatus(orgID string, clusterName string, ctx context.Context) (api.ClusterInfo_StatusInfo_Status, error) {
	clusterInspectReq := &api.ClusterInspectRequest{
		OrgId: orgID,
		Name: clusterName,
	}
	inspectRsp, err := p.InspectCluster(ctx, clusterInspectReq)
	if err != nil {
		return api.ClusterInfo_StatusInfo_Invalid, err
	}
	return inspectRsp.Cluster.Status.Status, nil
}
```

tthurlapati-px: I just noticed while writing the simplified code for _this_ PR, that in an older PR of yours, we can further simplify: https://github.com/portworx/torpedo/pull/1195#discussion_r1142934124
Though, it's working, so it might be fine.

tthurlapati-px (changes requested):
If the simplification can be made, then let's make it.

tests/backup/backup_locked_bucket_test.go
```diff
@@ -134,8 +135,11 @@ var _ = Describe("{BackupAlternatingBetweenLockedAndUnlockedBuckets}", func() {
 			log.FailOnError(err, "Fetching px-central-admin ctx")
 			err = CreateSourceAndDestClusters(orgID, "", "", ctx)
 			dash.VerifyFatal(err, nil, "Creating source and destination cluster")
-			clusterStatus, clusterUid = Inst().Backup.RegisterBackupCluster(orgID, SourceClusterName, "")
-			dash.VerifyFatal(clusterStatus, api.ClusterInfo_StatusInfo_Online, fmt.Sprintf("Verifying backup cluster %s registration", SourceClusterName))
+			clusterStatus, err = Inst().Backup.GetClusterStatus(orgID, SourceClusterName, ctx)
+			dash.VerifyFatal(err, nil, fmt.Sprintf("Fetching [%s] cluster status", SourceClusterName))
```
snigdha-px, line 139: Can we use log.FailOnError instead of dash.VerifyFatal for this verification to avoid too many log messages as now all the test cases will have this
[Not a mandatory change, only if it makes sense]

kphalgun-px: I am currently using dash.VerifyFatal because it was used before. Perhaps, we can revisit this once we establish a standardized structure and syntax for the test case.

drivers/backup/portworx/portworx.go
kphalgun-px: To maintain consistency with the legacy codebase, I am using EnumerateCluster since many of the helper functions in the older code rely on EnumerateRequest.

kphalgun-px:
> This is probably out of scope for this PR, but in all of the test cases, we're checking for the status of the Source Cluster, but not the destination cluster; why is that? Shouldn't we be testing the status of _both_, given that both are being created (in most cases) @mkoppal-px ?

I fail to comprehend the purpose of adding the destination cluster in numerous test cases, as it appears that all operations are being conducted solely within the source cluster.

snigdha-px (approved):
LGTM

drivers/backup/portworx/portworx.go
```diff
@@ -334,6 +334,22 @@ func (p *portworx) GetClusterName(ctx context.Context, orgID string, clusterUid
 	return "", fmt.Errorf("cluster with uid '%s' not found for org '%s'", clusterUid, orgID)
 }
 
+func (p *portworx) GetClusterStatus(orgID string, clusterName string, ctx context.Context) (api.ClusterInfo_StatusInfo_Status, error) {
+	clusterEnumerateReq := &api.ClusterEnumerateRequest{
+		OrgId: orgID,
+	}
+	enumerateRsp, err := p.EnumerateCluster(ctx, clusterEnumerateReq)
+	if err != nil {
+		return api.ClusterInfo_StatusInfo_Invalid, err
```
mkoppal-px, line 343: We cannot send Invalid as output

drivers/backup/portworx/portworx.go
```diff
@@ -334,6 +334,22 @@ func (p *portworx) GetClusterName(ctx context.Context, orgID string, clusterUid
...
+	}
+	for _, cluster := range enumerateRsp.GetClusters() {
+		if cluster.GetName() == clusterName {
+			return cluster.Status.Status, nil
+		}
+	}
+	return api.ClusterInfo_StatusInfo_Invalid, fmt.Errorf("cluster with name '%s' not found for org '%s'", clusterName, orgID)
```
mkoppal-px, line 350: Same here. Return nil

mkoppal-px:
Replace all `dash.VerifyFatal(err, nil, fmt.Sprintf("Fetching [%s] cluster status", SourceClusterName))` to `log.FailOnError`

Let us follow this across

drivers/backup/portworx/portworx.go
kphalgun-px: nil is not a subtype of api.ClusterInfo_StatusInfo_Status

drivers/backup/portworx/portworx.go
kphalgun-px: nil is not a subtype of api.ClusterInfo_StatusInfo_Status

drivers/backup/portworx/portworx.go
tthurlapati-px: This function will perform much better if you switch to `inspectCluster`. Also, I don't see how this will affect other functions which are using the enumerate method, as I'm not suggesting the removal of anything.
I feel consistency is useful for code readability, but in this case, I see no benefit in being consistent, especially when the alternative method (inspect) is faster and smaller.

@mkoppal-px  what do you suggest?

tthurlapati-px:
> > This is probably out of scope for this PR, but in all of the test cases, we're checking for the status of the Source Cluster, but not the destination cluster; why is that? Shouldn't we be testing the status of _both_, given that both are being created (in most cases) @mkoppal-px ?
> 
> I fail to comprehend the purpose of adding the destination cluster in numerous test cases, as it appears that all operations are being conducted solely within the source cluster.

✅ You're right. This isn't relevant now as we don't work with destination cluster. Maybe in the future, when we do deeper validation of restore.

tthurlapati-px (commented):
The lines where we get `clusteruid`:
```go
			clusterUid, err = Inst().Backup.GetClusterUID(ctx, orgID, SourceClusterName)
			dash.VerifyFatal(err, nil, fmt.Sprintf("Fetching [%s] cluster uid", SourceClusterName))
```
isn't this confusion the same as https://github.com/portworx/torpedo/pull/1228#issuecomment-1477506082 (mithun's suggestion)
So, it would make more sense to use the following, than to compare `err, nil` in `dsah.VerifyFatal`
```go
			clusterUid, err = Inst().Backup.GetClusterUID(ctx, orgID, SourceClusterName)
			log.FailOnError(err, "Fetching [%s] cluster uid", SourceClusterName)
```

kphalgun-px:
> The lines where we get `clusteruid`:
> 
> ```go
> 			clusterUid, err = Inst().Backup.GetClusterUID(ctx, orgID, SourceClusterName)
> 			dash.VerifyFatal(err, nil, fmt.Sprintf("Fetching [%s] cluster uid", SourceClusterName))
> ```
> 
> isn't this confusion the same as [#1228 (comment)](https://github.com/portworx/torpedo/pull/1228#issuecomment-1477506082) (mithun's suggestion) So, it would make more sense to use the following, than to compare `err, nil` in `dsah.VerifyFatal`
> 
> ```go
> 			clusterUid, err = Inst().Backup.GetClusterUID(ctx, orgID, SourceClusterName)
> 			log.FailOnError(err, "Fetching [%s] cluster uid", SourceClusterName)
> ```

I agree, but addressing this issue goes beyond the scope of this PR. Currently, there are no clear guidelines on when to use dash.VerifyFatal and log.FailOnError. I'm not sure if the Aetos Dashboard would show any failures if we completely stop using dash.VerifyFatal.

mkoppal-px (changes requested):
I see 3 tests failing with cluster UID issue. Please run these again and provide the results.

kphalgun-px:
You can view the Jenkins Build for all the test cases using the link below

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/659

Aetos Dashboard: http://aetos.pwx.purestorage.com/resultSet/testSetID/134142 [ONLY UNRELATED TESTS FAILED]

mkoppal-px (approved):
LGTM

tthurlapati-px (approved):
Didn't Mithun clarify that we can use `log.FailOnError` every place you're comparing `err` to `nil`? I guess you could make that change for when getting `clusterUid`, before you merge...

kphalgun-px:
> Didn't Mithun clarify that we can use `log.FailOnError` every place you're comparing `err` to `nil`? I guess you could make that change for when getting `clusterUid`, before you merge...

It was suggested to address cosmetic changes at a future point in time.

PR #1238, opened 2023-03-22
PWX-28271 Add VPS Volume Affinity and AntiAffinity Test

dahuang-purestorage:
Adds a new VPS test for Jenkins to automate. 
The new VPS test aims to run without adding too much functionality into Torpedo backend to update the `placement_strategy` annotation into StatefulSet or StorageClass.

The new mongo-vps app has 2 volumes per pod and deploys 2 pods, so 4 volumes in total. Each pod will have two volumes with label [`statefulset=mongo-$i`, `app=mongo-sts`]. 
The two new test does the following:
- VolumeAntiAffinity on `statefulset`: volume with the same `mongo-$i` label is expected to be deployed on different pool.
- VolumeAffinity on `app` label: all the volumes are expected to be deployed on the same pool. 

We validate the deployment but parsing through the deployed volumes to parse its label and the pool it deployed on. For each label, we count the unique number of pools it deployed on.
- VolumeAntiAffinity: we expect each `statefulset` label to have 2 pools. 
- VolumeAffinity: we expect each `app` label to have 1 pool.

PWX-28271

Test passed locally. 
Jenknis job passed: https://jenkins.pwx.dev.purestorage.com/view/Control%20Plane/job/Torpedo/job/tp-basic-vps/1/console


tests/basic/volumeplacementstrategy_test.go
```diff
@@ -0,0 +1,243 @@
...
+		})
+	})
+
+	AfterEach(func() {
+		AfterEachTest(contexts, testrailID, runID)
+		defer EndTorpedoTest()
+	})
+})
+
+type VolumePlaceMentStrategyTestCase interface {
```
lsrinivas-pure, line 101: can you move this and other non tests functions to new file under vpsutil. so that we can use in other tests if required

pkg/vpsutil/vpsutil.go
```diff
@@ -0,0 +1,102 @@
...
+					Enforcement:      v1beta1.EnforcementRequired,
+					MatchExpressions: matchExpression,
+				},
+			},
+		},
+	}
+}
+
+func getNodePlacement(vol *api.Volume) string {
+	return vol.ReplicaSets[0].PoolUuids[0]
```
sangleganesh, line 54: don't u need to get all not pools and node just pool[0] ?

sangleganesh (commented):
one minor comment else looks good

nikolaypopov (approved):
LGTM

PR #1239, opened 2023-03-22
Fixing failures in ShareAndRemoveBackupLocation and ClusterBackupShareToggle

mkoppal-px:
During one of our nightly runs [Aetos link](http://aetos.pwx.purestorage.com/resultSet/testSetID/131308), we faced failures in `ShareAndRemoveBackupLocation` and `ClusterBackupShareToggle` which was intermittent. To avoid this this test becoming flaky, we have done the following:

- [x] In `ShareAndRemoveBackupLocation` the older backup location is being deleted during the test case but there was not check if it is actually deleted before we enter cleanup. Added a `backupLocationDeleteStatusCheck` for this reason.
- [x] In `ClusterBackupShareToggle`, we are sharing the backups at cluster level, it is taking a few seconds to apply the access level to the backup objects, but even before that is completed we are trying to restore the shared backup which was resulting in intermittent failure. Fixed this by adding a check in `ClusterUpdateBackupShare` helper function. We will check for `BackupShareStatusInfo` and returning from the function only after the status is `Success`

Closes #PA-742

[Jenkins run](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/608/console)

### DIFF
--- a/tests/backup/backup_helper.go
+++ b/tests/backup/backup_helper.go
@@ -519,7 +519,9 @@ func ClusterUpdateBackupShare(clusterName string, groupNames []string, userNames
 	clusterBackupShareStatusCheck := func() (interface{}, bool, error) {
 		clusterReq := &api.ClusterInspectRequest{OrgId: orgID, Name: clusterName, IncludeSecrets: true}
 		clusterResp, err := backupDriver.InspectCluster(ctx, clusterReq)
-		log.FailOnError(err, "Cluster Object for cluster %s and Org id %s is empty", clusterName, orgID)
+		if err != nil {
+			return "", true, err
+		}
 		if clusterResp.GetCluster().BackupShareStatusInfo.GetStatus() != api.ClusterInfo_BackupShareStatusInfo_Success {
 			return "", true, fmt.Errorf("cluster backup share status for cluster %s is still %s", clusterName,
 				clusterResp.GetCluster().BackupShareStatusInfo.GetStatus())


tests/backup/backup_helper.go
```diff
@@ -515,6 +515,23 @@ func ClusterUpdateBackupShare(clusterName string, groupNames []string, userNames
 	if err != nil {
 		return err
 	}
+
+	clusterBackupShareStatusCheck := func() (interface{}, bool, error) {
+		clusterReq := &api.ClusterInspectRequest{OrgId: orgID, Name: clusterName, IncludeSecrets: true}
+		clusterResp, err := backupDriver.InspectCluster(ctx, clusterReq)
+		log.FailOnError(err, "Cluster Object for cluster %s and Org id %s is empty", clusterName, orgID)
```
kphalgun-px, line 522: How about we return the err?

lsrinivas-pure, line 522: return the "",true,err here

mkoppal-px: Fixed it

mkoppal-px: Added this

```
clusterResp, err := backupDriver.InspectCluster(ctx, clusterReq)
		if err != nil {
			return "", true, err
		}
```

lsrinivas-pure (approved):
LGTM

kphalgun-px (approved):
LGTM

PR #1253, opened 2023-03-26
PA-732: Fix `ReplicaChangeWhileRestore`  failure

kshithijiyer-px:
Testcase `ReplicaChangeWhileRestore` was failing in the nightly runs with StorageClass not found error as there was a 1 sec time difference  when we run the case between the time the sc is created and restore is done. To fix this we are just assigning the variable once now in the `step()` where we are creating the StorageClass. 

Change done in the PR:
- [x] Change code to only create scName once and reuse everywhere. 

Run without fix: 
http://aetos.pwx.purestorage.com/resultSet/testSetID/128428/testCaseID/644998

Run with fix: 
1. http://aetos.pwx.purestorage.com/resultSet/testSetID/136967 (Seeing some random 4 cases failing the one ReplicaChangeWhileRestore isn't failing)
2. http://aetos.pwx.purestorage.com/resultSet/testSetID/137903 (Seeing 2 testcase failure.)

 

Closes #PA-732

Will squash and merge if rebase of branch is required.


kphalgun-px (approved):
LGTM

tests/backup/backup_portworx_test.go
```diff
@@ -137,6 +137,7 @@ var _ = Describe("{ReplicaChangeWhileRestore}", func() {
 	labelSelectors := make(map[string]string)
 	params := make(map[string]string)
 	var backupNames []string
+	scName := fmt.Sprintf("replica-sc-%v", time.Now().Unix())
```
mkoppal-px, line 140: I would suggest to just declare the `scName` here like `var scName string` and then in the step where you are creating storage class add the time stamp.

```
scName = fmt.Sprintf("replica-sc-%v", time.Now().Unix())
```

This is because ginkgo initialises all the variables at `Describe` level as soon the ginkgo command is run. So if there is another future test case which us creating another sc with name `fmt.Sprintf("replica-sc-%v", time.Now().Unix())` then it would end up in conflict because both of them would have the same time stamp.

kshithijiyer-px: Done

tthurlapati-px (approved):
LGTM

ak-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #1182, opened 2023-03-08
[PA-614] Backup and Restore b/w Diff K8s version; and simultaneous restore in Diff namespace

tthurlapati-px:
## What this PR does

1. It adds automates the test: https://portworx.testrail.net/index.php?/cases/view/83716, where backup is done on `x` cluster version of `k8s` (1.19) and restore happens on `x+` cluster version (1.23). On top of that, two restores are done simultaneously, resulting in the first being a "success" and the second a "PartialSuccess". The "success" and "partialSuccess" only happens due to existence of cluster level resources.
2. This test is part of the testrail folder for "backup CRD and webhook support", which is actually a pre-requisite for running this test (i.e. the need for cluster level resource to be present). So, a special manifest involving CRDs and webhooks is also introduced, called "elasticsearch-crd-webhook" which uses a generic elastic-search deployment as the test app for backup.
3. BUGFIX: `RefreshDriver` functionality introduced for `ssh.go`, which needs to be invoked while `SetClusterContext` in order to make a complete context switch.
4. BUGFIX: `RefreshDriverEndpoints` was modified to refresh the px ns in order to ensure it uses the namepace corresponding to the current context.
5. ADD Ability to create and destroy CRs to `k8s.go`
6. ADD `ClusterVersion` to `k8s.go`, which is used within the test to verify validity of cluster versions.
7. ADD ability to create, validate and destroy `admissionregistraition` objects, like `ValidatingWebhookConfigurations` to `k8s.go`
8. ADD ability to create, validate and destroy `apiextension` objects i.e. CRDs to `k8s.go`
9. ADD ability to destroy `Rbacobjects` like `Clusterroles` `clusterrolesbindings` to `k8s.go`
10. REFACT: move schedule options creation to new function to use it whenever new `scheduler.contexts` are created (they contain the specs of the k8s objects registered)
11. ADD `CloneSpec` function to create clones of specs (which can be used to generate specs with other properties like different namespaces, which is useful for creating "restore" context specs)
12. ADD `UpdateNamespace` function to be able to change the namespaces of specs based on "from" to "to" mapping.
13. ADD `GetSpecNameKindNamepace` for getting the crucial details of a given spec
14. REFACT `SetClusterContext` to Refresh the ssh node driver
15. REFACT `SetClusterContext` to cache the current context so that repeated requests to switch to same context are ignored
16. ADD return err for `SetDestinationKubeConfig`
[...]

PR #1225, opened 2023-03-20
Added TC:Reboots node when backup is in progress

snigdha-px:
**What this PR does / why we need it**:
1)Added new testcase : Reboots node when backup is in progress
2)In testcase CancelAllRunningBackupJobs, DoRetryWithTimeout loop was returning nil with retry true, hence no retry was happening. Corrected and added error when retry is true
**Which issue(s) this PR fixes** (optional)
Closes #https://portworx.atlassian.net/browse/PA-706
https://portworx.atlassian.net/browse/PA-744

**Special notes for your reviewer**:
Aetos: http://aetos.pwx.purestorage.com/resultSet/testSetID/129822
Log after all the changes: http://aetos.pwx.purestorage.com/resultSet/testSetID/133361

Latest log : http://aetos.pwx.purestorage.com/resultSet/testSetID/133729

PR #1258, opened 2023-03-28
[PA-614] Backup and Restore b/w Diff K8s version....-partial 2 (FIX,REFACT(SetClusterContext))

tthurlapati-px:
This PR (for PA-614, PA-688) is part of a much bigger PR (which has been broken down). Refer this PR: https://github.com/portworx/torpedo/pull/1182 in order to get all the relevant details.

This PR fixes `SetClusterContext` so that it functions properly (fixing some bugs).

> NOTE: The run for ALL tests has already been done for this change as part of https://github.com/portworx/torpedo/pull/1182.


tests/common.go
```diff
@@ -2133,15 +2160,18 @@ func SetSourceKubeConfig() error {
...
 	}
-	SetClusterContext(sourceClusterConfigPath)
-	return nil
+	err = SetClusterContext(sourceClusterConfigPath)
+	return err
 }
 
 // SetDestinationKubeConfig sets current context to the kubeconfig passed as destination to the torpedo test
-func SetDestinationKubeConfig() {
+func SetDestinationKubeConfig() error {
```
kphalgun-px, line 2168: Please refer to PR #1225. How about handling errors where SetDestinationKubeConfig is used?

snigdha-px: Correct. SetDestinationKubeConfig is used in many test cases . Please handle them all

tthurlapati-px: Thanks for reminding me.
Making changes...

tthurlapati-px: ✅

tests/common.go
```diff
@@ -2111,19 +2114,43 @@ func AfterEachTest(contexts []*scheduler.Context, ids ...int) {
...
 	}
 
 	err = Inst().V.RefreshDriverEndpoints()
 	if err != nil {
-		return fmt.Errorf("Failed to switch to context. RefreshDriverEndpoints Error: [%v]", err)
+		return fmt.Errorf("failed to switch to context. RefreshDriverEndpoints Error: [%v]", err)
 	}
+
+	if sshNodeDriver, ok := Inst().N.(*ssh.SSH); ok {
+		err = ssh.RefreshDriver(sshNodeDriver)
```
kphalgun-px, line 2145: How about sshNodeDriver.updateDriver()?

tests/common.go
```diff
@@ -2133,30 +2160,39 @@ func SetSourceKubeConfig() error {
...
-func SetDestinationKubeConfig() {
+func SetDestinationKubeConfig() error {
 	destClusterConfigPath, err := GetDestinationClusterConfigPath()
-	expect(err).NotTo(haveOccurred())
-	SetClusterContext(destClusterConfigPath)
+	if err != nil {
+		return err
+	}
+	err = SetClusterContext(destClusterConfigPath)
+	return err
```
kphalgun-px, line 2174: To be consistent, how about

```
		if err != nil {
			return err
		}
		return nil
```

tests/backup/backup_share_test.go
```diff
@@ -3418,7 +3418,8 @@ var _ = Describe("{IssueMultipleRestoresWithNamespaceAndStorageClassMapping}", f
 			dash.VerifyFatal(err, nil, fmt.Sprintf("Creating new storage class %v on source cluster %s", scName, SourceClusterName))
 
 			log.InfoD("Switching cluster context to destination cluster")
-			SetDestinationKubeConfig()
+			err = SetDestinationKubeConfig()
+			log.FailOnError(err, "Failed to set destination kubeconfig")
```
kphalgun-px, line 3422: How about log.FailOnError(err, "Switching context to destination cluster failed") or log.FailOnError(err, "Failed to set destination kubeconfig") everywhere?

tests/common.go
tthurlapati-px: Refreshing, in this case, is the same as "updating" (which is what I named the function). But it doesn't have to be that way, so having the name "refresh" makes more sense here, especially if more logic is introduced in `RefreshDriver`, but not in `updateDriver` in the future, since `updateDriver` is also used by `Init`. This is more consistent with the behaviour of other drivers

tests/common.go
tthurlapati-px: There's no point in doing that, as this error is the last error, which means, there **is no concept** of "early return due to error". This is in fact idiomatic code, as an if statement is not a useful check here, as if it is `nil`, then it will be `nil` and we don't need to modify it.

You suggestion makes sense if this were a function which is constantly modified by people, so it would make sense to have "consistency". But this function might not change ever again, so it's fine.

tests/common.go
kphalgun-px: To ensure that refreshDriver always uses the current implementation of code, you can consider copying and pasting the updateDriver code instead of calling it.

tests/common.go
mkoppal-px: Why don't you just do `return SetClusterContext(destClusterConfigPath)`

drivers/node/ssh/ssh.go
```diff
@@ -194,6 +196,17 @@ func (s *SSH) Init(nodeOpts node.InitOptions) error {
 	return nil
 }
 
+// Init initializes SSH node driver
+func (s *SSH) Init(nodeOpts node.InitOptions) error {
+	s.specDir = nodeOpts.SpecDir
+	return s.updateDriver()
+}
+
+// RefreshDriver Refreshes the driver on a context switch
```
mkoppal-px, line 205: nit: make it `refreshes`

tests/basic/async_dr_test.go
```diff
@@ -67,7 +67,10 @@ var _ = Describe("{MigrateDeployment}", func() {
 	It("has to deploy app, create cluster pair, migrate app", func() {
 		Step("Deploy applications", func() {
 
-			SetSourceKubeConfig()
+			err := SetSourceKubeConfig()
+			if err != nil {
+				log.FailOnError(err, "Switching context to source cluster failed")
+			}
```
mkoppal-px, line 73: Just do `log.FailOnError(err, "Switching context to source cluster failed")`
Checking for error is unnecessary because that is handled in `log.FailOnError`

tests/common.go
tthurlapati-px: I don't understand why one would duplicate the code. We know that the logic in `updateDriver` _definitely_ has to be used whenever there's initialization _or_ reinit (aka refresh), so it makes sense to call it from both functions, not duplicate the code.

tests/common.go
tthurlapati-px: ✅

drivers/node/ssh/ssh.go
tthurlapati-px: ✅

tests/basic/async_dr_test.go
tthurlapati-px: Oh, yeah...
✅

tests/common.go
```diff
@@ -2111,19 +2114,43 @@ func AfterEachTest(contexts []*scheduler.Context, ids ...int) {
...
+	// This variable is used to clearly indicate that in logs
+	var clusterConfigPathOrDefault = clusterConfigPath
+	if clusterConfigPathOrDefault == "" {
+		clusterConfigPathOrDefault = "default"
+	}
+	log.InfoD("Switching context to [%s]", clusterConfigPathOrDefault)
+	if clusterConfigPath == CurrentClusterConfigPath {
+		log.InfoD("Switching context: The context is already [%s]", clusterConfigPathOrDefault)
+		return nil
+	}
```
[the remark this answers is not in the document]
tthurlapati-px: In this functions, I store the value of the current cluster config (kubeconfig which is active) in `CurrentClusterConfigPath`. So, the comparison of `clusterConfigPath` func arg happens with `CurrentClusterConfigPath`.
_But_, there's a case where if `clusterConfigPath` is `""` (empty string), then the entire function assumes that it is the default cluster (`KUBECONFIG` env var, afaik). The only **problem** is that if the string is empty, then it can confuse someone in logs when them see `"Switching context: The context is already []"`, which is why, I ensure the log message is complete, by making the message `"Switching context: The context is already [default]"` if the string is `""`.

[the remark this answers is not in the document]
mkoppal-px: Instead of doing this 

```
var clusterConfigPathOrDefault = clusterConfigPath
	if clusterConfigPathOrDefault == "" {
		clusterConfigPathOrDefault = "default"
	}
```

You can simply do this

```
if clusterConfigPath == CurrentClusterConfigPath {
                 if CurrentClusterConfigPath == "" {
		// construct log message
	}
		log.InfoD("Switching context: The context is already [%s]", clusterConfigPathOrDefault)
		return nil
	}
```

mkoppal-px (changes requested):
Provided a minor comment. Please attach the test result for all the tests that are using these functions that you have modified

tests/common.go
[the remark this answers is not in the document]
tthurlapati-px: I fixed it to make it slightly more readable. I still had to keep it mostly the way it was, as another logging statement at the end of the function uses `clusterConfigPathForLog` (previously `clusterConfigPathOrDefault`), so in my code we only assign it once, which is better (than comparing twice).

tests/common.go
kphalgun-px: Makes sense.

tthurlapati-px:
The test runs:

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/924/console
http://aetos.pwx.purestorage.com/resultSet/testSetID/148318

`AllNSBackupWithIncludeNewNSOption` didn't succeed as my stork isn't updated; otherwise it should be fine.

mkoppal-px (approved):
LGTM

kphalgun-px (approved):
LGTM

PR #1259, opened 2023-03-28
adding error checks in case of missed alerts

vprabhakar-px:
fixing test case to check the case when the script run for the first time and no errors are generated


lsrinivas-pure (approved):
lgtm

PR #1267, opened 2023-03-28
Added TC : CancelAllRunningRestoreJobs

ak-px:
Added testcase : CancelAllRunningRestoreJobs
https://portworx.testrail.net/index.php?/tests/view/22623367

Closes #PA-771

Jenkins BYOC run:
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/788/
http://aetos.pwx.purestorage.com/resultSet/testSetID/142409


tests/backup/backup_resiliency_test.go
```diff
@@ -1311,3 +1311,284 @@ var _ = Describe("{RebootNodesWhenBackupsAreInProgress}", func() {
...
+				}
+				_, err = task.DoRetryWithTimeout(restoreJobCancelStatus, maxWaitPeriodForRestoreJobCancellation*time.Minute, restoreJobCancellationRetryTime*time.Second)
+				if err != nil {
+					adminRestores, error := GetAllRestoresAdmin()
+					log.Infof("The list of restores after restore cancellation and wait of 20 minutes is %v", adminRestores)
+					log.FailOnError(error, "Getting the list of restores after restore cancellation and wait of 20 minutes")
+				}
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying restore jobs cancellation while restores is in progress"))
+			}
+			log.Infof("All the restores created by this testcase is deleted after restore cancellation")
```
kphalgun-px, line 1564: nit error: are* deleted

tests/backup/backup_resiliency_test.go
```diff
@@ -1311,3 +1311,284 @@ var _ = Describe("{RebootNodesWhenBackupsAreInProgress}", func() {
...
+			destClusterStatus, err = Inst().Backup.GetClusterStatus(orgID, destinationClusterName, ctx)
+			log.FailOnError(err, fmt.Sprintf("Fetching [%s] cluster status", destinationClusterName))
+			dash.VerifyFatal(destClusterStatus, api.ClusterInfo_StatusInfo_Online, fmt.Sprintf("Verifying if [%s] cluster is online", destinationClusterName))
+		})
+
+		Step("Taking backup of applications", func() {
+			log.InfoD("Taking backup of applications")
+			ctx, err := backup.GetAdminCtxFromSecret()
+			log.FailOnError(err, "Fetching px-central-admin ctx")
+			for _, namespace := range appNamespaces {
```
kphalgun-px, line 1441: How about taking backups in parallel?

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,284 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+				backupNames := make([]string, 0)
+				log.Infof("Taking manual backup of application without rule's")
```
tthurlapati-px, line 1690: nit: it should be `rules`, not `rule's`. (this error occurs in other places in the code too)

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,284 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+				log.Infof("Taking schedule backup of application without rule's")
+				scheduleName = fmt.Sprintf("%s-%s-%v", BackupNamePrefix, namespace, time.Now().Unix())
+				schPolicyUid, _ = Inst().Backup.GetSchedulePolicyUid(orgID, ctx, periodicPolicyName)
+				err = CreateScheduleBackup(scheduleName, SourceClusterName, bkpLocationName, backupLocationUID, []string{namespace},
+					labelSelectors, orgID, "", "", "", "", periodicPolicyName, schPolicyUid, ctx)
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Verification of creating schedule backup with schedule name - %s", scheduleName))
+				schBackupName, err = GetFirstScheduleBackupName(ctx, scheduleName, orgID)
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Fetching the name of the first schedule backup - %s", schBackupName))
+				backupNames = append(backupNames, schBackupName)
+				scheduleNames = append(scheduleNames, scheduleName)
```
tthurlapati-px, line 1706: I have a doubt about the behavior of this test:
You are creating both manual and scheduled backups, for both apps with and without rules.

Though, from the logic, it appears to me that, if we consider that you are backing up something with rules (like postgres), then it will be backed up 4 times, where two of the times are without using the rules despite the fact that postgres uses rules.
If we consider that you are backing up something without rules, then won't your logic do `Inst().Backup.GetRuleUid` for an app that doesn't have rules? In which case, it appears that you're ignoring the error value returned by `Inst().Backup.GetRuleUid`. **Though**, _this should still work_ as the `preRuleId` string will just be empty, which is the same as backing up without rules - and this is fine.

I guess what I'm saying is, if I'm understanding correctly, you can remove the logic where you create backup for app "without rules", since if an app doesn't have rules, the `preRuleId` and `postRuleId` will turn up empty, so it's enough to have 2 types of backups (mnual and schedule), where you look for pre and post rules in both cases and they will be used if the string is not `""`.

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,284 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+				backupNamesMap[namespace] = backupNames
+				threadCount = threadCount + len(backupNames)
+			}
+			log.Infof("The Map of backups taken are: %v", backupNamesMap)
+		})
+
+		var sem = make(chan struct{}, threadCount)
```
tthurlapati-px, line 1735: I'm not able to understand what `sem` and `threadcount` are doing here.

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,284 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+			log.InfoD("Restoring the backed up applications")
+			ctx, err := backup.GetAdminCtxFromSecret()
+			log.FailOnError(err, "Fetching px-central-admin ctx")
+			for _, namespace := range appNamespaces {
+				namespaceMapping := make(map[string]string)
+				for _, backupName := range backupNamesMap[namespace] {
+					sem <- struct{}{}
+					restoreName = fmt.Sprintf("%s-%s", restoreNamePrefix, backupName)
+					restoreNames = append(restoreNames, restoreName)
+					time.Sleep(1 * time.Second)
```
tthurlapati-px, line 1747: why do we sleep here?

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,284 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+						defer func() { <-sem }()
+						_, err = CreateRestoreWithoutCheck(restoreName, backupName, namespaceMapping, destinationClusterName, orgID, ctx)
+						dash.VerifyFatal(err, nil, fmt.Sprintf("Restoring the backup %s with name %s", backupName, restoreName))
+					}(restoreName, backupName, namespaceMapping)
+				}
+			}
+			wg.Wait()
+			log.Infof("The list of restores taken are: %v", restoreNames)
+			log.InfoD("Sleeping for 20 seconds so that the request reaches stork and the restore process is started")
+			time.Sleep(20 * time.Second)
```
tthurlapati-px, line 1762: Wouldn't it be better to check every second for status of a restore to be `InProgress`, and as soon as that happen, one can initiate a delete?  (I have seen a restore finish in 20 seconds, whereas I think this testcase requires it to be `InProgress` i.e. it should be running when deleted, right?)

Or is that unnecessary complexity and sleeping is a simpler alternative?

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,284 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+					defer GinkgoRecover()
+					defer wg.Done()
+					defer func() { <-sem }()
+					err := DeleteRestore(restoreName, orgID, ctx)
+					dash.VerifyFatal(err, nil, fmt.Sprintf("Deleting restore %s while restore is in progress", restoreName))
+				}(restoreName)
+			}
+			wg.Wait()
+			log.InfoD("Sleeping for 30 seconds for the restore cancellation to take place")
+			time.Sleep(30 * time.Second)
```
tthurlapati-px, line 1782: Similar to https://github.com/portworx/torpedo/pull/1267/files#r1151443361, would it or would it not be better to check status (to verify it is done, before doing `wg.Done()`), than to just sleep. Of course, we mostly use postgres or mysql, so we should be fine with "hardcoded" values.
Though, if you check status and wait for it to finish, you won't need to verify in the step below.

tthurlapati-px (commented):
Maybe, if you use the concept of status checks, then for each restore, you can create a go routine, which "creates a restore -> checks for inprogress -> initiates delete -> checks for deletion".

tests/backup/backup_resiliency_test.go
ak-px: Done

tests/backup/backup_resiliency_test.go
ak-px: Revereted the code for separate set of backups, And used parallel backup creation.

tests/backup/backup_resiliency_test.go
ak-px: The idea here was having seperate set of backup's. but currentky reverted the code to use normal backups due to some irregularities seen in updating rule from AppParameters.  We have a section called AppParameters where we define rules for any incoming apps so that there wont be any empty rules while creation of backup. 
The issue seen can be taken care when we test with mutiple apps.

tests/backup/backup_resiliency_test.go
ak-px: Since this is parallel execution of restore's, sleep is added such that new namespace will have different timestamp suffixed.

tests/backup/backup_resiliency_test.go
ak-px: handled above.

tests/backup/backup_resiliency_test.go
ak-px: Added logic to tiggger delete while restore status is in progress.

ak-px:
> Maybe, if you use the concept of status checks, then for each restore, you can create a go routine, which "creates a restore -> checks for inprogress -> initiates delete -> checks for deletion".

The idea of the testcase to cancel all the restore at a go, then verifies the status of cancellation, It has to be separate workflow.

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,211 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+					_, err := task.DoRetryWithTimeout(restoreProgressCheckFunc, maxWaitPeriodForRestoreCompletionInMinute*time.Minute, restoreJobProgressRetryTime*time.Second)
+					dash.VerifyFatal(err, nil, fmt.Sprintf("Verfiying restore %s is in progress", restoreName))
+					err = DeleteRestore(restoreName, orgID, ctx)
+					dash.VerifyFatal(err, nil, fmt.Sprintf("Deleting restore %s while restore is in progress", restoreName))
+				}(restoreName)
+			}
+			wg.Wait()
+		})
+
+		Step("Verifying if all the restore creation is cancelled", func() {
```
sn-px, line 1723: can you ensuring the restore count is zero before the test begins or to capture the number og restore objects

ak-px: Its handled at line number 1728

tests/backup/backup_resiliency_test.go
tthurlapati-px: oh yeah, makes sense...

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,211 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+						}
+						return "", true, fmt.Errorf("restore status for [%s] expected was [%v] but got [%s]", restoreName, statusExpected, actual)
+					}
+					_, err := task.DoRetryWithTimeout(restoreProgressCheckFunc, maxWaitPeriodForRestoreCompletionInMinute*time.Minute, restoreJobProgressRetryTime*time.Second)
+					dash.VerifyFatal(err, nil, fmt.Sprintf("Verfiying restore %s is in progress", restoreName))
+					err = DeleteRestore(restoreName, orgID, ctx)
+					dash.VerifyFatal(err, nil, fmt.Sprintf("Deleting restore %s while restore is in progress", restoreName))
+				}(restoreName)
+			}
+			wg.Wait()
```
tthurlapati-px, line 1720: Over here, you're waiting for all restores to be created before youstart checking for inprogress state.
This could result in an error, where one of the restores completes _before_ all the restores finish starting up, which results a failure in the step where you check for inprogress, as inprogress will never happen.

A solution is to combine the two steps into one. In each thread, you can start the restore, check for inprogress and delete, and _then_ return from the thread. That way, the "restore-inprogress-delete" process is **independent** of other restores for any restore. 

Like:
```
for (...){
time.sleep
go func() {
start restore()
if inprogrss{
delete restore()
}
}
}
```

tthurlapati-px (commented):
If we test with postgres or mysql app, then rules are required to be used, right?

tests/backup/backup_resiliency_test.go
ak-px: If you see the restores are created without success check and will be triggered parallely, And coming to cancellation of restore's all will be triggered parallely without any dependent wait time or restore. 
As per testcase both should be seperate workflow.

ak-px: Thanks for the suggestion, I've solved this case by creating custom namespce with random string instead of time stamp and sleep.

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,214 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+		Step("Taking backup of applications", func() {
+			log.InfoD("Taking backup of applications")
+			ctx, err := backup.GetAdminCtxFromSecret()
+			log.FailOnError(err, "Fetching px-central-admin ctx")
+			for _, namespace := range appNamespaces {
+				backupNames := make([]string, 0)
+				for i := 0; i < numberOfBackups; i++ {
+					time.Sleep(3 * time.Second)
+					backupName := fmt.Sprintf("%s-%s-%d-%v", BackupNamePrefix, namespace, i, time.Now().Unix())
+					backupNames = append(backupNames, backupName)
```
tthurlapati-px, line 1649: Does one still need `time.Sleep` now that you've made the names unique?

tthurlapati-px (commented):
I think the only thing one would have to worry about is if `CreateRestoreWithoutCheck` took too long (in which case, you would miss `InProgress` state).
But I don't think that's really an issue, as it should be fast enough. So, we can be happy with this code.

tests/backup/backup_resiliency_test.go
tthurlapati-px: Alright.

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,214 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+							return "", false, err
+						}
+						actual := resp.GetRestore().GetStatus().Status
+						if actual == statusExpected {
+							return "", false, nil
+						}
+						if actual == api.RestoreInfo_StatusInfo_Failed {
+							return "", false, fmt.Errorf("restore status for [%s] expected was [%v] but failed", restoreName, actual)
+						}
+						return "", true, fmt.Errorf("restore status for [%s] expected was [%v] but got [%s]", restoreName, statusExpected, actual)
```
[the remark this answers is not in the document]
ak-px: Since the inspect request's are triggered soon after the creation there is lower chance for restore to be succesful,mostly it will be in retrying in pending state. But yes for safer side will add that to assertion.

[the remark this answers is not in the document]
tthurlapati-px: Cancelling all "running" jobs means job should be inprogress. Otherwise, technically speaking, you cannot say the test was successful, since you are cancelling a job which has finished, right?

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,214 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+				for _, backupName := range backupNamesMap[namespace] {
+					restoreName := fmt.Sprintf("%s-%s", restoreNamePrefix, backupName)
+					restoreNames = append(restoreNames, restoreName)
+					customNamespace := fmt.Sprintf("new-namespace-%s", RandomString(10))
+					namespaceMapping[namespace] = customNamespace
+					wg.Add(1)
+					go func(restoreName string, backupName string, namespaceMapping map[string]string) {
+						defer GinkgoRecover()
+						defer wg.Done()
+						_, err = CreateRestoreWithoutCheck(restoreName, backupName, namespaceMapping, destinationClusterName, orgID, ctx)
```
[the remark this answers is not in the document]
ak-px: For safer side i'll handle it in progresssuccesscheck.

tests/backup/backup_resiliency_test.go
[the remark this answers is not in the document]
ak-px: That true, But here this is the way we can catch inprogress with automation scope from my understanding, or else to avoid the microseconds where restore is succesfull we may need to introduce any network delay points to hold restore in progress for long time.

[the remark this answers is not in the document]
ak-px: For safer side i'm adding other status in assertion.

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,216 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+							return "", true, fmt.Errorf("%v restore is still present", restoreName)
+						}
+					}
+					return "", false, nil
+				}
+				_, err = task.DoRetryWithTimeout(restoreJobCancelStatus, maxWaitPeriodForRestoreJobCancellation*time.Minute, restoreJobCancellationRetryTime*time.Second)
+				if err != nil {
+					adminRestores, error := GetAllRestoresAdmin()
+					log.Infof("The list of restores after restore cancellation  %v", adminRestores)
+					log.FailOnError(error, "Getting the list of restores after restore cancellation")
```
tthurlapati-px, line 1750: Shouldn't error checking be done first?

tthurlapati-px (changes requested):

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,216 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+				for i := 0; i < numberOfBackups; i++ {
+					time.Sleep(3 * time.Second)
+					backupName := fmt.Sprintf("%s-%s-%d-%v", BackupNamePrefix, namespace, i, time.Now().Unix())
+					backupNames = append(backupNames, backupName)
+					wg.Add(1)
+					go func(backupName string, namespace string) {
+						defer GinkgoRecover()
+						defer wg.Done()
+						err := CreateBackup(backupName, SourceClusterName, bkpLocationName, backupLocationUID,
+							[]string{namespace}, labelSelectors, orgID, srcClusterUid, "", "", "", "", ctx)
```
tthurlapati-px, line 1655: If we test with postgres or mysql app (which is the default), then rules are required to be used, right?

snigdha-px (approved):
LGTM

tests/backup/backup_resiliency_test.go
ak-px: Its already there

tests/backup/backup_resiliency_test.go
ak-px: It's  not mandatory to have rule's for backup, it was my intention to keep seperate set of backup's with and without rules.  for mysql we are yet give rule support in automation. which we can take it in another PR.

tthurlapati-px: I just saw that mysql doesn't have a rule. 👍🏻 

Also, I've seen many tests where rules are not used during backup, even though technically both backup **with and without labels** can be done in the same line. I'll take it up / clarify with Mithun.

tests/backup/backup_resiliency_test.go
ak-px: Moved the lines, Thanks

tthurlapati-px (approved):
LGTM

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,216 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+					for _, restoreName := range restoreNames {
+						if IsPresent(adminRestores, restoreName) {
+							return "", true, fmt.Errorf("%v restore is still present", restoreName)
+						}
+					}
+					return "", false, nil
+				}
+				_, err = task.DoRetryWithTimeout(restoreJobCancelStatus, maxWaitPeriodForRestoreJobCancellation*time.Minute, restoreJobCancellationRetryTime*time.Second)
+				if err != nil {
+					adminRestores, error := GetAllRestoresAdmin()
```
mkoppal-px, line 1748: Use `err` instead of `error`
It conflicts with the error interface

tests/backup/backup_resiliency_test.go
```diff
@@ -1558,3 +1558,216 @@ var _ = Describe("{ScaleDownPxBackupPodWhileBackupAndRestoreIsInProgress}", func
...
+				wg.Add(1)
+				go func(restoreName string) {
+					defer GinkgoRecover()
+					defer wg.Done()
+					restoreInspectRequest := &api.RestoreInspectRequest{
+						Name:  restoreName,
+						OrgId: orgID,
+					}
+					requiredStatus := api.RestoreInfo_StatusInfo_InProgress
+					statusExpected := api.RestoreInfo_StatusInfo_Pending
```
mkoppal-px, line 1703: Use the variable name `restoreInProgressStatus` and `restorePendingStatus`

ak-px:
All testcase run after merging to master.
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/928/

PR #1293, opened 2023-04-03
PA-707: Add CloudSnapsSafeWhenBackupLocationDeleteTest test

kshithijiyer-px:
- [x] Add test to check if cloud sanp is not delete on backup location delete

Run results: 
http://aetos.pwx.purestorage.com/resultSet/testSetID/157208 
~http://aetos.pwx.purestorage.com/resultSet/testSetID/147213~

Closes # PA-707

As usual will squash and merge.


tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+			}
+			_, err := task.DoRetryWithTimeout(checkBackupSync, 100*time.Minute, 30*time.Second)
+			log.FailOnError(err, "Wait for BackupSync to complete")
+			fetchedBackupNames, err := GetAllBackupsAdmin()
+			log.FailOnError(err, "Getting a list of all backups")
+			dash.VerifyFatal(len(fetchedBackupNames), len(backupNames), "Comparing the expected and actual number of backups")
+			var bkp *api.BackupObject
+			backupDriver := Inst().Backup
+			bkpEnumerateReq := &api.BackupEnumerateRequest{
+				OrgId: orgID}
```
tthurlapati-px, line 1621: nit: following looks better:
```
bkpEnumerateReq := &api.BackupEnumerateRequest{
				OrgId: orgID
}
```

tthurlapati-px (changes requested):

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+		log.InfoD("Deleting the deployed apps after the testcase")
+		opts := make(map[string]bool)
+		opts[SkipClusterScopedObjects] = true
+		ValidateAndDestroy(contexts, opts)
+
+		ctx, err := backup.GetAdminCtxFromSecret()
+		log.FailOnError(err, "Fetching px-central-admin ctx")
+		CleanupCloudSettingsAndClusters(backupLocationMap, credName, cloudCredUID, ctx)
+	})
+})
```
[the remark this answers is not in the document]
kshithijiyer-px: Removed

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
 		CleanupCloudSettingsAndClusters(backupLocationMap, credName, cloudCredUID, ctx)
 	})
 })
+
+// CloudSnapsSafeWhenBackupLocationDeleteTest take a good number of backups check if cloud snaps are removed
+// if backup location is removed
```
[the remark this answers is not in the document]
kshithijiyer-px: okay I'll update it.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+	var cloudCredUidList []string
+	var appContexts []*scheduler.Context
+	var bkpNamespaces []string
+	var clusterUid string
+	var clusterStatus api.ClusterInfo_StatusInfo_Status
+	var customBackupLocationName string
+	var credName string
+	bkpNamespaces = make([]string, 0)
+	backupNamespaceMap := make(map[string]string)
+	backupLocationMap := make(map[string]string)
```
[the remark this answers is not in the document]
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+
+	JustBeforeEach(func() {
+		StartTorpedoTest("CloudSnapsSafeWhenBackupLocationDeleteTest",
+			"Validate if cloud snaps presence if backup location is deleted", nil, 58069)
```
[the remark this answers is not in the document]
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+						defer GinkgoRecover()
+						defer wg.Done()
+						defer func() { <-sem }()
+						err = CreateBackup(backupName, SourceClusterName, customBackupLocationName, backupLocationUID, []string{namespace},
+							labelSelectors, orgID, clusterUid, "", "", "", "", ctx)
+						dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying backup creation: %s", backupName))
+					}(backupName)
+				}
+				wg.Wait()
+			}
```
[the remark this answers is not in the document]
kshithijiyer-px: You will not know if it passed or failed until it completes for all namespaces that's something we don't want right?

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+					sem <- struct{}{}
+					time.Sleep(timeBetweenConsecutiveBackups)
+					backupName := fmt.Sprintf("%s-%v", BackupNamePrefix, time.Now().Unix())
+					backupNames = append(backupNames, backupName)
+					wg.Add(1)
+					go func(backupName string) {
+						defer GinkgoRecover()
+						defer wg.Done()
+						defer func() { <-sem }()
+						err = CreateBackup(backupName, SourceClusterName, customBackupLocationName, backupLocationUID, []string{namespace},
```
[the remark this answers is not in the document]
kshithijiyer-px: Nice catch will fix it.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+		Step("Remove the backup location where backups were taken", func() {
+			log.InfoD("Remove backup location where backups were taken")
+			// Issue a remove backup location call
+			err := DeleteBackupLocation(customBackupLocationName, backupLocationUID, orgID, false)
+			dash.VerifySafely(err, nil, fmt.Sprintf("Deleting backup location %s", customBackupLocationName))
+
+			// Wait until backup location is removed
+			backupLocationDeleteStatusCheck := func() (interface{}, bool, error) {
+				ctx, err := backup.GetAdminCtxFromSecret()
+				log.FailOnError(err, "Fetching px-central-admin ctx")
```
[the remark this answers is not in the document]
kshithijiyer-px: No! This is not the case here. Hence not changing here. This will mostly be required if you run it as a goroutine and not as a standard process.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+				backupInspectRequest := &api.BackupInspectRequest{
+					Name:  bkp.Name,
+					Uid:   bkp.Uid,
+					OrgId: orgID,
+				}
+				resp, err := backupDriver.InspectBackup(ctx, backupInspectRequest)
+				log.FailOnError(err, "Inspect each backup from list")
+				actual := resp.GetBackup().GetStatus().Status
+				expected := api.BackupInfo_StatusInfo_Success
+				dash.VerifyFatal(actual, expected, fmt.Sprintf("Check each backup for success status %s", bkp.Name))
```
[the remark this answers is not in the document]
kshithijiyer-px: dash.VerifyFatal() will come up rad in case of failure on the dashboard.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+			}
+		})
+		Step("Taking backup of applications to trigger BackupSync goroutine", func() {
+			log.InfoD("Taking backup of applications to trigger BackupSync goroutine")
+			ctx, err := backup.GetAdminCtxFromSecret()
+			log.FailOnError(err, "Fetching px-central-admin ctx")
+			for _, namespace := range bkpNamespaces {
+				backupName = fmt.Sprintf("%s-%s-%v", BackupNamePrefix, namespace, time.Now().Unix())
+				backupNamespaceMap[namespace] = backupName
+				err = CreateBackup(backupName, SourceClusterName, customBackupLocationName, backupLocationUID, []string{namespace}, labelSelectors, orgID, clusterUid, "", "", "", "", ctx)
```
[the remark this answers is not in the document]
kshithijiyer-px: Make sense will add this.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+			dash.VerifySafely(err, nil, fmt.Sprintf("Deleting backup location %s", customBackupLocationName))
+		})
+
+		Step("Add the backup location again which had backups", func() {
+			log.InfoD("Add the backup location with backups back")
+			for _, provider := range providers {
+				customBackupLocationName = fmt.Sprintf("autogenerated-backup-location-%v", time.Now().Unix())
+				backupLocationMap[backupLocationUID] = customBackupLocationName
+				err := CreateBackupLocation(provider, customBackupLocationName, backupLocationUID, credName, cloudCredUID, getGlobalBucketName(provider), orgID, "")
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Creating backup location %s", customBackupLocationName))
```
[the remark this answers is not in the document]
kshithijiyer-px: What can you figure out from the  `backupLocationUID` this will be just unnecessary logging.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+				cloudCredUidList = append(cloudCredUidList, cloudCredUID)
+				backupLocationUID = uuid.New()
+				credName = fmt.Sprintf("autogenerated-cred-%v", time.Now().Unix())
+				err := CreateCloudCredential(provider, credName, cloudCredUID, orgID, ctx)
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying creation of cloud credential named [%s] for org [%s] with [%s] as provider", credName, orgID, provider))
+				log.InfoD("Created Cloud Credentials with name - %s", credName)
+				customBackupLocationName = fmt.Sprintf("autogenerated-backup-location-%v", time.Now().Unix())
+				err = CreateBackupLocation(provider, customBackupLocationName, backupLocationUID, credName, cloudCredUID, getGlobalBucketName(provider), orgID, "")
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Creating backup location %s", customBackupLocationName))
+				log.InfoD("Created Backup Location with name - %s", customBackupLocationName)
```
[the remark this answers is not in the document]
kshithijiyer-px: Make sense we can get rid of it.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,209 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+			for _, provider := range providers {
+				cloudCredUID = uuid.New()
+				cloudCredUidList = append(cloudCredUidList, cloudCredUID)
+				backupLocationUID = uuid.New()
+				credName = fmt.Sprintf("autogenerated-cred-%v", time.Now().Unix())
+				err := CreateCloudCredential(provider, credName, cloudCredUID, orgID, ctx)
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying creation of cloud credential named [%s] for org [%s] with [%s] as provider", credName, orgID, provider))
+				log.InfoD("Created Cloud Credentials with name - %s", credName)
+				customBackupLocationName = fmt.Sprintf("autogenerated-backup-location-%v", time.Now().Unix())
+				err = CreateBackupLocation(provider, customBackupLocationName, backupLocationUID, credName, cloudCredUID, getGlobalBucketName(provider), orgID, "")
```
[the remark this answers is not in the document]
kshithijiyer-px: Done let's populate in the map.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,220 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+							defer wg.Done()
+							defer func() { <-sem }()
+							err = CreateBackup(backupName, SourceClusterName, backupLocationName, backupLocationUID, []string{namespace},
+								labelSelectors, orgID, clusterUid, "", "", "", "", ctx)
+							dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying backup creation: %s", backupName))
+						}(backupName)
+					}
+					wg.Wait()
+				}
+			}
```
tthurlapati-px, line 1554: Continuation of https://github.com/portworx/torpedo/pull/1293#discussion_r1157065367

**You said**: You will not know if it passed or failed until it completes for all namespaces that's something we don't want right?

I don't see how the notification of passing or failing is dependent on _when_ we call `wg.Wait()`. Like, calling `wg.Wait()` outside the main for loop can ensure maximum parallelism. As for whether we know if `CreateBackup` in a goroutine passed/failed, that will occur as soon as `dash.VerifyFatal` is called, right? Because, if it failed, then `GinkgoRecover` will fail the spec immediately (**without waiting for any of the other goroutines, right?**), and if it succeeded, then it's execution as usual.

It's not like we get to know the status of a createbackup _only_ after `wg.Wait()`. As I as far as I can tell, ginkgo will fail the spec as soon as it fails, and not wait for `wg.Wait()` before failing, as there is no interdependency between ginkgo andnumber of currently running goroutines.

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
tthurlapati-px: Sure, but my comment was more focused on the wording in the logs. The current code (especially `log.FailOnError` doesn't provide enough info on what went wrong (simply says "Inspect each backup from list" without backup name). Hence, I suggested more informative log statements.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,220 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+			log.Infof("List of backups - %v", backupNames)
+		})
+
+		Step("Remove the backup locations where backups were taken", func() {
+			log.InfoD("Remove backup locations where backups were taken")
+			// Issue a remove backup location call
+			for backupLocationUID, customBackupLocationName = range backupLocationMap {
+				err := DeleteBackupLocation(customBackupLocationName, backupLocationUID, orgID, false)
+				dash.VerifySafely(err, nil, fmt.Sprintf("Deleting backup location %s", customBackupLocationName))
```
tthurlapati-px, line 1563: Shouldn't this be `log,FailOnError()`, since if we fail to delete a backup location, there's no point in moving on with the test.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,220 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+					if err != nil {
+						return "", true, fmt.Errorf("backup location %s still present with error %v", customBackupLocationName, err)
+					}
+					if status == true {
+						return "", true, fmt.Errorf("backup location %s is not deleted yet", customBackupLocationName)
+					}
+					return "", false, nil
+				}
+				_, err = task.DoRetryWithTimeout(backupLocationDeleteStatusCheck, 3*time.Minute, 30*time.Second)
+				dash.VerifySafely(err, nil, fmt.Sprintf("Deleting backup location %s", customBackupLocationName))
```
tthurlapati-px, line 1579: Shouldn't this be `dash.VerifyFatal()` as there's no point in moving on with the test if we failed to delete a backup location.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,220 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+			}
+		})
+
+		Step("Add the backup location again which had backups", func() {
+			log.InfoD("Add the backup location with backups back")
+			for provider := range credMap {
+				for cloudCredUID, credName := range credMap[provider] {
```
tthurlapati-px, line 1586: won't the following be faster:
```go
for _,cred  := range credMap {
				for cloudCredUID, credName := range cred {
```

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,220 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+					backupLocationUID = uuid.New()
+					customBackupLocationName = fmt.Sprintf("autogenerated-backup-location-%v", time.Now().Unix())
+					backupLocationMap[backupLocationUID] = customBackupLocationName
+					err := CreateBackupLocation(provider, customBackupLocationName, backupLocationUID, credName, cloudCredUID, getGlobalBucketName(provider), orgID, "")
+					dash.VerifyFatal(err, nil, fmt.Sprintf("Creating backup location %s", customBackupLocationName))
```
tthurlapati-px, line 1591: Specifying the **credential (`credName`) in which backup location was created** is useful info, right? (as we can know which cred a backup location belonged to, in case of failure of this test)

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,220 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+				backupLocationUID = uuid.New()
+				credName = fmt.Sprintf("autogenerated-cred-%v", time.Now().Unix())
+				credMap[provider][cloudCredUID] = credName
+				err := CreateCloudCredential(provider, credName, cloudCredUID, orgID, ctx)
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying creation of cloud credential named [%s] for org [%s] with [%s] as provider", credName, orgID, provider))
+				log.InfoD("Created Cloud Credentials with name - %s", credName)
+				customBackupLocationName = fmt.Sprintf("autogenerated-backup-location-%v", time.Now().Unix())
+				err = CreateBackupLocation(provider, customBackupLocationName, backupLocationUID, credName, cloudCredUID, getGlobalBucketName(provider), orgID, "")
+				backupLocationMap[backupLocationUID] = customBackupLocationName
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Creating backup location %s", customBackupLocationName))
```
tthurlapati-px, line 1511: Specifying the **credential (`credname`) in which backup location was created** is useful info, right? (as we can know which cred a backup location belonged to, in case of failure of this test. The credname will lead up to the provider through the logs).

tthurlapati-px (changes requested):

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: I have tried it out this on;t work

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: No

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
kshithijiyer-px: @tthurlapati-px can you share a sample code on how this can be done?

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: @mkoppal-px Can you please provide some info here ^

tests/backup/backup_restore_basic_test.go
```diff
@@ -1442,3 +1442,220 @@ var _ = Describe("{BackupMultipleNsWithSameLabel}", func() {
...
+					status, err := IsBackupLocationPresent(customBackupLocationName, ctx, orgID)
+					if err != nil {
+						return "", true, fmt.Errorf("backup location %s still present with error %v", customBackupLocationName, err)
+					}
+					if status == true {
+						return "", true, fmt.Errorf("backup location %s is not deleted yet", customBackupLocationName)
+					}
+					return "", false, nil
+				}
+				_, err = task.DoRetryWithTimeout(backupLocationDeleteStatusCheck, 3*time.Minute, 30*time.Second)
```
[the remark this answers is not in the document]
kshithijiyer-px: @mkoppal-px Can you please provide some info here ^

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: What will you do with this info?

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: What's the problem with `dash.VerifyFatal()`?

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: @mkoppal-px Please share some info here ^

tests/backup/backup_restore_basic_test.go
mkoppal-px: We will know the cred name in the logs anyway, just above this line because he is calling out the cred name here.

`log.InfoD("Created Cloud Credentials with name - %s", credName)`

tests/backup/backup_restore_basic_test.go
mkoppal-px: Your question is not clear. But the objective is to fail even if one of the backups fail which will happen anyway.
`wg.Wait()` is only for the happy path, where we will ask the main thread to wait till all the go routines are done.

tests/backup/backup_restore_basic_test.go
mkoppal-px: @tthurlapati-px As I had mentioned earlier in slack as well, the main objective of the test case revolves around backup location deletion, so it is ok to use `dash.VerifyFatal` and not `log.FailOnError`

@kshithijiyer-px Please change it to `dash.VerifyFatal` because we wouldn't want to proceed with the test in case backup location deletion fails right?

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
mkoppal-px: @tthurlapati-px `defer ginkgoRecover()` is needed only when we are running in a gorountine. But here it will be running the main thread. So not needed.

tests/backup/backup_restore_basic_test.go
mkoppal-px: @kshithijiyer-px If you don't want to proceed with the test case if this task fails then we have to use `dash.VerifyFatal()`
For tasks in cleanup where we dont want to hold up the remaining cleanup tasks because the first cleanup task fails we use `dash.VerifySafely()`

tthurlapati-px: the problem is that it continues with the code, even if it fails to delete a backup location. But if we fail to delete, then we can't actually test if creating the location again, will still have the backups (after sync), right?

tests/backup/backup_restore_basic_test.go
tthurlapati-px: I guess that's intuitive enough 👍🏻

tests/backup/backup_restore_basic_test.go
tthurlapati-px: Yeah, I was basically alluding to either `dash.VerifyFatal` or `log.FailOnError`. I preferred `log.FailOnError` because this code is not the actual confirmation of deletion; this is just the command to delete. The actual verification of deletion happens later on in the code, _which is where_ `dash.VerifyFatal` is used.

tests/backup/backup_restore_basic_test.go
tthurlapati-px: Since `wg.Wait()` is only for the happy path, we can put it out of all the loops, right? (for max concurrency)

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
tthurlapati-px: @mkoppal-px but in the code for `task.DoRetryWithTimeout`, I see that the callback is being invoked in a goroutine (https://vscode.dev/github/portworx/torpedo/blob/b4594a71680262773211f1a098da604d09812545/vendor/github.com/portworx/sched-ops/task/task.go#L39)

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
tthurlapati-px: 😅 I already did in the first comment. But here's a consolidation of it:
```go
                   	curBackups, err := backupDriver.EnumerateBackup(ctx, bkpEnumerateReq)
                        log.InfoD("Check each backup for success status")
			for _, bkp = range curBackups.GetBackups() {
				backupInspectRequest := &api.BackupInspectRequest{
					Name:  bkp.Name,
					Uid:   bkp.Uid,
					OrgId: orgID,
				}
				resp, err := backupDriver.InspectBackup(ctx, backupInspectRequest)
				log.FailOnError(err, fmt.Sprintf("failed to Inspect Backup %s", bkp.Name))
				actual := resp.GetBackup().GetStatus().Status
				expected := api.BackupInfo_StatusInfo_Success
				dash.VerifyFatal(actual, expected, fmt.Sprintf("backup [%s] has success status", bkp.Name))
```
😄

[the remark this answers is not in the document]
kshithijiyer-px: Okay done.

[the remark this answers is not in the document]
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
tthurlapati-px: - After a discussion with Mithun, we confirmed that this is definitely a previously missed error. But we decided that this error needs to be fixed for many other functions, so we'll fix it in a separate PR.

- This code can be made faster as there's no need for getting `ctx` each time. **Following is faster**:
```go
                // Wait until backup location is removed
				ctx, err := backup.GetAdminCtxFromSecret()
				log.FailOnError(err, "Fetching px-central-admin ctx")
				backupLocationDeleteStatusCheck := func() (interface{}, bool, error) {
					status, err := IsBackupLocationPresent(customBackupLocationName, ctx, orgID)
					if err != nil {
						return "", true, fmt.Errorf("backup location %s still present with error %v", customBackupLocationName, err)
					}
					if status == true {
						return "", true, fmt.Errorf("backup location %s is not deleted yet", customBackupLocationName)
					}
					return "", false, nil
				}
				_, err = task.DoRetryWithTimeout(backupLocationDeleteStatusCheck, 3*time.Minute, 30*time.Second)
```

tthurlapati-px (approved):
Other than the follow two "speedups", everything else looks good to me!

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
kshithijiyer-px: okay done.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Okay will do it.

tests/backup/backup_restore_basic_test.go
```diff
@@ -1603,3 +1603,224 @@ var _ = Describe("{MultipleCustomRestoreSameTimeDiffStorageClassMapping}", func(
...
+				credMap[provider] = make(map[string]string)
+				credMap[provider][cloudCredUID] = credName
+				err := CreateCloudCredential(provider, credName, cloudCredUID, orgID, ctx)
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying creation of cloud credential named [%s] for org [%s] with [%s] as provider", credName, orgID, provider))
+				log.InfoD("Created Cloud Credentials with name - %s", credName)
+				customBackupLocationName = fmt.Sprintf("autogenerated-backup-location-%v", time.Now().Unix())
+				err = CreateBackupLocation(provider, customBackupLocationName, backupLocationUID, credName, cloudCredUID, getGlobalBucketName(provider), orgID, "")
+				backupLocationMap[backupLocationUID] = customBackupLocationName
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Creating backup location %s", customBackupLocationName))
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Creating backup location %s", customBackupLocationName))
```
lsrinivas-pure, line 1675: delete one verifyfatal

tests/backup/backup_restore_basic_test.go
```diff
@@ -1603,3 +1603,224 @@ var _ = Describe("{MultipleCustomRestoreSameTimeDiffStorageClassMapping}", func(
...
+				backupLocationUID = uuid.New()
+				credName = fmt.Sprintf("autogenerated-cred-%v", time.Now().Unix())
+				credMap[provider] = make(map[string]string)
+				credMap[provider][cloudCredUID] = credName
+				err := CreateCloudCredential(provider, credName, cloudCredUID, orgID, ctx)
+				dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying creation of cloud credential named [%s] for org [%s] with [%s] as provider", credName, orgID, provider))
+				log.InfoD("Created Cloud Credentials with name - %s", credName)
+				customBackupLocationName = fmt.Sprintf("autogenerated-backup-location-%v", time.Now().Unix())
+				err = CreateBackupLocation(provider, customBackupLocationName, backupLocationUID, credName, cloudCredUID, getGlobalBucketName(provider), orgID, "")
+				backupLocationMap[backupLocationUID] = customBackupLocationName
```
lsrinivas-pure, line 1673: move this statement after err validation

tests/backup/backup_restore_basic_test.go
```diff
@@ -1603,3 +1603,224 @@ var _ = Describe("{MultipleCustomRestoreSameTimeDiffStorageClassMapping}", func(
...
+			log.FailOnError(err, "Getting a list of all backups")
+			dash.VerifyFatal(len(fetchedBackupNames), len(backupNames), "Comparing the expected and actual number of backups")
+			var bkp *api.BackupObject
+			backupDriver := Inst().Backup
+			bkpEnumerateReq := &api.BackupEnumerateRequest{
+				OrgId: orgID,
+			}
+			ctx, err := backup.GetAdminCtxFromSecret()
+			log.FailOnError(err, "Fetching px-central-admin ctx")
+			curBackups, err := backupDriver.EnumerateBackup(ctx, bkpEnumerateReq)
```
lsrinivas-pure, line 1799: validate the err here

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Ah copy pasted multiple times!

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
```diff
@@ -2075,3 +2076,224 @@ var _ = Describe("{ManualAndScheduleBackupUsingNamespaceLabel}", func() {
...
+					status, err := IsBackupLocationPresent(customBackupLocationName, ctx, orgID)
+					if err != nil {
+						return "", true, fmt.Errorf("backup location %s still present with error %v", customBackupLocationName, err)
+					}
+					if status == true {
+						return "", true, fmt.Errorf("backup location %s is not deleted yet", customBackupLocationName)
+					}
+					return "", false, nil
+				}
+				_, err = task.DoRetryWithTimeout(backupLocationDeleteStatusCheck, 3*time.Minute, 30*time.Second)
```
tthurlapati-px, line 2214: Use `DoRetryWithTimeoutWithGinkgoRecover` (Mithun mentioned that all non-library functions must use for consistency)

tests/backup/backup_restore_basic_test.go
```diff
@@ -2075,3 +2076,224 @@ var _ = Describe("{ManualAndScheduleBackupUsingNamespaceLabel}", func() {
...
+				fetchedBackupNames, err := GetAllBackupsAdmin()
+				if err != nil {
+					return "", true, fmt.Errorf("unable to fetch backups. Error: %s", err.Error())
+				}
+				if len(fetchedBackupNames) == len(backupNames) {
+					return "", false, nil
+				}
+				return "", true, fmt.Errorf("expected: %d and actual: %d", len(backupNames), len(fetchedBackupNames))
+			}
+			_, err := task.DoRetryWithTimeout(checkBackupSync, 100*time.Minute, 30*time.Second)
```
tthurlapati-px, line 2259: same modification as above

tthurlapati-px (changes requested):
some modifications needed

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tthurlapati-px (approved):
LGTM

tests/backup/backup_restore_basic_test.go
```diff
@@ -2229,3 +2230,224 @@ var _ = Describe("{MultipleInPlaceRestoreSameTime}", func() {
 		CleanupCloudSettingsAndClusters(backupLocationMap, cloudCredName, cloudCredUID, ctx)
 	})
 })
+
+// CloudSnapsSafeWhenBackupLocationDeleteTest takes a good number of backups to check if cloud snaps are
+// safe (not deleted) if backup location is removed
+var _ = Describe("{CloudSnapsSafeWhenBackupLocationDeleteTest}", func() {
+	numberOfBackups, _ := strconv.Atoi(getEnv(maxBackupsToBeCreated, "10"))
```
mkoppal-px, line 2237: Should we add this as part of `deployment.sh`

mkoppal-px (approved):
LGTM

tthurlapati-px (dismissed):
Sorry about merge-blocking, but I just found something problematic:

tests/backup/backup_restore_basic_test.go
```diff
@@ -2229,3 +2230,224 @@ var _ = Describe("{MultipleInPlaceRestoreSameTime}", func() {
...
+							defer wg.Done()
+							defer func() { <-sem }()
+							err = CreateBackup(backupName, SourceClusterName, backupLocationName, backupLocationUID, []string{namespace},
+								labelSelectors, orgID, clusterUid, "", "", "", "", ctx)
+							dash.VerifyFatal(err, nil, fmt.Sprintf("Verifying backup creation: %s", backupName))
+						}(backupName)
+					}
+				}
+				wg.Wait()
+			}
```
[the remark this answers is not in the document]
kshithijiyer-px: @tthurlapati-px If I go by what you are saying why is it not failing for all cases https://github.com/portworx/torpedo/blob/master/tests/backup/backup_restore_basic_test.go#L1212 ? The same code is being using at multiple places and seem to be working fine without whatever changes you are suggesting.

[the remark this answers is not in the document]
tthurlapati-px: In the particular example you have shown over here the **flaw does exist**, but it doesn't manifest because `wg.wait()` is being used such that `namespace` variable will remain the same until all corresponding goroutines are done, thus averting a race condition.

But the example code can also be further optimized and made faster by moving `wg.Wait()` outside the outer for loop, **and** make the change to goroutine as `go func(backupName, namespace string) {` in order to prevent race condition.

So, my comments on the current code are still valid

tthurlapati-px (changes requested):
I have added information to prove relevance of request.

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
kshithijiyer-px: @tthurlapati-px I get your point but there is no point of fixing it in just a single place. We can't do all the optimisation in a single PR. We can take this forward in the next one and do it at all places. I see you have approved this PR 3 times and went back and opened new comments 3 times. This is of no use and the PR will always be in an infinite loop forever and never get merged. We can check if whatever you are saying works and then we can do these changes at all places instead of doing it at just once place which doesn't really help.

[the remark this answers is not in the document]
tthurlapati-px: While I understand what you're saying about the _optimization_ part of things, even if you don't optmize the `wg.Wait()`, there **still** exists a bug that you may have slipped past you. 

If you compare the example code you gave and your own current code, you'll notice that the example code ensures that `wg.Wait()` prevents race conditions as `namespace` doesn't change until all goroutine are complete.

But in your code, `namespace` **does** change before all goroutines complete, keeping the window open for race conditions. I'm not sure why you haven't hit an error. (is it because of limiting the numberofsimultneousbackups?)

tthurlapati-px (changes requested):
I have explained this in the comment, but if you test your code with two namespace (`postgres-backup,mysql-backup`) at the same time, you will most certainly hit a bug. But if we are not testing with two apps simultaneously (I'm not aware what is tested in nightly), then we can ignore this bug and I'll approve 👍🏻 .

tthurlapati-px (approved):
LGTM

tests/backup/backup_restore_basic_test.go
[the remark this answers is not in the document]
tthurlapati-px: I approved the PR.
The optimizations can be done later 👍🏻

PR #1305, opened 2023-04-05
Bug Fix for Independent App TC

dbhatnagar-px:
After old code purging, looks like this test case was still not using the new module to create deployments and hence it was facing issues

Jenkins Run: https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/193/


Madanagopal19 (approved):
LGTM

aravindranath14 (approved):
LGTM

PR #1307, opened 2023-04-05
Wait for valid uptime after VM reset

Santhoshairody:
There is a known issue when we reboot worker node through debug pod on a OCP cluster configured with custom CIDR.
https://access.redhat.com/support/cases/#/case/03435166
Hence need to reboot worker node through VM guest reset, it has been  observed that linux command "uptime" takes time to load after VMReset, this code changes will wait for valid uptime.

Closes #

Output : OCP 4.12.x
https://jenkins.pwx.dev.purestorage.com/job/Torpedo/view/Torpedo%20NextPX/job/tp-ocp-dev-csi/72/
OCP 4.11.x
https://jenkins.pwx.dev.purestorage.com/job/Torpedo/view/Torpedo%20NextPX/job/tp-ocp-candidate/112/consoleFull


drivers/node/ssh/ssh.go
```diff
@@ -86,6 +87,29 @@ func (s *SSH) IsUsingSSH() bool {
...
+	t := func() (interface{}, bool, error) {
+		out, err := s.doCmd(n, node.ConnectionOpts{
+			Timeout:         1 * time.Minute,
+			TimeBeforeRetry: 10 * time.Second,
+		}, uptimeCmd, true)
+		return out, true, err
+	}
+	uptimecall := func() (interface{}, bool, error) {
+		out, err := task.DoRetryWithTimeout(t, 1*time.Minute, 10*time.Second)
+		if err != nil {
```
Rohit-PX, line 102:
```suggestion
		if err != nil || if len(strings.Fields(strings.TrimSpace(out.(string)))) == 0 {
```

Rohit-PX (approved):
LGTM. Except for the minor change.

drivers/node/ssh/ssh.go
```diff
@@ -86,6 +87,29 @@ func (s *SSH) IsUsingSSH() bool {
...
 
+// WaitForValidUptime Wait for valid uptime.
+func (s *SSH) WaitForValidUptime(n node.Node) (interface{}, error) {
+	uptimeCmd := "sudo uptime -s"
+	t := func() (interface{}, bool, error) {
+		out, err := s.doCmd(n, node.ConnectionOpts{
+			Timeout:         1 * time.Minute,
+			TimeBeforeRetry: 10 * time.Second,
+		}, uptimeCmd, true)
+		return out, true, err
```
lsrinivas-pure, line 98: is it always retry true?
don't we need a false condition?

drivers/node/ssh/ssh.go
```diff
@@ -86,6 +87,29 @@ func (s *SSH) IsUsingSSH() bool {
 	return len(os.Getenv("TORPEDO_SSH_KEY")) > 0 || len(os.Getenv("TORPEDO_SSH_PASSWORD")) > 0
 }
 
+// WaitForValidUptime Wait for valid uptime.
+func (s *SSH) WaitForValidUptime(n node.Node) (interface{}, error) {
+	uptimeCmd := "sudo uptime -s"
+	t := func() (interface{}, bool, error) {
+		out, err := s.doCmd(n, node.ConnectionOpts{
+			Timeout:         1 * time.Minute,
```
lsrinivas-pure, line 95: can you check if you can use RunCommand function, I think it does the same thing

drivers/node/ssh/ssh.go
```diff
@@ -106,7 +130,14 @@ func (s *SSH) IsNodeRebootedInGivenTimeRange(n node.Node, timerange time.Duratio
 			Cause: fmt.Sprintf("Failed to run uptime command in node %v", n),
 		}
 	}
-
+	if len(strings.TrimSpace(out.(string))) == 0 {
```
jainsmit, line 133: This method needs to be executed after node is up so why needs to be wait inside it and not before invoking this method ?

drivers/node/ssh/ssh.go
```diff
@@ -86,6 +87,29 @@ func (s *SSH) IsUsingSSH() bool {
 	return len(os.Getenv("TORPEDO_SSH_KEY")) > 0 || len(os.Getenv("TORPEDO_SSH_PASSWORD")) > 0
 }
 
+// WaitForValidUptime Wait for valid uptime.
+func (s *SSH) WaitForValidUptime(n node.Node) (interface{}, error) {
```
nikolaypopov, line 91: Let's not create an extra func for this, we already have this logic on line 92, let's just improve the existing logic to retry and return only when you get the uptime out not empty and no errors

drivers/node/ssh/ssh.go
Santhoshairody: Changed logic and now extra function is not required.

drivers/node/ssh/ssh.go
Santhoshairody: Changed logic and now extra function is not required.

drivers/node/ssh/ssh.go
Santhoshairody: changed logic now.

drivers/node/ssh/ssh.go
Santhoshairody: Now logic is incorporated in a  single method.

nikolaypopov (approved):
LGTM

Santhoshairody:
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Santosh/job/BasicTorpedo-custom/87/consoleFull

PR #1319, opened 2023-04-12
Fix for ValidatePDSHealthInCaseOfFailures testcase

Madanagopal19:
Fixes the race condition of not catching the health status when there is a change in health status of data services

logs: https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/214/


aravindranath14 (approved):
LGTM

PR #1320, opened 2023-04-12
Ptx 16727

vprabhakar-px:
this test is for ford customer escalations , where one or more volumes go in to out of quorum state ,


tests/basic/customer_issues.go
```diff
@@ -0,0 +1,352 @@
...
+
+		time.Sleep(2 * time.Minute)
+		ValidateApplications(contexts)
+		defer appsValidateAndDestroy(contexts)
+
+		var volumesPresent []*volume.Volume
+
+		for _, ctx := range contexts {
+			vols, err := Inst().S.GetVolumes(ctx)
+			if err != nil {
```
lsrinivas-pure, line 183: if is not required here

tests/basic/customer_issues.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+
+		// Check if the cluster consists of 10 nodes
+		log.InfoD("Get all nodes present in the cluster")
+
+		allNodes := []node.Node{}
+		// create an array with storage and storage less nodes added
+		for _, each := range node.GetStorageNodes() {
+			allNodes = append(allNodes, each)
+		}
```
lsrinivas-pure, line 200: getstoragedrivernodes will give you all the nodes in the setup

tests/basic/customer_issues.go
```diff
@@ -0,0 +1,352 @@
...
+
+		for _, each := range node.GetStorageLessNodes() {
+			allNodes = append(allNodes, each)
+		}
+
+		// Verify total nodes available is 10
+		dash.VerifyFatal(len(allNodes) == 10, true, "required minimum of 10 nodes for the test to run")
```
lsrinivas-pure, line 207: make it >=10. also can we add this check even before deploying apps?

tests/basic/customer_issues.go
```diff
@@ -0,0 +1,352 @@
...
+				}
+			}
+		}
+
+		// force flush iptables on all the nodes at the end
+		defer flushiptables()
+
+		volumeState, err := GetVolumesInDegradedState(contexts)
+		if err != nil {
+			fmt.Println("Error not Nil")
```
lsrinivas-pure, line 274: remove these print statements

tests/basic/customer_issues.go
```diff
@@ -0,0 +1,352 @@
...
+		log.InfoD("Killing KVDB PID from KVDB Master Node")
+		log.FailOnError(KillKvdbNode(getKvdbLeaderNode), "failed to Kill Kvdb Node")
+		// Block IPtable rules on the kvdb node to all the nodes in zone 1
+		kvdb := []node.Node{getKvdbLeaderNode}
+
+		time.Sleep(10 * time.Minute)
+		log.FailOnError(blockIptableRules(kvdb, zone1, false), "Set IPTable rules on kvdb node failed")
+
+		// Wait for some time before checking for file system goes back online
+		time.Sleep(30 * time.Minute)
```
lsrinivas-pure, line 317: add log statments for these long waits. like how much time you are waiting and why are you waiting

tests/basic/storage_pool_test.go
```diff
@@ -6359,8 +6360,8 @@ var _ = Describe("{VerifyPoolDeleteInvalidPoolID}", func() {
 		}
 
 		commonText := "service mode delete pool.*unable to delete pool with ID.*[0-9]+.*cause.*"
-		compileText := fmt.Sprintf("[%s]operation is not supported", commonText)
-		compileTextMaintenanceError := fmt.Sprintf("[%s]Requires pool maintenance mode", commonText)
+		compileText := fmt.Sprintf("%soperation is not supported", commonText)
```
lsrinivas-pure, line 6363: add space after indentifier

tests/basic/storage_pool_test.go
```diff
@@ -8609,6 +8610,27 @@ var _ = Describe("{DriveAddAsJournal}", func() {
...
 })
 
+func waitTillVolumeStatusUp(vol *volume.Volume) error {
+	now := time.Now()
+	targetTime := now.Add(30 * time.Minute)
+
+	for {
+		if now.After(targetTime) {
+			log.Error("Failed as the timeout of 0 Min is reached before resync triggered")
+			return fmt.Errorf("Failed to get volume status to UP")
```
lsrinivas-pure, line 8620: just return the error here

tests/common.go
```diff
@@ -6423,3 +6424,134 @@ func DoRetryWithTimeoutWithGinkgoRecover(taskFunc func() (interface{}, bool, err
...
+		var kvdbNodes []kvdbNode
+
+		err := json.Unmarshal([]byte(fmt.Sprintf("[%s]", jsonString)), &nodes)
+		if err != nil {
+			return nil, err
+		}
+		for _, nodeMap := range nodes {
+			for id, value := range nodeMap {
+				kvdb.ID = id
+				kvdb.Leader = value.Leader
```
lsrinivas-pure, line 6496: no need to explicitly convert, jsin unmarshal will take care of it if &kvdbnodes are given

lsrinivas-pure (changes requested):

tests/basic/customer_issues.go
vprabhakar-px: this is done to get the list of volumes present in the cluster at the start of the script , 
this is to avoid the failure in case if pods are not accessible after running the script ,

I have seen the instance where few pods goes to error state and later errors out while getting the volume details

tests/basic/customer_issues.go
vprabhakar-px: fixed !

tests/basic/customer_issues.go
vprabhakar-px: removed the complete section , thanks for pointing out , this was added as a debug statement

tests/basic/customer_issues.go
vprabhakar-px: done

tests/basic/storage_pool_test.go
vprabhakar-px: this is a regex expression , so ".*cause.*operation"

drivers/node/node_registry.go
```diff
@@ -77,6 +77,16 @@ func GetMasterNodes() []Node {
 	return nodeList
 }
 
+// IsMasterNode returns true is node is Masternode
```
jainsmit, line 80: Please correct the comment, it should be `IsMasterNode returns true if node is a Masternode` instead of `IsMasterNode returns true is node is Masternode`

vprabhakar-px: fixed

tests/common.go
vprabhakar-px: fixed

tests/basic/customer_issues.go
```diff
@@ -0,0 +1,350 @@
...
+		// Check if the cluster consists of 10 nodes
+		log.InfoD("Get all nodes present in the cluster")
+		allNodes := []node.Node{}
+		// create an array with storage and storage less nodes added
+		for _, each := range node.GetStorageDriverNodes() {
+			allNodes = append(allNodes, each)
+		}
+
+		for _, each := range node.GetStorageLessNodes() {
+			allNodes = append(allNodes, each)
```
lsrinivas-pure, line 179: you can remove this as GetStorageDriverNodes() will include storageless nodes also

tests/basic/customer_issues.go
lsrinivas-pure: what I meant is you can directly use log.FailOnError(err, "failed to get volumes from the contexts"). It will check whether err is nil or not., no need of if check here

tests/common.go
```diff
@@ -5994,3 +5995,140 @@ func GetVolumesFromPoolID(contexts []*scheduler.Context, poolUuid string) ([]*vo
 	}
 	return volumes, nil
 }
+
+// DoRetryWithTimeoutWithGinkgoRecover calls `task.DoRetryWithTimeout` along with `ginkgo.GinkgoRecover()`, to be used in callbacks with panics or ginkgo assertions
+func DoRetryWithTimeoutWithGinkgoRecover(taskFunc func() (interface{}, bool, error), timeout, timeBeforeRetry time.Duration) (interface{}, error) {
```
lsrinivas-pure, line 6000: I dont think this is you change. If not  can you rebase in your local and check once

lsrinivas-pure (approved):
LGTM

PR #1324, opened 2023-04-13
Fixdelpvc jk

jkagliwal-px:
Fixed PV and PVC level cleanup for every testcase post deleting the created deployment.

Failure cause -> delete deployment is called twice due to which the pods in "Terminating" state are not recognised and the test exits.

Fix provided -> Added a flag to check if the deployment is already deleted. 

Jenkins Run :

https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/223/console


gauraa2 (approved):
LGTM

Madanagopal19 (approved):
LGTM

PR #1329, opened 2023-04-17
Resiliency Test: Reboot Node During PDS App Version Update

Madanagopal19:
1. Trigger app version upgrade through PDS
2. While app upgrade is happening, reboot one of the nodes thats hosting the app
3. Once app upgrade completes, trigger a workload and verify data

logs: https://aetos.pwx.purestorage.com/resultSet/testSetID/164643

### DIFF
--- a/tests/pds/dataservice_test.go
+++ b/tests/pds/dataservice_test.go
@@ -1120,9 +1120,11 @@ func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID 
 		if deployOldVersion {
 			dsVersion = ds.OldVersion
 			dsImage = ds.OldImage
+			log.Debugf("Deploying old version %s and image %s", dsVersion, dsImage)
 		} else {
 			dsVersion = ds.Version
 			dsImage = ds.Image
+			log.Debugf("Deploying latest version %s and image %s", dsVersion, dsImage)
 		}
 
 		log.InfoD("Deploying DataService %v ", ds.Name)
--- a/tests/pds/dataservice_test.go
+++ b/tests/pds/dataservice_test.go
@@ -1112,8 +1112,19 @@ var _ = Describe("{DeployAllDataServices}", func() {
 	})
 })
 
-func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID string) (*pds.ModelsDeployment, map[string][]string, map[string][]string, error) {
+func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID string, deployOldVersion bool) (*pds.ModelsDeployment, map[string][]string, map[string][]string, error) {
 	Step("Deploy Data Services", func() {
+		var dsVersion string
+		var dsImage string
+
+		if deployOldVersion {
+			dsVersion = ds.OldVersion
+			dsImage = ds.OldImage
+		} else {
+			dsVersion = ds.Version
+			dsImage = ds.Image
+		}
+
 		log.InfoD("Deploying DataService %v ", ds.Name)
 		dataServiceDefaultResourceTemplateID, err = pdslib.GetResourceTemplate(tenantID, ds.Name)
 		log.FailOnError(err, "Error while getting resource template")
@@ -1139,8 +1150,8 @@ func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID 
 			serviceType,
 			dataServiceDefaultResourceTemplateID,
 			storageTemplateID,
-			ds.Version,
-			ds.Image,
+			dsVersion,
+			dsImage,
 			namespace,
 		)
 		log.FailOnError(err, "Error while deploying data services")
@@ -1150,7 +1161,7 @@ func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID 
 
 func DeployandValidateDataServices(ds PDSDataService, namespace, tenantID, projectID string) (*pds.ModelsDeployment, map[string][]string, map[string][]string, error) {
 	log.InfoD("Data Service Deployment Triggered")
-	deployment, dataServiceImageMap, dataServiceVersionBuildMap, err := TriggerDeployDataService(ds, namespace, tenantID, projectID)
+	deployment, dataServiceImageMap, dataServiceVersionBuildMap, err := TriggerDeployDataService(ds, namespace, tenantID, projectID, false)
 	Step("Validate Data Service Configurations", func() {
 		err = pdslib.ValidateDataServiceDeployment(deployment, namespace)
 		log.FailOnError(err, fmt.Sprintf("Error while validating dataservice deployment %v", *deployment.ClusterResourceName))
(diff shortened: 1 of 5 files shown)


tests/pds/dataservice_test.go
```diff
@@ -1238,67 +1251,73 @@ func UpgradeDataService(dataservice, oldVersion, oldImage, dsVersion, dsBuild st
...
+		log.Infof("Version/Build: %v %v", dsVersion, dsBuild)
+		updatedDeployment, err := pdslib.UpdateDataServiceVerison(deployment.GetDataServiceId(), deployment.GetId(),
+			dataServiceDefaultAppConfigID,
+			replicas, dataServiceDefaultResourceTemplateID, dsBuild, dsVersion)
+		log.FailOnError(err, "Error while updating data services")
+		log.InfoD("data service deployed %v ", dataservice)
 
-			resourceTemp, storageOp, config, err := pdslib.ValidateDataServiceVolumes(updatedDeployment, dataservice, dataServiceDefaultResourceTemplateID, storageTemplateID, namespace)
-			log.FailOnError(err, "error on ValidateDataServiceVolumes method")
+		err = pdslib.ValidateDataServiceDeployment(updatedDeployment, namespace)
```
dbhatnagar-px, line 1292: For now it's ok, but it's becoming a big mess soon. We should not be incorporating logic in test case file. Let's soon re-design our test case writing to just have a absolutely neccessary number of calls to functions that does all the magic and make it extremely modular. Plus let's remove all functions from test case file as well. 
@gauraa2, @Madanagopal19  & @jkagliwal-px : Let's sit in a meeting and re-design the tests. This model will cause our test case files to go beyond tens of thousands of lines soon - making the tests totally unmanageable. 

^ This is a generic comment not specific to this change. For now it is ok, but we will change it soon.

tests/pds/dataservice_test.go
```diff
@@ -1157,8 +1157,21 @@ var _ = Describe("{DeployAllDataServices}", func() {
 	})
 })
 
-func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID string) (*pds.ModelsDeployment, map[string][]string, map[string][]string, error) {
+func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID string, deployOldVersion bool) (*pds.ModelsDeployment, map[string][]string, map[string][]string, error) {
```
dbhatnagar-px, line 1160: Function UpgradeDataService is doing a similar task. Let's try to reuse either that or make changes in that to re-use this code.

tests/pds/resiliency_test.go
```diff
@@ -165,6 +165,131 @@ var _ = Describe("{RebootActiveNodeDuringDeployment}", func() {
...
+				failuretype := pdslib.TypeOfFailure{
+					Type: ActiveNodeRebootDuringDeployment,
+					Method: func() error {
+						return pdslib.RebootActiveNodeDuringDeployment(params.InfraToTest.Namespace)
+					},
+				}
+				pdslib.DefineFailureType(failuretype)
+
+				log.Infof("Version/Build: %v %v", ds.Version, ds.Image)
+				updatedDeployment, err := pdslib.UpdateDataServiceVerison(deployment.GetDataServiceId(), deployment.GetId(),
```
dbhatnagar-px, line 220: Won't this wait till Update is successful due to implicit call to "ValidateDataServiceDeployment" ? Your use case is to come out after just triggering update. Am I wrong ?

tests/pds/resiliency_test.go
```diff
@@ -165,6 +165,131 @@ var _ = Describe("{RebootActiveNodeDuringDeployment}", func() {
...
+					dataServiceDefaultAppConfigID,
+					int32(ds.Replicas), dataServiceDefaultResourceTemplateID, ds.Image, ds.Version)
+				log.FailOnError(err, "Error while updating data services")
+				log.InfoD("data service %v update triggered", ds.Name)
+
+				err = pdslib.InduceFailureAfterWaitingForCondition(updatedDeployment, namespace, params.ResiliencyTest.CheckTillReplica)
```
dbhatnagar-px, line 226: Issue with this line is , GetPdsSs will always return true immediately in this case without validating that upgrade has triggered or not - as there will be more than 1 pod active here for any multi-node deployment.

dbhatnagar-px (changes requested):

tests/pds/dataservice_test.go
Madanagopal19: sure

tests/pds/dataservice_test.go
Madanagopal19: yes I have checked it, but UpgradeDataService function is having lot of validations which is not required and the easiest way of implementation is with this method.

tests/pds/resiliency_test.go
Madanagopal19: I have removed the `ValidateDataServiceDeployment ` from `UpdateDataServiceVerison ` method so we are good here.

tests/pds/resiliency_test.go
Madanagopal19: I have triggered the `UpdateDataServiceVerison ` method just before `InduceFailureAfterWaitingForCondition ` method, and that method handles the issue you mentioned `validating that upgrade has triggered or not`. For me there just be atleast 1 active pods to trigger the reboot.

tests/pds/dataservice_test.go
dbhatnagar-px: Ok, in that case let's reuse this code in UpgradeDataService function.

tests/pds/resiliency_test.go
dbhatnagar-px: Exactly what we don't want as you moved the Validation part to test case directly. But as discussed above, let's take it up later.

dbhatnagar-px (changes requested):

tests/pds/dataservice_test.go
jkagliwal-px: I had same concern ! Lets re-design.

tests/pds/dataservice_test.go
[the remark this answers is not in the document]
Madanagopal19: Lets have a story for optimisation and figure it out what can be optimised. Lets discuss and create task for it.

tests/pds/dataservice_test.go
```diff
@@ -1195,7 +1208,7 @@ func TriggerDeployDataService(ds PDSDataService, namespace, tenantID, projectID
 
 func DeployandValidateDataServices(ds PDSDataService, namespace, tenantID, projectID string) (*pds.ModelsDeployment, map[string][]string, map[string][]string, error) {
 	log.InfoD("Data Service Deployment Triggered")
-	deployment, dataServiceImageMap, dataServiceVersionBuildMap, err := TriggerDeployDataService(ds, namespace, tenantID, projectID)
+	deployment, dataServiceImageMap, dataServiceVersionBuildMap, err := TriggerDeployDataService(ds, namespace, tenantID, projectID, false)
```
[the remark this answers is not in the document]
Madanagopal19: Its updated in all affected places and those changes are part of this PR

dbhatnagar-px (approved):
Lgtm

aravindranath14 (approved):
Looks good

PR #1349, opened 2023-04-26
PA-899: Add libraries for large resource creation

kshithijiyer-px:
Adding the below functions to create large resource backup and restores: 
- [x]  Add func createNamespaces() to create large number of namespaces as per need. 
- [x]  Add func createConfigMaps() to create large number of Configmaps in a set of namespaces. 
- [x]  Add func createSecrets() to create large number of secrets in a set of namespaces. 
- [x]  Add func deleteNamespaces() to delete a set of namespaces. 

**Things I'll take up in the next PR:**
1. Add func createApp() deployments to deploy a specific app multiple times with volume on a given set of namespaces N number of times. 

Closes # PA-899

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -6622,16 +6622,6 @@ func AddMetadataDisk(n node.Node) error {
 
 }
 
-// generateRandomString Creates a random string of length n and returns it
-func generateRandomString(n int) string {
-	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
-	randomString := make([]rune, n)
-	for i := range randomString {
-		randomString[i] = letters[rand.Intn(len(letters))]
-	}
-	return string(randomString)
-}
-
 // createNamespaces Create N number of namespaces and return namespace list
 func createNamespaces(numberOfNamespaces int) ([]string, error) {
 
@@ -6665,12 +6655,13 @@ func createConfigMaps(namespaces []string, numberOfConfigmap int, numberOfEntrie
 	// Create random data to add in ConfigMap
 	var randomData = make(map[string]string)
 	for i := 0; i < numberOfEntries; i++ {
-		randomString := generateRandomString(80)
+		randomString := RandomString(80)
 		randomStringWithTimeStamp := fmt.Sprintf("%v.%v", randomString, time.Now().Unix())
 		randomData[randomStringWithTimeStamp] = randomString
 	}
 
 	// create the number of configmaps needed
+	k8sCore = core.Instance()
 	for _, namespace := range namespaces {
 		for i := 0; i < numberOfConfigmap; i++ {
 			name := fmt.Sprintf("configmap-%s-%d-%v", namespace, i, time.Now().Unix())
@@ -6684,7 +6675,6 @@ func createConfigMaps(namespaces []string, numberOfConfigmap int, numberOfEntrie
 				Data:       randomData,
 			}
 
-			k8sCore = core.Instance()
 			_, err := k8sCore.CreateConfigMap(obj)
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


lsrinivas-pure (approved):
lgtm

tests/common.go
```diff
@@ -6624,3 +6625,123 @@ func AddMetadataDisk(n node.Node) error {
 	return nil
 
 }
+
+// generateRandomString Creates a random string of length n and returns it
+func generateRandomString(n int) string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	randomString := make([]rune, n)
+	for i := range randomString {
+		randomString[i] = letters[rand.Intn(len(letters))]
```
ak-px, line 6634: minor optimization, len(letters) can be cached outside loop to avoid it calling repeatedly.

tests/common.go
```diff
@@ -6624,3 +6625,123 @@ func AddMetadataDisk(n node.Node) error {
...
+			metaObj := metaV1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+			}
+			obj := &corev1.ConfigMap{
+				ObjectMeta: metaObj,
+				Data:       randomData,
+			}
+
+			k8sCore = core.Instance()
```
ak-px, line 6691: k8score  variable can be initialized outside the loop. which can reduce unnecessary initilization for every iteration.

tests/common.go
```diff
@@ -6624,3 +6625,123 @@ func AddMetadataDisk(n node.Node) error {
 	return nil
 
 }
+
+// generateRandomString Creates a random string of length n and returns it
+func generateRandomString(n int) string {
```
kphalgun-px, line 6630: common.go contains a function called RandomString which generates a random string of lowercase characters with a length specified by the length parameter.

tests/common.go
```diff
@@ -6624,3 +6625,123 @@ func AddMetadataDisk(n node.Node) error {
...
+			metaObj := metaV1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+			}
+			obj := &corev1.Secret{
+				ObjectMeta: metaObj,
+				StringData: randomData,
+			}
+
+			k8sCore = core.Instance()
```
ak-px, line 6726: same as above.

tests/common.go
```diff
@@ -6624,3 +6625,123 @@ func AddMetadataDisk(n node.Node) error {
...
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	randomString := make([]rune, n)
+	for i := range randomString {
+		randomString[i] = letters[rand.Intn(len(letters))]
+	}
+	return string(randomString)
+}
+
+// createNamespaces Create N number of namespaces and return namespace list
+func createNamespaces(numberOfNamespaces int) ([]string, error) {
```
kphalgun-px, line 6640: Is it possible to parallelize this task?

tests/common.go
kshithijiyer-px: Okay let's remove it.

tests/common.go
kshithijiyer-px: No longer needed as I am using the existing `RandomString()` function

tests/common.go
kshithijiyer-px: Done

tests/common.go
kshithijiyer-px: Done

ak-px (approved):
LGTM

tests/common.go
kshithijiyer-px: I had a look at the `TaskHandler()` function in your PR and had a look at go routine as well. Based on this what I felt we should ideally use in the step. Say lets say you want to create 100 configmaps You can run a go rountine or the TaskHnadler which calls these functions from within the step and creates 10 configmaps at a time. This will help make sure that the k8s api isn't overloaded and also reduce time. In the core function I feel it's better to have a dumb logic.

mkoppal-px: @kshithijiyer-px We can go ahead with the current implementation without `TaskHandler()`.
Once it is merged into master we can optimize it for faster parallel execution.

mkoppal-px (approved):
LGTM

kphalgun-px (approved):
LGTM

tests/common.go
kshithijiyer-px: ACK

PR #1360, opened 2023-04-29
Added cluster and ctx parameter support for DeleteSchedule utility

ak-px:
Currently DeleteSchedule which delete the backup schedules doesnt take any cluster parameter and by default delete schedules only on source cluster, with this PR added support to pass clusterName as parameter and will be able to delete schedules on any cluster..

Also made the code to fetch schedule uid within `DeleteSchedule` Function.

Closes #PA-814

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/1252/consoleFull

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -3265,32 +3265,38 @@ func DeleteBackupLocation(name string, backupLocationUID string, orgID string, D
 }
 
 // DeleteSchedule deletes backup schedule
-func DeleteSchedule(backupScheduleName, backupScheduleUID, OrgID string) error {
+func DeleteSchedule(backupScheduleName string, clusterName string, orgID string, ctx context1.Context) error {
 	backupDriver := Inst().Backup
+	backupScheduleInspectRequest := &api.BackupScheduleInspectRequest{
+		Name:  backupScheduleName,
+		Uid:   "",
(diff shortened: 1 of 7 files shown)


tests/common.go
```diff
@@ -3263,7 +3263,7 @@ func DeleteBackupLocation(name string, backupLocationUID string, orgID string, D
 }
 
 // DeleteSchedule deletes backup schedule
-func DeleteSchedule(backupScheduleName, backupScheduleUID, OrgID string) error {
+func DeleteSchedule(backupScheduleName, backupScheduleUID, clusterName, OrgID string) error {
```
kphalgun-px, line 3266: This might be out of scope for this PR. How about we pass ctx as well?

kphalgun-px: Thanks.

ak-px: It good have the ctx as well, Included it.
Thanks for suggestion.

kphalgun-px (approved):
LGTM

kshithijiyer-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #1370, opened 2023-05-08
PA-1013: Fix cleanup in RebootNodesWhenBackupsAreInProgress

kshithijiyer-px:
Link: https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/px-backup-system-test-rke/100/consoleFull 
Atoes Dashboard: https://aetos.pwx.purestorage.com/resultSet/testSetID/184291

`RebootNodesWhenBackupsAreInProgress` failed on `ValidateAndDestroy` as the postgres pods where not up, post which all the other testcases failed as the subsequent context switch happening [here](https://github.com/portworx/torpedo/blob/master/tests/backup/backup_resiliency_test.go#L1317) wasn't executed.  If I check the current state of the pods in the namespace they are still in `CrashLoopBackOff`:

```
postgres-backup-pxbackuptask-0-05-07-12h53m58s   postgres-646f6f7487-8vfcm                                      0/1     CrashLoopBackOff   258 (105s ago)   15h
```

**Fix done in this PR:**
- [x] Add a func `DestroyApps()` which destroys the apps without validating it. 
- [x] Move the context switch inside `defer func() { }` block. 
- [x] Change the function call from  `ValidateAndDestroy()` to `DestroyApps()` in`RebootNodesWhenBackupsAreInProgress` test. 

Dashboard URL:
https://aetos.pwx.purestorage.com/resultSet/testSetID/185172

Follow up ticket to add validation: https://portworx.atlassian.net/browse/PA-1018

Closes # PA-1013

### DIFF
--- a/tests/backup/backup_resiliency_test.go
+++ b/tests/backup/backup_resiliency_test.go
@@ -1290,6 +1290,10 @@ var _ = Describe("{RebootNodesWhenBackupsAreInProgress}", func() {
 	})
 
 	JustAfterEach(func() {
+		defer func() {
+			err := SetSourceKubeConfig()
+			log.FailOnError(err, "Switching context to source cluster")
+		}()
 		defer EndPxBackupTorpedoTest(contexts)
 		log.InfoD("Check if the rebooted nodes on application cluster are up now")
 		log.Infof("Switching cluster context to destination cluster")
@@ -1312,7 +1316,7 @@ var _ = Describe("{RebootNodesWhenBackupsAreInProgress}", func() {
 		log.Infof("Deleting the deployed applications on application cluster")
 		opts := make(map[string]bool)
 		opts[SkipClusterScopedObjects] = true
-		ValidateAndDestroy(contexts, opts)
+		DestroyApps(contexts, opts)
 		log.Infof("Switching cluster context back to source cluster")
 		err = SetSourceKubeConfig()
 		log.FailOnError(err, "Switching context to source cluster")
--- a/tests/common.go
+++ b/tests/common.go
@@ -1790,6 +1790,16 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
 	})
 }
 
+// DestroyApps destroy applications without validating them
+func DestroyApps(contexts []*scheduler.Context, opts map[string]bool) {
+	Step("destroy apps", func() {
+		log.InfoD("Destroying apps")
+		for _, ctx := range contexts {
+			TearDownContext(ctx, opts)
+		}
+	})
+}
+
 // AddLabelsOnNode adds labels on the node
 func AddLabelsOnNode(n node.Node, labels map[string]string) error {
 	for labelKey, labelValue := range labels {


tests/backup/backup_resiliency_test.go
```diff
@@ -1312,10 +1312,12 @@ var _ = Describe("{RebootNodesWhenBackupsAreInProgress}", func() {
 		log.Infof("Deleting the deployed applications on application cluster")
 		opts := make(map[string]bool)
 		opts[SkipClusterScopedObjects] = true
-		ValidateAndDestroy(contexts, opts)
+		DestroyApps(contexts, opts)
```
ak-px, line 1315: Can we have all testcases replaced with DestroyApps , I think this ValidateAndDestroy takes 2-3 mins of execution, if we are replacing it in all testcases it would save significant time from the overall execution time of nightly run.

kshithijiyer-px: @ak-px it would be a good improvement  to do but not in this PR. This PR is mostly to do a quick fix in this testcase only.

ak-px: Fine, It can be taken in another PR. Thanks.

tests/common.go
```diff
@@ -1790,6 +1790,16 @@ func ValidateAndDestroy(contexts []*scheduler.Context, opts map[string]bool) {
 	})
 }
 
+// DestroyApps destroy applications without validating thme
```
vpinisetti-px, line 1793: spell correction `them'

tests/backup/backup_resiliency_test.go
lsrinivas-pure, line 1315: Why do you want to skip validation, skipping validation might lead to missing the bug.
Can you check why postgres is not up?

vpinisetti-px (approved):
LGTM

ak-px (approved):
LGTM

tests/backup/backup_resiliency_test.go
```diff
@@ -1312,10 +1312,12 @@ var _ = Describe("{RebootNodesWhenBackupsAreInProgress}", func() {
...
 		log.Infof("Switching cluster context back to source cluster")
-		err = SetSourceKubeConfig()
-		log.FailOnError(err, "Switching context to source cluster")
+		defer func() {
+			err = SetSourceKubeConfig()
+			log.FailOnError(err, "Switching context to source cluster")
+		}()
```
snigdha-px, line 1320: Correct me if my understanding is wrong
Here if we use defer , it will not run that block of SetSourceKubeConfig , go to 
ctx, err := backup.GetAdminCtxFromSecret() and fail as context was of destination cluster
Adding err = SetSourceKubeConfig() in defer is correct but we need to add err = SetSourceKubeConfig() again before line 1321 i.e getting the src cluster context

tests/common.go
kshithijiyer-px: Thanks done

tests/backup/backup_resiliency_test.go
kshithijiyer-px: @lsrinivas-pure What's the point of validating if the app is running in `JustAfterEach()`? The problem with the current implementation is that once it fails in `ValidateAndDestroy ` it doesn't reset the context causing the subsequent  tests to fails. 

The postgres pods are failing as the they are unable to find the pgbench database. we aren't sure if this issue is something coming from the portworx side of things. 

Describe of the pod:
```
Events:
  Type     Reason     Age                     From     Message
  ----     ------     ----                    ----     -------
  Warning  Unhealthy  15m (x798 over 17h)     kubelet  Liveness probe failed: psql: FATAL:  database "pgbench" does not exist
  Warning  BackOff    5m34s (x3450 over 17h)  kubelet  Back-off restarting failed container
  Warning  Unhealthy  27s (x1922 over 17h)    kubelet  Readiness probe failed: psql: FATAL:  database "pgbench" does not exist
```

logs of the pods: 
```
[root@ip-10-13-201-90 ~]# kubectl logs -n postgres-backup-pxbackuptask-0-05-07-12h53m58s  postgres-646f6f7487-8vfcm

PostgreSQL Database directory appears to contain a database; Skipping initialization

LOG:  could not create IPv6 socket: Address family not supported by protocol
LOG:  database system was shut down at 2023-05-08 04:39:16 UTC
LOG:  MultiXact member wraparound protections are now enabled
LOG:  database system is ready to accept connections
LOG:  autovacuum launcher started
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
FATAL:  database "pgbench" does not exist
LOG:  received fast shutdown request
LOG:  aborting any active transactions
LOG:  autovacuum launcher shutting down
LOG:  shutting down
LOG:  database system is shut down
```

tests/backup/backup_resiliency_test.go
kshithijiyer-px: @mkoppal-px Please confirm

tests/backup/backup_resiliency_test.go
vpinisetti-px: I believe, it's better to split and call this in two places 
1. ValidateApplications() - > Should call right after the node reboot . So that will get to know if there is any issue after node reboot.
2. DestroyApps() -> Lets keep this in teardown part

kphalgun-px: This might be out of scope for this PR. To ensure that errors are properly handled, it would be appropriate to implement an errChan parameter in the ValidateAndDestroy function. This will enable the channel to be updated whenever errors occur.

tests/backup/backup_resiliency_test.go
mkoppal-px: @kshithijiyer-px we need to keep the existing context switch to source cluster in JustAfterEach l. Just make sure you add this defer block to the first line of JustAfterEach

kshithijiyer-px: Ack doing it.

tests/backup/backup_resiliency_test.go
lsrinivas-pure: @kshithijiyer-px 
Ideally validating the apps should be in your test itself.
I would suggest you move this validation to inside the test fail the test if any validation fails and since context switch is happening in aftereach it should work

mkoppal-px: @lsrinivas-pure We are validating the deployed app in our test case itself. We have this step
```
Step("Validating the deployed applications", func() {
			log.InfoD("Validating the deployed applications on destination cluster")
		})
```
The test case actually failed in this step itself. So now we are asking if it is necessary to validate again in the `JustAfterEach` block again because it will never pass the validation here.

mkoppal-px: So we have `ValidateApplications` inside the test, so I think it is safe to use `DestroyApps` in the cleanup part, right?

snigdha-px (approved):
LGTM

mkoppal-px (approved):
LGTM

tests/backup/backup_resiliency_test.go
kshithijiyer-px: @lsrinivas-pure Waiting for your final approval.

lsrinivas-pure: as per discussion with @kshithijiyer-px this change of moving validations from AfterEach to test will be taken care by @snigdha-px in a separate PR

@snigdha-px please mention that task link in this PR

lsrinivas-pure (approved):
lgtm

kphalgun-px (approved):
LGTM

tests/backup/backup_resiliency_test.go
vpinisetti-px: Yes @lsrinivas-pure we have created a bug to handle this state where the application is up & running or not after reboot the nodes. With this approach, app state validation goes to testcase and app deletion will be in teardown.
https://portworx.atlassian.net/jira/software/c/projects/PA/issues/PA-1018

PR #1381, opened 2023-05-10
PTX-17570 Add an optional method to upgrade PX Operator during PX upgrade hops based on the PX endpoint

nikolaypopov:
Add an optional method to upgrade PX Operator during PX upgrade hops based on the PX endpoint


drivers/volume/portworx/portworx.go
```diff
@@ -2952,6 +2939,107 @@ func (d *portworx) upgradePortworxDaemonset(specGenUrl string) error {
...
+	// Get k8s version
+	kubeVersion, err := d.schedOps.GetKubernetesVersion()
+	if err != nil {
+		return err
+	}
+
+	// Getting PX Operator spec
+	u, err := url.Parse(specGenUrl)
+	if err != nil {
+		fmt.Printf("failed to parse URL [%s], Err: %v", specGenUrl, err)
```
aghodke1312, line 3002: log?

drivers/volume/portworx/portworx.go
```diff
@@ -2952,6 +2939,107 @@ func (d *portworx) upgradePortworxDaemonset(specGenUrl string) error {
...
+		var envVars []corev1.EnvVar
+		// Add Release Manifest URL incase of edge spec
+		envVars, err := configurePxReleaseManifestEnvVars(cluster.Spec.Env, specGenUrl)
+		if err != nil {
+			return nil
+		}
+		cluster.Spec.Env = envVars
+		return cluster
+	}
+
```
aghodke1312, line 2977: Can you please call stats API here. We are doing something like that in upgrade test. Send stats for upgrade-operator. 
`dash.UpdateStats("px-upgrade-stats", "px-enterprise", "upgrade", majorVersion, statsData)`

drivers/volume/portworx/portworx.go
nikolaypopov: Good catch, fix it in another place I copied this func from

drivers/volume/portworx/portworx.go
nikolaypopov: I don't want to add it inside this function, something like that would need to be added in the test itself

nikolaypopov: other upgrade tests will probably need that as well, let's do it as a part of a separate PR

aghodke1312 (approved):

aghodke1312 (approved):

PR #1396, opened 2023-05-17
PWX-30979: Reduces pool resize timeout from 6 hours to 1

Pure-AdamuKaapan:
Pool resizes should not take 6 hours to complete: if they do, something is very wrong. Reduced this to a still long, but more reasonable 1 hour.

### DIFF
--- a/tests/basic/storage_pool_test.go
+++ b/tests/basic/storage_pool_test.go
@@ -31,7 +31,7 @@ import (
 
 const (
 	replicationUpdateTimeout = 4 * time.Hour
-	poolResizeTimeout        = time.Minute * 360
+	poolResizeTimeout        = time.Minute * 60
 	retryTimeout             = time.Minute * 2
 	addDriveUpTimeOut        = time.Minute * 15
 )


tests/basic/storage_pool_test.go
```diff
@@ -31,7 +31,7 @@ import (
 
 const (
 	replicationUpdateTimeout = 4 * time.Hour
-	poolResizeTimeout        = time.Minute * 360
+	poolResizeTimeout        = time.Minute * 60
```
jainsmit, line 34: Actually In longevity setup where pool size become large there it could take more time !!

Pure-AdamuKaapan: Even for resize disk, not add disk? Add disk has a separate timeout.

Is there a way we could change the timeout based on if it's longevity or not? The goal is to keep tests from stalling for a really long time before failing

lsrinivas-pure: @Pure-AdamuKaapan can you raise a PTX regarding this for me.. Ill make changes to different timeout for drive add, add-disk and resize-disk. need to go through the tests and see they have appropriate time out

Pure-AdamuKaapan: Sounds good, thanks so much! I'll close this PR for now then.

Pure-AdamuKaapan: https://portworx.atlassian.net/browse/PTX-17782

Pure-AdamuKaapan:
Closing this, assigned https://portworx.atlassian.net/browse/PTX-17782 to @lsrinivas-pure to tackle more.

PR #1397, opened 2023-05-17
[PA-567] Adding deeper validation in Restore Operation

tthurlapati-px:
## What this PR does

1. It adds the ability to Validate the finer details of a restore (to ensure that it has been done correctly). This includes whether all specs that are scheduled (including volumes and other properties) have been restored (as per what px-backup api tells us).
2. Updated `BasicBackupCreation` to use a "standardized" method of writing code that obeys invariants that are built into my implementation of deeper restore validation such that there is lesser chance of writing code that can fail.
3. `GetUpdatedSpec` will now help get the latest spec, which I also use in a _twisted_ way to check if the object exists on the destination cluster (after restore), by "getting it" and I know it doesn't exist if I fail to get it.
4. Added the ability to clean up destination cluster after a restore has been done, so that it can be done if desired.
5. `TransformAppContextWithMappings` can now be used to do it what it says it does.
6. ADD missing `kind` field for `ConfigMap` and `StatefulSets`.
7. ADD `CreateRestoreWithValidation`
8. ADD `ClusterConfigPathMap` to translate between cluster names (created in px-backup) and the corresponding k8s cluster kubeconfig path.

### Tests using `ValidateRestore`

- [x] BasicBackupCreation

## Which issue(s) this PR fixes
Closes #PA-567

## Special notes for your reviewer

postgres-backup (different cluster for source and destination) - BasicBackupCreation 
- https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/1388/console
- https://aetos.pwx.purestorage.com/resultSet/testSetID/195830

postgres-backup (same cluster for source and destination) - BasicBackupCreation 
- https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/1530/
- https://aetos.pwx.purestorage.com/resultSet/testSetID/211471

mysql-backup (different cluster for source and destination)  - BasicBackupCreation 
- https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/1390/console
- https://aetos.pwx.purestorage.com/resultSet/testSetID/195880

mysql-backup (same cluster for source and destination) - BasicBackupCreation 
- https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/1595/
- https://aetos.pwx.purestorage.com/resultSet/testSetID/215216

## HACK

The `kind` field missing on the objects returned by certain sched-ops functions (that are used in `k8s.go` seems to be an issue with kubernetes itself:
https://github.com/kubernetes/kubernetes/issues/80609
https://github.com/kubernetes/kubernetes/issues/3030

Currently, this issue seems to be affecting those calls in sched-ops which get a list and then filter out the required object, like the following example where Services and fetched and service is filtered: `c.kubernetes.CoreV1().Services(ns).Create(context.TODO(), service, metav1.CreateOptions{})`
That's the reason some of them don't have kind (list being fetched causes apiVersion and kind to be dropped due to k8s bug where TypeMeta is not found, I think), whereas some do have kind.

Basically, in `k8s.go` this is what is happening:
```go
else if obj, ok := spec.(*corev1.Service); ok {
		obj.Namespace = ns.Name
// obj.Kind is PRESENT as per logs
		svc, err := k8sCore.CreateService(obj) //this could be GetService(...)
		if err != nil {
			return nil, &scheduler.ErrFailedToScheduleApp{
				App:   app,
				Cause: fmt.Sprintf("Failed to create Service: %v. Err: %v", obj.Name, err),
			}
		}
// svc.Kind is EMPTY as per logs
		// This is a hack, because the `Kind` is empty for some reason
		svc.Kind = "Service"
// now svc.Kind is PRESENT as per logs
		log.Infof("[%v] Created Service: %v", app.Key, svc.Name)
		return svc, nil

	} 
```

So, as a solution, as seen above, I am manually assigning the `Kind` field. When the K8s bug gets fixed, we don't need to do that any longer.

refer PA-1113 for more info


kshithijiyer-px (approved):
LGTM

mkoppal-px:
Please run the test with `mysql-backup app` as well

drivers/scheduler/k8s/k8s.go
```diff
@@ -1367,11 +1367,49 @@ func (k *K8s) UpdateTasksID(ctx *scheduler.Context, id string) error {
 	return nil
 }
 
-// GetUpdatedSpec takes a spec and returns the latest version of it by `GET`-ing it
+// GetUpdatedSpec gets the updated spec of a K8s Object. Just `name`, and if required `namespace` must be spcified on `spec` in order to GET the spec from K8s (if it exists).
```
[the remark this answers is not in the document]
tthurlapati-px: ✅

tthurlapati-px:
I already tested mysql-backup. You mean on the same cluster?

tests/backup/backup_helper.go
```diff
@@ -1629,6 +1667,310 @@ func restoreSuccessWithReplacePolicy(restoreName string, orgID string, retryDura
...
+		restoreStatusReason := theRestore.GetStatus().Reason
+		return fmt.Errorf("ValidateRestore requires restore [%s] to have a status of Success or PartialSuccess, but found [%s] with reason [%s]", restoreName, restoreStatus, restoreStatusReason)
+	}
+
+	var errors []error
+
+	// check if all the objects in the spec are present in the restore as per what px-backup reports
+	for _, expectedRestoredAppContext := range expectedRestoredAppContexts {
+
+		expectedRestoredAppContextNamespace := expectedRestoredAppContext.ScheduleOptions.Namespace
```
kphalgun-px, line 1702: How about using GetAppNamespace?

tests/backup/backup_helper.go
```diff
@@ -1629,6 +1667,310 @@ func restoreSuccessWithReplacePolicy(restoreName string, orgID string, retryDura
...
+	if !allStorageClassMappingsPresent {
+		restoreAppContext.SkipVolumeValidation = true
+	}
+
+	return &restoreAppContext, nil
+}
+
+// CloneSpec clones a given spec and returns it. It returns an error if the object (spec) provided is not supported by this function
+func TransformToRestoredSpec(spec interface{}, storageClassMapping map[string]string) error {
+	if specObj, ok := spec.(*corev1.PersistentVolumeClaim); ok {
```
kphalgun-px, line 1964: How about returning an error if it is not ok?

tests/backup/backup_helper.go
```diff
@@ -1629,6 +1667,310 @@ func restoreSuccessWithReplacePolicy(restoreName string, orgID string, retryDura
...
+		if sc, ok := storageClassMapping[*specObj.Spec.StorageClassName]; ok {
```
kphalgun-px, line 1965: How about returning an error if it's not okay? Also, log specObj.Spec.StorageClassName, sc.

tests/backup/backup_test.go
```diff
@@ -232,7 +236,7 @@ var _ = Describe("{BasicBackupCreation}", func() {
 				scheduledNamespace := appCtx.ScheduleOptions.Namespace
 				backupName := fmt.Sprintf("%s-%v", "backup", time.Now().Unix())
 				log.InfoD("creating backup [%s] in source cluster [%s] (%s), organization [%s], of namespace [%s], in backup location [%s]", backupName, SourceClusterName, sourceClusterUid, orgID, scheduledNamespace, backupLocationName)
-				err := CreateBackupWithValidation(ctx, backupName, SourceClusterName, backupLocationName, backupLocationUID, []*scheduler.Context{scheduledAppContexts[i]}, labelSelectors, orgID, sourceClusterUid, "", "", "", "")
+				err := CreateBackupWithValidation(ctx, backupName, SourceClusterName, backupLocationName, backupLocationUID, scheduledAppContexts[i:i+1], labelSelectors, orgID, sourceClusterUid, "", "", "", "")
```
kphalgun-px, line 239: How about scheduledAppContexts[i]? Since scheduledAppContexts is a slice of pointers, does scheduledAppContexts[i:i+1] only point to scheduledAppContexts[i]?

tests/backup/backup_helper.go
```diff
@@ -1629,6 +1667,310 @@ func restoreSuccessWithReplacePolicy(restoreName string, orgID string, retryDura
...
+
+	if len(errStrings) > 0 {
+		return nil, fmt.Errorf("TransformAppContextWithMappings Errors: {%s}", strings.Join(errStrings, "}\n{"))
+	}
+
+	app := *appContext.App
+	app.SpecList = specObjects
+	restoreAppContext.App = &app
+
+	// we're having to do this as we're under the assumption that `ScheduleOptions.Namespace` will always contain the namespace of the scheduled app
```
kphalgun-px, line 1945: grammar: we have* to

tests/backup/backup_helper.go
```diff
@@ -1629,6 +1667,310 @@ func restoreSuccessWithReplacePolicy(restoreName string, orgID string, retryDura
...
+
+	if len(errStrings) > 0 {
+		return fmt.Errorf("ValidateRestore Errors: {%s}", strings.Join(errStrings, "}\n{"))
+	} else {
+		return nil
+	}
+}
+
+// TransformAppContextWithMappings clones an appContext and transforms it according to the maps provided. When `forRestore` is true, VolumeSnapshots and StorageClasses are ignored. To be used after switching k8s context (cluster) which has the namespace
+func TransformAppContextWithMappings(appContext *scheduler.Context, namespaceMapping map[string]string, storageClassMapping map[string]string, forRestore bool) (*scheduler.Context, error) {
```
kphalgun-px, line 1879: How about using DeepCopy of scheduler.Context?

tests/backup/backup_helper.go
```diff
@@ -1629,6 +1667,310 @@ func restoreSuccessWithReplacePolicy(restoreName string, orgID string, retryDura
...
+		if err != nil {
+			errStrings = append(errStrings, err.Error())
+		}
+	}
+
+	if len(errStrings) > 0 {
+		return fmt.Errorf("ValidateRestore Errors: {%s}", strings.Join(errStrings, "}\n{"))
+	} else {
+		return nil
+	}
```
kphalgun-px, line 1875: How about we return `[]error` and convert `[]error` to a string for logging outside? @mkoppal-px

tests/backup/backup_helper.go
tthurlapati-px: it's not an error if it's not okay. This is just transforming stuff _if_ it exists.

tests/backup/backup_helper.go
tthurlapati-px: it's not an error if it's not okay. This is just transforming stuff _if_ it exists.
I feel that there's no point in logging this as this is fed in from the outside. Whatever logging needs to be done, needs to be done by the person who created the mapping as they own the data, not I. Though, logging _if_ there is PVC in itself is useful as that also tells us that SCMapping took place, though that requires unnecessary logic.

tests/backup/backup_test.go
tthurlapati-px: No, we can't do that.
In go lang (as in most other langs):
- `scheduledAppContexts[i]` is a **single element / OBJECT** at location `i`, and it is NOT a slice
- `[]*scheduler.Context{scheduledAppContexts[i]}` is a slice that we create with a single **element**, the element being `scheduledAppContexts[i]`
- `scheduledAppContexts[i:i+1]` is a SLICE that we create _with_ a **single** element, which is the element at location `i`

What we need is either the 2nd or 3rd option, since the **function signature requires a slice**. I think it's simpler to specify an existing slice of array, rather than choose an element and creating a new slice before passing it to the function.

tests/backup/backup_helper.go
tthurlapati-px: It's a grammatically correct sentence, but I changed it to make more sense.

tests/backup/backup_helper.go
tthurlapati-px: I don't think that function works properly. It's missing some fields.

tests/backup/backup_helper.go
tthurlapati-px: That is what I did originally. But Mithun suggested that we just return just one error. I feel like that makes sense in this case as _any_ error _is_ an error in this function and should be treated as such.

tests/backup/backup_helper.go
kphalgun-px: The function signature implies that it should throw an error, but the function does not actually throw any errors.

tests/backup/backup_helper.go
kphalgun-px: The function signature implies that it should throw an error, but the function does not actually throw any errors.

tests/backup/backup_helper.go
kphalgun-px: Yes, but modifying the context will alter the original value since it is a pointer to the context.

tests/backup/backup_helper.go
kphalgun-px: Yeah, I think it is better if we can create a new function called log.FailOnMultipleErrors or something similar. This function would take an array slice and print each error along with its corresponding index.

tests/backup/backup_test.go
kphalgun-px: My bad. Thanks.

tests/backup/backup_helper.go
tthurlapati-px: not possible right now

tests/backup/backup_helper.go
tthurlapati-px: I changed it

tests/backup/backup_helper.go
tthurlapati-px: changed it. see latest commit

tests/backup/backup_helper.go
tthurlapati-px: no, changed function name to be more clear. It clones and transforms

tests/backup/backup_helper.go
tthurlapati-px: While this might be useful, returning a single error has the advantage of calling this in other functions that return a single error. So, at least for now, let's leave it. 

Though, technically, given that this is a "part of testing", we should directly do `log.Errorf` wherever there's an error (and continue) and then we can set flags that we can use to determine if we want to fail due to error in validation (as validation is not an obstruction in testing for other errors in the same testcase, before ending the whole thing).

for now, let it be thus

tthurlapati-px:
@mkoppal-px  see PR description for all test runs. I have done all requested.

mkoppal-px (approved):
LGTM

PR #1404, opened 2023-05-19
PA-914 - Validate deployments health in pds-system namespace instead …

aravindranath14:
This PR fixes PA-914, Instead of verifying individual pods in the pds-system namespace, we now verify the deployment health. If there are some deployments that have 0 pods deployed, then this corner case is caught with this fix
Here is a successful run: 
https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-system-test/view/PDS%20System%20Tests/job/pds-system-test-anthos/140/console


gauraa2 (approved):
LGTM

dbhatnagar-px (approved):
lgtm

PR #1406, opened 2023-05-19
Replacing ValidateBackupCluster function

mkoppal-px:
The function `ValidateBackupCluster` was written during the early days of backup automation when we dint not have a separate `backup_helper.go`. During the creation of upgrade test case we wrote a function `ValidateAllPodsInPxBackupNamespace` which does the exact same thing, so we are finally retiring `ValidateBackupCluster`

Closes #PA-777

[Jenkins Run](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Sumit/job/Custom-Pipelines/job/px-backup-on-demand-system-test-byoc/1413/)

### DIFF
--- a/drivers/backup/backup.go
+++ b/drivers/backup/backup.go
@@ -132,9 +132,6 @@ type Cluster interface {
 		timeBeforeRetry time.Duration,
 	) error
 
-	// ValidateBackupCluster validates if backup pods are up or not
-	ValidateBackupCluster() error
-
 	// GetClusterUID returns uid of the given cluster name in an organization
 	GetClusterUID(ctx context.Context, orgID string, clusterName string) (string, error)
 
--- a/drivers/backup/portworx/portworx.go
+++ b/drivers/backup/portworx/portworx.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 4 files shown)


tthurlapati-px (approved):
LGTM

kphalgun-px:
ValidateBackupCluster has a more comprehensive validation approach. Can we retire ValidateAllPodsInPxBackupNamespace instead?

lsrinivas-pure (approved):
lgtm

mkoppal-px:
> ValidateBackupCluster has a more comprehensive validation approach. Can we retire ValidateAllPodsInPxBackupNamespace instead?

What do you think it is doing extra in order to make sure that Px-Backup is healthy?

kphalgun-px:
> > ValidateBackupCluster has a more comprehensive validation approach. Can we retire ValidateAllPodsInPxBackupNamespace instead?
> 
> What do you think it is doing extra in order to make sure that Px-Backup is healthy?

I am not entirely sure, but the excluded pods might have a different lifecycle compared to other pods in the backup cluster, or the default timeout might be exceeded when validating those pods.

mkoppal-px:
> > > ValidateBackupCluster has a more comprehensive validation approach. Can we retire ValidateAllPodsInPxBackupNamespace instead?
> > 
> > 
> > What do you think it is doing extra in order to make sure that Px-Backup is healthy?
> 
> I am not entirely sure, but the excluded pods might have a different lifecycle compared to other pods in the backup cluster, or the default timeout might be exceeded when validating those pods.

The excluded pods are not indicative of whether the Px-Backup app is healthy or not. We can discuss more on this offline.

PR #1407, opened 2023-05-19
PTX-17751 Update vendor for operator to resolve Telemetry race condition

nikolaypopov:
Update vendor for operator to resolve Telemetry race condition


mborodin-px (approved):
lgtm

PR #1409, opened 2023-05-20
PTX-17975 Update vendor to not validate telemetry secret ref in op below 1.10

nikolaypopov:
Update operator vendor package to not validate telemetry secret ref in op below 1.10


mborodin-px (approved):
lgtm

PR #1420, opened 2023-05-26
PTX-17615:add drive testcase should be forced to make new pool when the poolid is -1

soma-purestorage:
make testcases
PoolAddDrive
AddDriveAndPXRestart
AddDriveWithPXRestart
use new pool when using default poolid as -1

Closes #

https://jenkins.pwx.dev.purestorage.com/job/Users/job/Somashekar/job/Torpedo/job/eks-soma/24/console


tests/basic/storage_pool_test.go
```diff
@@ -8828,3 +8828,141 @@ var _ = Describe("{ReplResyncOnPoolExpand}", func() {
...
+		stepLog = fmt.Sprintf("expand pool %s using add-disk", selectedPool.Uuid)
+		var expandedExpectedPoolSize uint64
+		Step(stepLog, func() {
+			log.InfoD(stepLog)
+			expandedExpectedPoolSize = (selectedPool.TotalSize / units.GiB) * 2
+
+			log.FailOnError(err, "Failed to check if Journal enabled")
+
+			log.InfoD("Initiate add cloud drive and validate")
+			err := addCloudDrive(*selectedNode, selectedPool.ID)
```
lsrinivas-pure, line 8926: make the changes to create a new pool rather than adding disk to existing pool.
since legacy add drive to existing pool is not the recommended way

lsrinivas-pure (changes requested):
Add more details on the PR like what it does and PTX

drivers/volume/portworx/portworx.go
```diff
@@ -5016,9 +5016,11 @@ func addDrive(n node.Node, drivePath string, poolID int32, d *portworx) error {
 		driveAddFlag = fmt.Sprintf("-s %s", drivePath)
 		if poolID != -1 {
 			driveAddFlag = fmt.Sprintf("%s -p %d", driveAddFlag, poolID)
+		} else if poolID == -1 {
```
lsrinivas-pure, line 5019: just else is fine

drivers/volume/portworx/portworx.go
```diff
@@ -5016,9 +5016,11 @@ func addDrive(n node.Node, drivePath string, poolID int32, d *portworx) error {
...
+			driveAddFlag = fmt.Sprintf("%s %s", driveAddFlag, "--newpool")
 		}
 	}
-
+	fmt.Printf("drivepathcommand %s", driveAddFlag)
```
lsrinivas-pure, line 5023: remove this print

lsrinivas-pure (changes requested):

drivers/volume/portworx/portworx.go
soma-purestorage: reoslved

drivers/volume/portworx/portworx.go
soma-purestorage: else means even if we uuid of poolid it will go to new pool so specifically used else if

lsrinivas-pure (approved):
LGTM

PR #1424, opened 2023-05-28
Fixing decomm node test

lsrinivas-pure:

Closes #PTX-17686


vprabhakar-px (approved):

drivers/volume/portworx/portworx.go
```diff
@@ -3238,17 +3238,41 @@ func (d *portworx) DecommissionNode(n *node.Node) error {
...
+			break
+		}
+	}
+
+	t = func() (interface{}, bool, error) {
+		out, err := d.nodeDriver.RunCommandWithNoRetry(cmdNode, cmd, node.ConnectionOpts{
+			Timeout:         2 * time.Minute,
+			TimeBeforeRetry: 10 * time.Second,
+		})
+		if err != nil && strings.Contains(err.Error(), "Node remove is pending") {
```
mkoppal-px, line 3257: Are you sure about the upper/lower case here?

lsrinivas-pure: yes..that is the exact string we get when delete command is executed

drivers/volume/portworx/portworx.go
```diff
@@ -3238,17 +3238,41 @@ func (d *portworx) DecommissionNode(n *node.Node) error {
...
+
+			return out, false, nil
+
 		}
+		if err != nil {
+			return "", true, err
```
mkoppal-px, line 3263: Why do you want to retry in case of error? Here the error would mean the command did not exit with status 0 right?

mkoppal-px (approved):
LGTM

PR #1443, opened 2023-06-02
added new multi reboot test

dbhatnagar-px:
This test does the following: 
1. Starts creating a data service
2. While DS is coming up, it will reboot one of the nodes three times in succession (on which DS is coming up)
3. Once done, we will validate DS still comes up

Jenkins Run: https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/298/

### DIFF
--- a/drivers/pds/lib/resiliency_utils.go
+++ b/drivers/pds/lib/resiliency_utils.go
@@ -320,7 +320,7 @@ func NodeRebootDurinAppVersionUpdate(ns string, deployment *pds.ModelsDeployment
 }
 
 // Reboot the Active Node onto which the application pod is coming up
-func RebootActiveNodeDuringDeployment(ns string, deployment *pds.ModelsDeployment) error {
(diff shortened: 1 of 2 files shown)


drivers/pds/lib/resiliency_utils.go
```diff
@@ -382,6 +382,30 @@ func RebootActiveNodeDuringDeployment(ns string, deployment *pds.ModelsDeploymen
 				CapturedErrors <- testError
 				return testError
 			}
+			if num_reboots > 1 {
+				for i := 1; i <= num_reboots; i++ {
```
Madanagopal19, line 386: nit: rename i as index

Madanagopal19 (approved):
minor comment, rest looks good to me.

drivers/pds/lib/resiliency_utils.go
dbhatnagar-px: done

aravindranath14 (approved):
LGTM

PR #1464, opened 2023-06-12
Adding Backuplocation status to logs while cleaning up

mkoppal-px:
Adding Backup location status to logs while cleaning up
Currently, we are not printing the status of backup location after we send the delete request. Found in NFS tests that the backup location deletion is not happening

Closes #


kshithijiyer-px (approved):
LGTM

PR #1473, opened 2023-06-17
Fixing test `DeleteBackupAndCheckIfBucketIsEmpty` to work with both S3 and NFS backup location

mkoppal-px:
Here are the fixes in this PR:

- [x] Made sure we are creating custom buckets for the test `DeleteBackupAndCheckIfBucketIsEmpty` to avoid residue files left from other tests while validating the bucket
- [x] `DeleteBackupAndCheckIfBucketIsEmpty` now works with both NFS and S3 backup locations
- [x] Added a common function `IsBucketEmpty` which checks the bucket/sub path depending on whether the backup location type is NFS or S3
- [x] Created a helper function `DeleteS3Objects` which can be used in future for any tests
- [x] Increased the backup deletion timeout from 20 min to 60 min since NFS based backups takes a lot of time to get deleted especially if the backup dependency chain is long
- [x] Handled the error during node initialisation

Closes #PA-1241

[Jenkins Run with NFS backup location](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/1698/console)
[Jenkins Run with S3 backup location](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/1699/)


kphalgun-px (approved):
LGTM

tests/common.go
```diff
@@ -4595,6 +4606,70 @@ func CreateBucket(provider string, bucketName string) {
 	})
 }
 
+// IsBucketEmpty returns true if the bucket for a provider is empty
+func IsBucketEmpty(provider, bucketName string) (bool, error) {
```
trenukarya-px, line 4610: This function is confusing since we are using it for both Aws and Nfs. Can we define generic name ?

tests/common.go
```diff
@@ -4595,6 +4606,70 @@ func CreateBucket(provider string, bucketName string) {
...
+		}
+		for _, cmd := range umountCmds {
+			err := runCmd(cmd, masterNode)
+			log.FailOnError(err, fmt.Sprintf("Failed to run [%s] command on node [%s], error : [%s]", cmd, masterNode, err))
+		}
+	}()
+
+	// List the files in subpath from NFS share path.
+	log.Infof("Checking the contents in NFS share subpath: [%s] from path: [%s] on server: [%s]", subPath, creds.NfsPath, creds.NfsServerAddress)
+	fileCountCmd := fmt.Sprintf("find %s/%s -type f | wc -l", mountDir, subPath)
```
trenukarya-px, line 4661: Are we checking file count here or just directory exists not ? Otherwise, find command succeeds if directory exists and no files exists in that directory.

pkg/s3utils/s3utils.go
```diff
@@ -128,3 +128,57 @@ func GetS3Objects(clusterID string, nodeName string, getPreviousFolder bool) ([]
...
+		Bucket: aws.String(bucket),
+		Delete: &s3.Delete{
+			Objects: objectsToDelete,
+			Quiet:   aws.Bool(false),
+		},
+	}
+
+	// Delete the objects from the bucket
+	_, err = S3Client.DeleteObjects(deleteInput)
+	if err != nil {
```
lsrinivas-pure, line 180: nit: if check not required, just return err,
err will be nil if no error returned by the function call

tests/common.go
```diff
@@ -4595,6 +4606,70 @@ func CreateBucket(provider string, bucketName string) {
 	})
 }
 
+// IsBucketEmpty returns true if the bucket for a provider is empty
+func IsBucketEmpty(provider, bucketName string) (bool, error) {
+	switch provider {
+	case drivers.ProviderAws:
+		result, err := IsS3BucketEmpty(bucketName)
+		if err != nil {
```
lsrinivas-pure, line 4614: no need of if check here as you are returning result and err anyway

tests/common.go
```diff
@@ -4595,6 +4606,70 @@ func CreateBucket(provider string, bucketName string) {
...
+			return result, err
+		}
+		return result, err
+	case drivers.ProviderNfs:
+		result, err := IsNFSSubPathEmpty(bucketName)
+		if err != nil {
```
lsrinivas-pure, line 4620: same as above

tests/common.go
mkoppal-px: We are just checking the file count which is why I have added `-type f`. Directories can exist.

tests/common.go
mkoppal-px: Changed it to `IsBackupLocationEmpty`

pkg/s3utils/s3utils.go
mkoppal-px: Done

tests/common.go
mkoppal-px: Done

lsrinivas-pure (approved):
LGTM

tests/common.go
trenukarya-px: @mkoppal-px Please see this:

trenukarya--Mac14:repos trenukarya$ find ~/work/repos/torpedo/ -type f | wc -l
   11198
trenukarya--Mac14:repos trenukarya$ find ~/work/temp/ -type f | wc -l
       0
trenukarya--Mac14:repos trenukarya$ echo $?
0

mkoppal-px: Did not get you here? We are checking for the total number of files present in the subpath

mkoppal-px: Clarified this issue in-person

PR #1479, opened 2023-06-21
Fix failed to retrieve organization org-name: object not found issue

sgajawada-px:
There is an issue in passing the parameters for 
func (*portworx).GetBackupUID(ctx context.Context, backupName string, orgID string) (string, error)

So the parameters are passed correctly for the caller methods. 


siva-portworx (approved):

mkoppal-px (approved):
LGTM

PR #1516, opened 2023-07-07
PA-1325 ScaleDownScaleUpPxPods need to scaledown only 2 nodes max

aravindranath14:
It was noticed that on smaller clusters, ie., e2e clusters with 5 worker nodes and 3 data services, the testcase was stopping Px on 3 nodes and hence the Px was losing quorum and Data service pods  did not have a replica node to re-schedule on.
With this fix, The total number of nodes on which Px is stopped is capped at 2

Here is a branch build log:
https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/452/console


dbhatnagar-px (approved):
Lgtm

tests/pds/dataservice_test.go
```diff
@@ -567,7 +567,9 @@ var _ = Describe("{ScaleDownScaleupPXCluster}", func() {
 					nodes, err := pdslib.GetNodesOfSS(*deployment.ClusterResourceName, namespace)
 					log.FailOnError(err, "Error while getting Data Serice Nodes")
 					deploymentNode := nodes[0]
-					nodeList = append(nodeList, deploymentNode)
+					if len(nodeList) < 2 {
+						nodeList = append(nodeList, deploymentNode)
+					}
```
mkoppal-px, line 572: How will you handle the case when the `nodeList` is greater than 2?

aravindranath14: We are creating the nodeList here. So we are selecting how many nodes to shutdown Px on. For eg, We have 5 worker nodes, Then if we shutdown 3 px nodes, we do not have enough nodes with volume replicas and the pods go into pending state.
with this we are only controlling how many nodes we shut down Px on

mkoppal-px (approved):
LGTM

PR #1523, opened 2023-07-12
PB-4015: To support pure file volume driver

sgajawada-px:
- update pso volume driver to register pure file. if the default pure driver is used for pure block.

add support for pure-file volume driver

Closes #PB-4015

**TestResults**:
Screenshot from 2023-07-12 14-09-31
Screenshot from 2023-07-12 10-20-46
Screenshot from 2023-07-11 23-04-03
Screenshot from 2023-07-11 23-03-50
Screenshot from 2023-07-11 22-20-18
Screenshot from 2023-07-11 22-20-06
Screenshot from 2023-07-11 22-16-49

### DIFF
--- a/drivers/volume/pso/pso.go
+++ b/drivers/volume/pso/pso.go
@@ -2,24 +2,27 @@ package pso
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/portworx/sched-ops/k8s/core"
 	torpedovolume "github.com/portworx/torpedo/drivers/volume"
 	"github.com/portworx/torpedo/drivers/volume/portworx"
 	"github.com/portworx/torpedo/drivers/volume/portworx/schedops"
 	"github.com/portworx/torpedo/pkg/log"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"strings"
 )
 
 const (
 	// PureDriverName is the name of the portworx-pure driver implementation
-	PureDriverName = "pso"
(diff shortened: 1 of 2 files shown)


drivers/node/ssh/ssh.go
```diff
@@ -160,18 +160,17 @@ func (s *SSH) GetDeviceMapperCount(n node.Node, timerange time.Duration) (int, e
...
 	if s.IsUsingSSH() {
 		err = s.initSSH()
 	} else {
+		// The namespace of the portworx-service in the current context
+		if execPodNamespace, err := getExecPodNamespace(); err != nil {
+			return err
+		} else {
+			s.execPodNamespace = execPodNamespace
+			log.InfoD("UpdateDriver: ssh driver's namespace (the 'portworx-service' namespace) is updated to [%s]", execPodNamespace)
+		}
```
mkoppal-px, line 173: Why did we do this? We would need `s.execPodNamespace` to be set in case `s.IsUsingSSH()` is also `true`. This may affect the Px tests

sgajawada-px: Ok, then we need to fix this.
But why by default ssh driver is checking the portworx-service.
What if the soure cluster doesn't have portworx like we are testing with pso

sgajawada-px: Changed as discussed

mkoppal-px (approved):
LGTM

PR #1533, opened 2023-07-15
Added an addtional parameter in DeleteCluster

mkoppal-px:

I introduced a regression when I added `DeleteBackups` and `DeleteRestores` in `DeleteCluster` function.

**Why was this added in the first place?**
To make sure that all the backups and restores are cleaned up before deleting the cluster. This was working fine with Px based backups. But when we tested on PSO with KDMP backups, we saw that the backups were not getting cleaned up during backup location deletion, so we added another check in `DeleteCluster` function to be fully sure that the backups are deleted before deleting the cluster. This was necessary because, once we delete the cluster, deleting the backups was getting stuck because KDMP/CSI backups need cluster reference to delete the backups.

**Why and where did it regress?**
In the test `DeleteNSDeleteClusterRestore` we are actually trying to just remove the cluster and add it again without cleaning up the backups. Here we saw that the backups are also getting cleaned up when we removed the cluster and the restores were failing in the subsequent steps.

**What is the fix?**
Added another parameter `cleanupBackupsRestores` to `DeleteCluster` function. Now the caller can decide whether we need to cleanup all backups and restores before deleting the cluster or retain the backups and restores and just delete the cluster

Closes # To be created

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -3439,14 +3439,14 @@ func DeleteBackup(backupName string, backupUID string, orgID string, ctx context
 }
 
 // DeleteCluster deletes/de-registers cluster from px-backup
-func DeleteCluster(name string, orgID string, ctx context1.Context) error {
+func DeleteCluster(name string, orgID string, ctx context1.Context, cleanupBackupsRestores bool) error {
 
 	backupDriver := Inst().Backup
 	clusterDeleteReq := &api.ClusterDeleteRequest{
 		OrgId:          orgID,
 		Name:           name,
-		DeleteBackups:  true,
-		DeleteRestores: true,
+		DeleteBackups:  cleanupBackupsRestores,
+		DeleteRestores: cleanupBackupsRestores,
 	}
 	_, err := backupDriver.DeleteCluster(ctx, clusterDeleteReq)
 	return err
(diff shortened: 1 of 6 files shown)


tests/common.go
```diff
@@ -3439,14 +3439,14 @@ func DeleteBackup(backupName string, backupUID string, orgID string, ctx context
 }
 
 // DeleteCluster deletes/de-registers cluster from px-backup
-func DeleteCluster(name string, orgID string, ctx context1.Context) error {
+func DeleteCluster(name string, orgID string, ctx context1.Context, cleanupBackupsRestores bool) error {
```
kphalgun-px, line 3442: Can we have two Boolean parameters for DeleteBackups and DeleteRestores?

kphalgun-px (approved):
LGTM

tests/common.go
mkoppal-px: I thought about that. But this is mainly for cleanup. We will either cleanup everything or not cleanup at all.
This cleanup can only be done from the API, from API we can never cleanup the backups and restores while deleting the cluster.

mkoppal-px:
Tests are passing - https://aetos.pwx.purestorage.com/resultSet/testSetID/266751

lsrinivas-pure (approved):
LGTM

PR #1539, opened 2023-07-17
Adding annotations to SC in case of CSI apps 

aghodke1312:
Triggered a run with the change:
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Aniket/job/tp-all/36/parameters/

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -2017,15 +2017,9 @@ func (k *K8s) addSecurityAnnotation(spec interface{}, configMap *corev1.ConfigMa
 	}
 
 	if obj, ok := spec.(*storageapi.StorageClass); ok {
-		log.Infof("Deploying SC")
-		log.Infof("%v", k.secureApps)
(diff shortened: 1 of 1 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -2039,6 +2038,12 @@ func (k *K8s) addSecurityAnnotation(spec interface{}, configMap *corev1.ConfigMa
 						obj.Parameters[CsiNodePublishSecretNamespace] = configMap.Data[secretNamespaceKey]
 						obj.Parameters[CsiControllerExpandSecretNamespace] = configMap.Data[secretNamespaceKey]
 					}
+					if app.IsCSI {
```
CharudathGopal, line 2041: you are doing app.IsCSI check twice here, can this block be before line 2030 as a single if block ?

aghodke1312: This is required since the annotations will be needed only for CSI app.

CharudathGopal (approved):

PR #1546, opened 2023-07-19
Removing the refrence test testParseConfig

sn-px:
**Special notes for your reviewer**: Removing the refrence test testParseConfig


kshithijiyer-px (approved):
LGTM

PR #1549, opened 2023-07-19
Added method to create rule for multiple application in one rule

snigdha-px:
1)At present , we have function to create one rule for one application. But the test cases of 2.5.1 requires multiple application backup with rule.
2)ctx parameter is not present in current method
Hence added a function to create one rule for multiple backups.

Closes #PA-1334

log : https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2075/consoleFull
[Had purposely commented out the delete part during verification to make sure rules are created as expected]

Note : Have introduced a new variable AppParametersNew and added a new method "CreateRuleForBackupWithMultipleApplications" to add rules 
Reason for new method: 
1)At present , we have function to create one rule for one application. But the test cases of 2.5.1 requires multiple application backup with rule.
2)ctx parameter is not present in current method

Reason for new variable  AppParametersNew
1)Existing method uses AppParameter which uses map and this method is used for many test cases. Could not modify the existing variable and change to struct

Have created another task :https://portworx.atlassian.net/browse/PA-1427 in which old testcases will also use the new method to create rules and existing method CreateRuleForBackup and variable will be deleted.

### DIFF
--- a/drivers/backup/backup.go
+++ b/drivers/backup/backup.go
@@ -28,6 +28,27 @@ const (
 	Sunday            = "Sun"
 )
 
+type RuleSpec struct {
+	ActionList      []string
+	PodSelectorList []string
+	Background      []string
+	RunInSinglePod  []string
+	Container       []string
+}
+
+type PreRule struct {
+	Rule RuleSpec
+}
+
+type PostRule struct {
+	Rule RuleSpec
+}
+
+type AppRule struct {
+	PreRule  PreRule
+	PostRule PostRule
+}
+
 // Driver for backup
 type Driver interface {
 	// Org interface
--- a/tests/backup/backup_helper.go
+++ b/tests/backup/backup_helper.go
@@ -179,6 +179,63 @@ const (
 	Parallel
 )
 
+var (
+	// AppRuleMaster is a map of struct for all the value for rules
+	// This map needs to be updated for new applications as and whe required
+	AppRuleMaster = map[string]backup.AppRule{
+		"cassandra": {
+			PreRule: backup.PreRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"nodetool flush -- keyspace1;", "echo 'test"}, PodSelectorList: []string{"app=cassandra", "app=cassandra1"}, Background: []string{"false", "false"}, RunInSinglePod: []string{"false", "false"}, Container: []string{},
+				},
+			},
+			PostRule: backup.PostRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"nodetool verify -- keyspace1;", "nodetool verify -- keyspace1;"}, PodSelectorList: []string{"app=cassandra", "app=cassandra1"}, Background: []string{"false", "false"}, RunInSinglePod: []string{"false", "false"}, Container: []string{},
+				},
+			},
+		},
+		"mysql": {
+			PreRule: backup.PreRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH TABLES WITH READ LOCK;system ${WAIT_CMD};'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"true"}, RunInSinglePod: []string{"false"}, Container: []string{},
+				},
+			},
+			PostRule: backup.PostRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH LOGS; UNLOCK TABLES;'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"false"}, RunInSinglePod: []string{"false"}, Container: []string{},
+				},
+			},
+		},
+		"mysql-backup": {
+			PreRule: backup.PreRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH TABLES WITH READ LOCK;system ${WAIT_CMD};'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"true"}, RunInSinglePod: []string{"false"}, Container: []string{},
+				},
+			},
+			PostRule: backup.PostRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH LOGS; UNLOCK TABLES;'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"false"}, RunInSinglePod: []string{"false"}, Container: []string{},
+				},
+			},
+		},
+		"postgres": {
+			PreRule: backup.PreRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"PGPASSWORD=$POSTGRES_PASSWORD; psql -U \"$POSTGRES_USER\" -c \"CHECKPOINT\""}, PodSelectorList: []string{"app=postgres"}, Background: []string{"false"}, RunInSinglePod: []string{"false"}, Container: []string{},
+				},
+			},
+		},
+		"postgres-backup": {
+			PreRule: backup.PreRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"PGPASSWORD=$POSTGRES_PASSWORD; psql -U \"$POSTGRES_USER\" -c \"CHECKPOINT\""}, PodSelectorList: []string{"app=postgres"}, Background: []string{"false"}, RunInSinglePod: []string{"false"}, Container: []string{},
+				},
+			},
+		},
+	}
+)
+
 // Set default provider as aws
 func getProviders() []string {
 	providersStr := os.Getenv("PROVIDERS")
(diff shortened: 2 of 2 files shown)


drivers/backup/portworx/portworx.go
```diff
@@ -1255,6 +1255,36 @@ var (
 		},
 		},
 	}
+
+	AppParametersNew = map[string]map[string]map[string]string{
```
ak-px, line 1259: can we add rule for mysql as well.

drivers/backup/backup.go
```diff
@@ -355,6 +355,9 @@ type Rule interface {
 	// CreateRuleForBackup creates backup rule
 	CreateRuleForBackup(appName string, orgID string, prePostFlag string) (bool, string, error)
 
+	// CreateRuleForBackupWithMultipleApplications creates backup rule for multiple application
+	CreateRuleForBackupWithMultipleApplications(ruleName string, orgID string, appList []string, prePostFlag string, ctx context.Context) (bool, string, error)
```
kphalgun-px, line 359: How about we pass `AppParameters` instead of `appList`? Then, we would not have to define `AppParametersNew`.

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1565,107 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+				ps := strings.Split(AppParameters[appList[i]]["post"]["pod_selector_list"][i], "=")
+				psMap := make(map[string]string)
+				psMap[ps[0]] = ps[1]
+				podSelector = append(podSelector, psMap)
+				actionValue = append(actionValue, AppParametersNew[appList[i]]["post"]["post_action_list"])
+				backgroundVal, _ := strconv.ParseBool(AppParametersNew[appList[i]]["post"]["background"])
+				background = append(background, backgroundVal)
+				podVal, _ := strconv.ParseBool(AppParametersNew[appList[i]]["post"]["runInSinglePod"])
+				runInSinglePod = append(runInSinglePod, podVal)
+				// Here user has to set env for each app container if required in the format container-<app name>
```
kphalgun-px, line 1607: I think we can get it from AppParameters.

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1565,107 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
 	return nil
 }
 
+// CreateRuleForBackupWithMultipleApplications creates backup rule for multiple application in one rule
+func (p *portworx) CreateRuleForBackupWithMultipleApplications(ruleName string, orgID string, appList []string, prePostFlag string, ctx context.Context) (bool, string, error) {
```
kphalgun-px, line 1569: How about CreateRuleForBackupFromAppParameters?

kphalgun-px (approved):
LGTM

drivers/backup/portworx/portworx.go
snigdha-px: done

drivers/backup/portworx/portworx.go
snigdha-px: I just wanted to mention "MultipleApplication" in the name because as discussed we already have the method to create rule
But at present that method will create different rule for different app
In case we create backup with multiple applications, we will not be able to use rules for each application while taking backup.
Let me know if I should still change the name or suggest any other name .

kphalgun-px: Are you not planning to take AppParameters as input?

snigdha-px: AppParameters is like a superset containing details for each app
eg : As of now we have postgres, cassandra and mysql  details is AppParameters
I will deploy only cassandra and mysql for backup[not postgres]. Then I will pass appList which has value of only cassandra and mysql taken from ginkgo cli.

snigdha-px: 1) I am not sure about this
2)Why would I want to work with postgres rule If I do not have postgres deployed?

kphalgun-px: I thought you were creating one rule for all applications. My bad. How about you take both AppParameters and AppList as inputs and create rules based on the AppList? For instance, if I want to use custom rules for a test case, then I can create new custom AppParameters and use this function; otherwise, I would need to create a similar function again.

trenukarya-px: Agree with Krishna. @snigdha-px I would suggest to have on struct for each application and it should have its own implementation for giving pre and post rules. Please implement from that perspective.

drivers/backup/portworx/portworx.go
```diff
@@ -1254,6 +1254,32 @@ var (
 			"container":         {"", ""},
 		},
 		},
+		"mysql-backup": {"pre": {"pre_action_list": {"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH TABLES WITH READ LOCK;system ${WAIT_CMD};'"},
+			"background":        {"false"},
```
ak-px, line 1258: background is true for pre-rule as per docs.

drivers/backup/portworx/portworx.go
```diff
@@ -1254,6 +1254,32 @@ var (
...
+		},
+			"post": {"post_action_list": {"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH LOGS; UNLOCK TABLES;'"},
+				"background":        {"false"},
+				"pod_selector_list": {"app=mysql"},
+				"runInSinglePod":    {"false"},
+				"container":         {""},
+			},
+		},
+		"mysql": {"pre": {"pre_action_list": {"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH TABLES WITH READ LOCK;system ${WAIT_CMD};'"},
+			"background":        {"false"},
```
ak-px, line 1271: same as above

drivers/backup/portworx/portworx.go
snigdha-px: corrected

drivers/backup/portworx/portworx.go
snigdha-px: done

ak-px (approved):
LGTM

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1561,111 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+	_, err := p.CreateRule(ctx, RuleCreateReq)
+	if err != nil {
+		err = fmt.Errorf("failed to create backup rules: [%v]", err)
+		return false, "", err
+	}
+	log.Infof("Validate rules for backup")
+	RuleEnumerateReq := &api.RuleEnumerateRequest{
+		OrgId: orgID,
+	}
+	ruleList, err := p.EnumerateRule(ctx, RuleEnumerateReq)
```
kphalgun-px, line 1646: How about we return an error?

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1561,111 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+		Uid:   uid,
+	}
+	_, err = p.InspectRule(ctx, RuleInspectReq)
+	log.Infof(" The error in inspect rule is", err)
+	if err != nil {
+		log.InfoD(" Iside this")
+		err = fmt.Errorf("failed to validate the created rule with Error: [%v]", err)
+		return false, uid, err
+	}
+	log.InfoD(" Resched here")
```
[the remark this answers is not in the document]
snigdha-px: done

drivers/backup/portworx/portworx.go
snigdha-px: done

vpinisetti-px:
I thought you are creating single pre & post exec rules for all the applications.
Ideally it should be as below
- Create single pre & post exec rule for all the applications
- In testcase we just take decision whether you are going apply pre/post exec rule. Accordingly it will create a rule based on the application type. No need to pass the application list & rule name in the testcase.

drivers/backup/portworx/portworx.go
```diff
@@ -68,6 +68,79 @@ type portworx struct {
...
+		"mysql-backup": {"pre": {"pre_action_list": {"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH TABLES WITH READ LOCK;system ${WAIT_CMD};'"},
+			"background":        {"true"},
+			"runInSinglePod":    {"false"},
+			"pod_selector_list": {"app=mysql"},
+			"container":         {"", ""},
+		},
+			"post": {"post_action_list": {"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH LOGS; UNLOCK TABLES;'"},
+				"background":        {"false"},
+				"pod_selector_list": {"app=mysql"},
+				"runInSinglePod":    {"false"},
```
vpinisetti-px, line 111: I thought you are creating single pre & post exec rules for all the applications.
Ideally it should be as below

Create single pre & post exec rule for all the applications
In testcase we just take decision whether you are going apply pre/post exec rule. Accordingly it will create a rule based on the application type. No need to pass the application list & rule name in the testcase.

vpinisetti-px (changes requested):
Will discuss more on this approach and take a decision.

drivers/backup/backup.go
```diff
@@ -16,6 +16,14 @@ type Image struct {
 	Version string
 }
 
+type RuleStruct struct {
```
mkoppal-px, line 19: Lets not have a variable name with `Struct` in it. It should only be called `Rule`

drivers/backup/backup.go
```diff
@@ -355,6 +363,9 @@ type Rule interface {
 	// CreateRuleForBackup creates backup rule
 	CreateRuleForBackup(appName string, orgID string, prePostFlag string) (bool, string, error)
 
+	// CreateRuleForBackupWithMultipleApplications creates backup rule for multiple application
+	CreateRuleForBackupWithMultipleApplications(ruleName string, orgID string, appList []string, prePostFlag string, ctx context.Context, appParameters map[string]map[string]RuleStruct) (bool, string, error)
```
mkoppal-px, line 367: This method signature can be optimised.
We don't need to pass `prePostFlag`. The struct where we are maintaining the map of rules for apps needs to be validated. And that itself should create a pre-rule if a pre-rule exists, if not, it should not.
We also do not need to pass `appParameters map[string]map[string]RuleStruct`, it is a constant.

Why are we returning 3 things here? We can just return string and error

drivers/backup/portworx/portworx.go
```diff
@@ -68,6 +68,79 @@ type portworx struct {
...
+		"mysql": {"pre": backup.RuleStruct{ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH TABLES WITH READ LOCK;system ${WAIT_CMD};'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"true"}, RunInSinglePod: []string{"false"}, Container: []string{}},
+			"post": backup.RuleStruct{ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH LOGS; UNLOCK TABLES;'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"false"}, RunInSinglePod: []string{"false"}, Container: []string{}}},
+		"mysql-backup": {"pre": backup.RuleStruct{ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH TABLES WITH READ LOCK;system ${WAIT_CMD};'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"true"}, RunInSinglePod: []string{"false"}, Container: []string{}},
+			"post": backup.RuleStruct{ActionList: []string{"mysql --user=root --password=$MYSQL_ROOT_PASSWORD -Bse 'FLUSH LOGS; UNLOCK TABLES;'"}, PodSelectorList: []string{"app=mysql"}, Background: []string{"false"}, RunInSinglePod: []string{"false"}, Container: []string{}}},
+	}
```
mkoppal-px, line 141: This whole thing can be even more optimised by declaring our struct in this fashion:

```
type RuleSpec struct {
	ActionList      []string
	PodSelectorList []string
	Background      []string
	RunInSinglePod  []string
	Container       []string
}

type PreRule struct {
	Rule RuleSpec
}

type PostRule struct {
	Rule RuleSpec
}

type AppRule struct {
	PreRule  PreRule
	PostRule PostRule
}
```

By doing this, we can convert the map of map to just a map. Here is a sample of how the AppParams will look:

```
var(
  AppRuleMaster = map[string]backup.AppRule{
		"cassandra": {
			PreRule: backup.PreRule{
				Rule: backup.RuleSpec{
					ActionList: []string{"nodetool flush -- keyspace1;", "echo 'test"}, PodSelectorList: []string{"app=cassandra", "app=cassandra1"}, Background: []string{"false", "false"}, RunInSinglePod: []string{"false", "false"}, Container: []string{},
				},
			},
			PostRule: backup.PostRule{
				Rule: backup.RuleSpec{
					ActionList: []string{"nodetool verify -- keyspace1;", "nodetool verify -- keyspace1;"}, PodSelectorList: []string{"app=cassandra", "app=cassandra1"}, Background: []string{"false", "false"}, RunInSinglePod: []string{"false", "false"}, Container: []string{},
				},
			},
		},
	}

)
```

mkoppal-px: This will look cleaner, well formatted and most importantly we can just check if PreRule is not nil, apply Pre-rule, else don't.

mkoppal-px: Let me know you thoughts

drivers/backup/backup.go
snigdha-px: sure

drivers/backup/portworx/portworx.go
snigdha-px: done

drivers/backup/backup.go
snigdha-px: done

drivers/backup/backup.go
snigdha-px: done

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1617,166 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+		preBackground      []bool
+		preRunInSinglePod  []bool
+		preRulesInfo       api.RulesInfo
+		postRulesInfo      api.RulesInfo
+		prePodSelector     []map[string]string
+		postPodSelector    []map[string]string
+	)
+	if len(appParameters) == 0 {
+		appParameters = append(appParameters, AppRuleMaster)
+	}
```
mkoppal-px, line 1642: Can we do

```
if len(appParameters) == 0 {
		appParameter = AppRuleMaster
	} else {
                appParameter = appParameters[0]
}
```
Then we can just keep using `appParameter` instead of `appParameters[0]` everywhere

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1617,166 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+
+	for i := 0; i < len(appList); i++ {
+		if reflect.DeepEqual(appParameters[0][appList[i]].PreRule, backup.PreRule{}) {
```
mkoppal-px, line 1645: I see that we are using `appParameters[0][appList[i]]` at many places. Can we hold this in a variable and use it across?

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1617,166 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+				preBackground = append(preBackground, backgroundVal)
+				podVal, _ := strconv.ParseBool(appParameters[0][appList[i]].PreRule.Rule.RunInSinglePod[j])
+				preRunInSinglePod = append(preRunInSinglePod, podVal)
+				containerName := fmt.Sprintf("%s-%s", "container", appList[i])
+				preContainer = append(preContainer, os.Getenv(containerName))
+			}
+		}
+
+		if reflect.DeepEqual(appParameters[0][appList[i]].PostRule, backup.PostRule{}) {
+			log.Infof("Post rule not required for application %v", appList[i])
```
mkoppal-px, line 1664: For some reason the `reflect.DeepEqual` doesn't seem to work. But we can create a separate common function which checks if a field exists in the interface.

I have tried it here - https://go.dev/play/p/otE34z6d6m3

Please check

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1617,166 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+				postContainer = append(postContainer, os.Getenv(containerName))
+			}
+		}
+
+	}
+	totalPreRules := len(preActionValue)
+	totalPostRules := len(postActionValue)
+
+	if totalPreRules != 0 {
+		preRuleName = fmt.Sprintf("pre-rule-%v", time.Now().Unix())
```
mkoppal-px, line 1686: Can we change all `time.Now().Unix()` to generating random chars?

drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1617,166 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
 	return nil
 }
 
+// CreateRuleForBackupWithMultipleApplications creates backup rule for multiple application in one rule
+func (p *portworx) CreateRuleForBackupWithMultipleApplications(orgID string, appList []string, ctx context.Context, appParameters ...map[string]backup.AppRule) (string, string, error) {
```
mkoppal-px, line 1621: Can we split this into 2 different functions - `CreatePreRuleForBackupWithMultipleApplications` and `CreatePostRuleForBackupWithMultipleApplications` ?
This is because here we are returning 2 strings and if an app does not have either pre or post rule, then how are we handling it? In the test case itself?

drivers/backup/portworx/portworx.go
snigdha-px: But I see the log getting printed  for postgres which does not have post rule
17:56:54  2023-07-25 12:26:54 +0000:[INFO] [{BasicBackupCreation}] [portworx.(*portworx).CreateRuleForBackupWithMultipleApplications:#1671] - Post rule not required for application postgres-backup

drivers/backup/portworx/portworx.go
snigdha-px: done

drivers/backup/portworx/portworx.go
snigdha-px: reflect.DeepEqual(appParameters[0][appList[i]].PostRule, backup.PostRule{}) verifies if appParameters[0][appList[i]] DOES NOT have PostRule field

drivers/backup/portworx/portworx.go
snigdha-px: done.
Cannot use "RandomString" from common.go as it would create "import cycle not allowed" error

mkoppal-px: Then what we are doing here is incorrect. This function should be part of `backup_helper.go`.
Can we move the function and AppParameters? There is no reason it should be in `portworx.go`

snigdha-px: done.

tests/backup/backup_helper.go
```diff
@@ -179,6 +179,63 @@ const (
...
 )
 
+var (
+	// AppRuleMaster is a map of struct for all the value for rules
+	// This map needs to be updated for new applications as and whe required
+	AppRuleMaster = map[string]backup.AppRule{
+		"cassandra": {
+			PreRule: backup.PreRule{
+				Rule: backup.RuleSpec{
+					ActionList: []string{"nodetool flush -- keyspace1;", "echo 'test"}, PodSelectorList: []string{"app=cassandra", "app=cassandra1"}, Background: []string{"false", "false"}, RunInSinglePod: []string{"false", "false"}, Container: []string{},
```
trenukarya-px, line 189: Isn't this PodSelectorList coming from "AppParameters" defined portworx.go ?

kshithijiyer-px (approved):
LGTM

mkoppal-px (approved):
LGTM

tests/backup/backup_helper.go
snigdha-px: It is not coming from "AppParameters" defined portworx.go

As mentioned in description, once https://portworx.atlassian.net/browse/PA-1427 is merged, there will be only AppRuleMaster and AppParameter will be discarded.

trenukarya-px (approved):

vpinisetti-px (approved):
LGTM

PR #1552, opened 2023-07-20
Adding data integrity validations

lsrinivas-pure:

Closes #PTX-18922


tests/common.go
```diff
@@ -7725,3 +7777,183 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
...
+						//To-Do update the command if set up is dmthin
+						cmd := fmt.Sprintf("/opt/pwx/bin/runc exec -t portworx md5sum /var/.px/%d/%s/pxdev", poolID, inspectVolume.Id)
+						output, err := Inst().N.RunCommand(*nodeDetail, cmd, node.ConnectionOpts{
+							Timeout:         defaultTimeout,
+							TimeBeforeRetry: defaultRetryInterval,
+							Sudo:            true,
+						})
+						if err != nil {
+							return err
+						}
```
Santhoshairody, line 7890: Can you test this on aggregated volume as well

tests/common.go
```diff
@@ -7725,3 +7777,183 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
...
+			}
+
+			//reverting application scale
+			applicationScaleUpMap := make(map[string]int32, len(ctx.App.SpecList))
+
+			for name, scale := range applicationScaleMap {
+				log.InfoD(fmt.Sprintf("scale up app %s to %d", ctx.App.Key, scale))
+				applicationScaleUpMap[name] = scale
+			}
+			err = Inst().S.ScaleApplication(ctx, applicationScaleUpMap)
```
Santhoshairody, line 7912: Need a log statment here

tests/testTriggers.go
```diff
@@ -571,6 +571,38 @@ func TriggerDeployNewApps(contexts *[]*scheduler.Context, recordChan *chan *Even
...
+
+		if len(*contexts) > 2 {
+			mid := len(*contexts) / 2
+			for i, ctx := range *contexts {
+				appVolumes, err := Inst().S.GetVolumes(ctx)
+				UpdateOutcome(event, err)
+				var volumeSpecUpdate *opsapi.VolumeSpecUpdate
+				if i < mid {
+					volumeSpecUpdate = &opsapi.VolumeSpecUpdate{
+						IoThrottleOpt: &opsapi.VolumeSpecUpdate_IoThrottle{
```
Santhoshairody, line 583: Do it only on few FIO volumes only . Can you measure how much IOPS and Bandwidth consistently getting on these FIO volumes (may be check for 15 minutes and then throttle IOPS and Bandwidth to it's 50% so that we throttle to the correct number.

tests/testTriggers.go
```diff
@@ -571,6 +571,38 @@ func TriggerDeployNewApps(contexts *[]*scheduler.Context, recordChan *chan *Even
...
+								WriteIops: 1024,
+							},
+						},
+					}
+				}
+				for _, appVol := range appVolumes {
+					log.Infof(fmt.Sprintf("updating volume %s [app:%s] with volume spec: %+v", appVol.Name, ctx.App.Key, volumeSpecUpdate))
+					err = Inst().V.UpdateVolumeSpec(appVol, volumeSpecUpdate)
+				}
+			}
```
Santhoshairody, line 604: We should also need to verify that IOPS and Bandwidth is not going beyond throttled value.

tests/common.go
```diff
@@ -7725,3 +7777,183 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
...
+			for _, v := range appVolumes {
+				replicaSets, err := Inst().V.GetReplicaSets(v)
+				if err != nil {
+					return err
+				}
+
+				var checksum string
+				var primaryPool string
+				var primaryNode *node.Node
+				for _, rs := range replicaSets {
```
Santhoshairody, line 7864: Skip if repl is 1

Santhoshairody (commented):
Regarding DataIntegrity test. Data verification is a long running operations so it might long time to finish the test if you enable this on all the test.
IMO we should have a knob per test as well, so that while running test , user can decide which are the tests they need to run dataverification.

tests/common.go
```diff
@@ -7725,3 +7777,183 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
 	return poolSizeUsed, nil
 }
 
+func ValidateDataIntegrity(contexts *[]*scheduler.Context) error {
+
+	if enableDataIntegrityValidation && (Inst().N.String() == ssh.DriverName || Inst().N.String() == vsphere.DriverName) {
```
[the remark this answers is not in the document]
lsrinivas-pure: yes only one test runs at a time

tests/common.go
```diff
@@ -7725,3 +7777,183 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
...
+							return err
+						}
+
+						inspectVolume, err := Inst().V.InspectVolume(v.ID)
+						if err != nil {
+							return err
+						}
+						log.Infof("Getting md5sum for volume %s on pool %s in node %s", inspectVolume.Id, poolID, nodeDetail.Name)
+						//To-Do update the command if set up is dmthin
+						cmd := fmt.Sprintf("/opt/pwx/bin/runc exec -t portworx md5sum /var/.px/%d/%s/pxdev", poolID, inspectVolume.Id)
```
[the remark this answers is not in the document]
lsrinivas-pure: fixed

tests/common.go
lsrinivas-pure: it already added while creating applicationScaleUpMap

tests/common.go
lsrinivas-pure: I have separate PR in-progress for aggregated volumes, will test this as part of that

tests/common.go
lsrinivas-pure: fixed

tests/common.go
```diff
@@ -7728,3 +7781,274 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
...
+
+		//waiting for volumes replication status should be up before calculating md5sum
+		for _, v := range appVolumes {
+			replicaSets, err := Inst().V.GetReplicaSets(v)
+			if err != nil {
+				return err
+			}
+
+			//skipping the validation if volume is repl 2
+			if len(replicaSets) == 1 && len(replicaSets[0].PoolUuids) == 1 {
```
Rohit-PX, line 7848: avoid the use of `continue` whenever possible.
Instead change the condition to be:
```
      if len(replicaSets) > 1 && len(replicaSets[0].PoolUuids) > 1 {
          ..............
          ..............
      }
```

tests/common.go
```diff
@@ -7728,3 +7781,283 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
...
+			t := func() (interface{}, bool, error) {
+				replicationStatus, err := GetVolumeReplicationStatus(v)
+				if err != nil {
+					return nil, false, err
+				}
+
+				if replicationStatus == "Up" {
+					return "", false, nil
+				}
+				if replicationStatus == "Resync" {
```
Santhoshairody, line 7860: What about other state like `Degraded` and `not in Quorum` , should n't need to skip?

tests/common.go
```diff
@@ -7728,3 +7781,283 @@ func GetPoolCapacityUsed(poolUUID string) (float64, error) {
...
+		applicationScaleUpMap := make(map[string]int32, len(ctx.App.SpecList))
+
+		for name, scale := range applicationScaleMap {
+			log.InfoD(fmt.Sprintf("scale up app %s to %d", ctx.App.Key, scale))
+			applicationScaleUpMap[name] = scale
+		}
+		err = Inst().S.ScaleApplication(ctx, applicationScaleUpMap)
+		if err != nil {
+			return err
+		}
```
Santhoshairody, line 8012: scaling all apps at a  time can lead to push lots of data at a time, which can lead to high latency on the pool . can we scaleup corresponding app just after checksum verification so that we donot lots of IO at a time.

Santhoshairody (approved):
Please address the review comments before merging.

aravindranath14 (approved):
LGTM

tests/testTriggers.go
lsrinivas-pure: will take as part of different when how to verify this is cleared

tests/testTriggers.go
lsrinivas-pure: will take as part of different when how to verify this is cleared

tests/common.go
lsrinivas-pure: in both the case it will come out and fail

tests/common.go
lsrinivas-pure: yes it scales only the app which we are doing checksum validation

PR #1560, opened 2023-07-24
Logivity test suite for add drive

vprabhakar-px:
longivity test case addition to add new pool and resize disk on the new pool created


lsrinivas-pure (approved):
lgtm

PR #1578, opened 2023-07-27
Vanilla pipeline is failing intermittently while getting the Targetcl…

jkagliwal-px:
Added a polling logic to check the cluster health check and error-out if not in "Healthy" state. 

Supporting Logs - 
1. When cluster is HEALTHY - https://aetos.pwx.purestorage.com/resultSet/testSetID/281979
2. When cluster is NOT HEALTHY - Attached screenshot in the JIRA Ticket. 
     https://portworx.atlassian.net/browse/PA-1440

### DIFF
--- a/drivers/pds/targetcluster/targetcluster.go
+++ b/drivers/pds/targetcluster/targetcluster.go
@@ -37,6 +37,7 @@ const (
 	MaxTimeout     = 30 * time.Minute
 	timeOut        = 30 * time.Minute
 	timeInterval   = 10 * time.Second
+	minTimeOut     = 5 * time.Minute
 
 	// PDSNamespace PDS
 	PDSNamespace = "pds-system"
@@ -69,14 +70,14 @@ type TargetCluster struct {
 
 func (tc *TargetCluster) GetDeploymentTargetID(clusterID, tenantID string) (string, error) {
 	log.InfoD("Get the Target cluster details")
-	err = wait.Poll(DefaultRetryInterval, 5*time.Minute, func() (bool, error) {
+	err = wait.Poll(DefaultRetryInterval, minTimeOut, func() (bool, error) {
 		targetClusters, err := components.DeploymentTarget.ListDeploymentTargetsBelongsToTenant(tenantID)
 		var targetClusterStatus string
 		if err != nil {
-			return true, fmt.Errorf("error while listing deployments: %v", err)
+			return true, fmt.Errorf("error while listing deployment targets: %v", err)
 		}
 		if targetClusters == nil {
-			return true, fmt.Errorf("target cluster passed is not available to the account/tenant %v", err)
+			return true, fmt.Errorf("target cluster passed is not available to the account/tenant")
 		}
 		for i := 0; i < len(targetClusters); i++ {
 			if targetClusters[i].GetClusterId() == clusterID {
--- a/drivers/pds/targetcluster/targetcluster.go
+++ b/drivers/pds/targetcluster/targetcluster.go
@@ -69,14 +69,14 @@ type TargetCluster struct {
 
 func (tc *TargetCluster) GetDeploymentTargetID(clusterID, tenantID string) (string, error) {
 	log.InfoD("Get the Target cluster details")
-	err = wait.Poll(DefaultRetryInterval, 5*time.Minute, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


drivers/pds/targetcluster/targetcluster.go
```diff
@@ -85,9 +85,13 @@ func (tc *TargetCluster) GetDeploymentTargetID(clusterID, tenantID string) (stri
 			targetClusterStatus = targetClusters[i].GetStatus()
 		}
 	}
-	if targetClusterStatus != "healthy" {
-		return "", fmt.Errorf("target Cluster is not in healthy state due to error : %v", err)
-	}
+	err = wait.Poll(DefaultRetryInterval, 5*time.Minute, func() (bool, error) {
+		if targetClusterStatus != "healthy" {
```
Madanagopal19, line 89: targetClusterStatus will not get updated, unless you call 
`targetClusters, err := components.DeploymentTarget.ListDeploymentTargetsBelongsToTenant(tenantID)` this inside the polling method.

Madanagopal19 (changes requested):

drivers/pds/targetcluster/targetcluster.go
```diff
@@ -69,25 +69,29 @@ type TargetCluster struct {
...
-		if targetClusters[i].GetClusterId() == clusterID {
-			deploymentTargetID = targetClusters[i].GetId()
-			log.Infof("deploymentTargetID %v", deploymentTargetID)
-			log.InfoD("Cluster ID: %v, Name: %v,Status: %v", targetClusters[i].GetClusterId(), targetClusters[i].GetName(), targetClusters[i].GetStatus())
-			targetClusterStatus = targetClusters[i].GetStatus()
+	err = wait.Poll(DefaultRetryInterval, MaxTimeout, func() (bool, error) {
+		targetClusters, err := components.DeploymentTarget.ListDeploymentTargetsBelongsToTenant(tenantID)
+		var targetClusterStatus string
+		if err != nil {
+			return true, fmt.Errorf("error while listing deployments: %v", err)
```
Madanagopal19, line 76: return false

drivers/pds/targetcluster/targetcluster.go
```diff
@@ -69,25 +69,29 @@ type TargetCluster struct {
...
 		}
-	}
-	if targetClusterStatus != "healthy" {
-		return "", fmt.Errorf("target Cluster is not in healthy state due to error : %v", err)
-	}
+		if targetClusters == nil {
+			return true, fmt.Errorf("target cluster passed is not available to the account/tenant %v", err)
```
Madanagopal19, line 79: return false, why are you returning true ? you need to return true once the condition is satisfied.

drivers/pds/targetcluster/targetcluster.go
Madanagopal19:
```
if targetClusterStatus == "healthy" {
   return true, nil
}
```

drivers/pds/targetcluster/targetcluster.go
Madanagopal19: and also error will be nil here, no need to pass err

Madanagopal19 (changes requested):

drivers/pds/targetcluster/targetcluster.go
```diff
@@ -69,24 +69,31 @@ type TargetCluster struct {
...
-	if targetClusters == nil {
-		return "", fmt.Errorf("target cluster passed is not available to the account/tenant %v", err)
-	}
-	for i := 0; i < len(targetClusters); i++ {
-		if targetClusters[i].GetClusterId() == clusterID {
-			deploymentTargetID = targetClusters[i].GetId()
-			log.Infof("deploymentTargetID %v", deploymentTargetID)
-			log.InfoD("Cluster ID: %v, Name: %v,Status: %v", targetClusters[i].GetClusterId(), targetClusters[i].GetName(), targetClusters[i].GetStatus())
-			targetClusterStatus = targetClusters[i].GetStatus()
+	err = wait.Poll(DefaultRetryInterval, 5*time.Minute, func() (bool, error) {
```
Madanagopal19, line 72: add a constant for defaulttime out

Madanagopal19 (approved):
LGTM

aravindranath14 (approved):
LGTM

PR #1598, opened 2023-08-04
PWX-32475: Create a copy of the original storage class and set the binding mode to immediate

ragsingh-px:
Create a copy of the original storage class and set the binding mode to immediate for CSICloneTest and CSI snapshot test
PWX-32475
This PR fixes the issue with the CSICloneTest and CSISnapshotTest where the PVCs are waiting for the pod to come up because the VolumeBindingMode is set to WaitForFirstConsumer. However, since the method that waits for PVC to get bounded is on loop, the method to to create pods does not get called and PVC waits and subsequently get timed out. This issue can be resolved by setting the VolumeBindingMode to Immediate. But instead of doing that to the original storageClass spec which  might be needed for other tests, we create a clone of the StorageClass, modify the VolumeBindingMode and use it for our tests.
Closes #

Jenkins link:- https://jenkins.pwx.dev.purestorage.com/job/Users/job/Hari/job/forragavendra/16/
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Hari/job/forragavendra/19/


drivers/scheduler/k8s/k8s.go
```diff
@@ -7517,3 +7528,25 @@ func ClusterVersion() (string, error) {
...
 }
+
+func createClonedStorageClassIfRequired(originalStorageClass *storageapi.StorageClass) (string, error) {
+	clonedSCName := originalStorageClass.Name + "-clone"
+	clonedSCobj, err := k8sStorage.GetStorageClass(clonedSCName)
+	if err == nil {
+		return clonedSCobj.Name, nil
+	} else {
+		log.Infof("Cloned SC with the name %v does not exist and has to be created", clonedSCName)
+		originalStorageClass.Name = clonedSCName
```
alicelyu-px, line 7539: Mutation of Input Parameter should always be avoided. Please make a copy of the original instead or create new from scratch.

drivers/scheduler/k8s/k8s.go
```diff
@@ -7517,3 +7528,25 @@ func ClusterVersion() (string, error) {
 	}
 	return strings.TrimLeft(ver.String(), "v"), nil
 }
+
+func createClonedStorageClassIfRequired(originalStorageClass *storageapi.StorageClass) (string, error) {
+	clonedSCName := originalStorageClass.Name + "-clone"
+	clonedSCobj, err := k8sStorage.GetStorageClass(clonedSCName)
```
alicelyu-px, line 7534: A `Get` method returning an error does not necessarily mean the object does not exist. The error should be checked for network issues.

drivers/scheduler/k8s/k8s.go
```diff
@@ -7517,3 +7528,25 @@ func ClusterVersion() (string, error) {
...
+		originalStorageClass.Name = clonedSCName
+		immediate := storageapi.VolumeBindingImmediate
+		originalStorageClass.VolumeBindingMode = &immediate
+		originalStorageClass.ResourceVersion = ""
+		originalStorageClass.UID = ""
+
+		clonedSCobj, err := k8sStorage.CreateStorageClass(originalStorageClass)
+		if err != nil {
+			log.Infof("Error occurred while creating cloned storage class:- %v", err)
+			return "", fmt.Errorf("failed to create cloned storage class with the name %s : %s", originalStorageClass.Name, err)
```
alicelyu-px, line 7548: nit: "failed to create cloned storage class" should be added by the caller. 
It would be more clear to consistently use error reporting for errors that should be handled by the caller, and use logging for informational messages or errors that this function can recover from.

drivers/scheduler/k8s/k8s.go
```diff
@@ -6400,8 +6405,13 @@ func (k *K8s) CSICloneTest(ctx *scheduler.Context, request scheduler.CSICloneReq
 	if err != nil {
 		return fmt.Errorf("failed to retrieve storage class for PVC %s in namespace: %s : %s", request.OriginalPVCName, request.Namespace, err)
 	}
-	storageClassName := originalStorageClass.Name
-	log.Infof("Proceeding with storage class %s", storageClassName)
+
+	//PWX-32475:- Creating a clone of the original SC so that we can change the VolumeBindingMode to Immediate without changing the original storageclass spec
+	storageClassName, err := createClonedStorageClassIfRequired(originalStorageClass)
+	if err != nil {
```
alicelyu-px, line 6411: Another logging and error consistency issue. 
Please consistently use error reporting for errors that should be handled by the caller, and use logging for informational messages or errors that this function can recover from.

alicelyu-px (changes requested):
Please see comments.

drivers/scheduler/k8s/k8s.go
ragsingh-px: Created a copy.

drivers/scheduler/k8s/k8s.go
ragsingh-px: made changes to explicitly check for not found errors.

drivers/scheduler/k8s/k8s.go
ragsingh-px: fixed.

drivers/scheduler/k8s/k8s.go
ragsingh-px: fixed.

ragsingh-px:
> Please see comments.

Addressed review comments. Please review.

drivers/scheduler/k8s/k8s.go
```diff
@@ -7517,3 +7528,29 @@ func ClusterVersion() (string, error) {
...
 	return strings.TrimLeft(ver.String(), "v"), nil
 }
+
+func createClonedStorageClassIfRequired(originalStorageClass *storageapi.StorageClass) (string, error) {
+	cloneStorageClass := originalStorageClass
+	clonedSCName := cloneStorageClass.Name + "-clone"
+	clonedSCobj, err := k8sStorage.GetStorageClass(clonedSCName)
+	if err == nil {
+		return clonedSCobj.Name, nil
+	} else if strings.Contains(err.Error(), fmt.Sprintf("storageclasses.storage.k8s.io \"%v\" not found", clonedSCName)) {
```
alicelyu-px, line 7538: nit for maintainability: 
Could you please provide a source of this error message format? 
If this code ever fails to function appropriately, future code maintainer should be able to find exactly where such a format is defined/changed.

drivers/scheduler/k8s/k8s.go
```diff
@@ -7517,3 +7528,29 @@ func ClusterVersion() (string, error) {
...
+		log.Infof("Cloned SC with the name %v does not exist and will be created", clonedSCName)
+
+		cloneStorageClass.Name = clonedSCName
```
alicelyu-px, line 7541: Has this been tested? 
If err isn't nil, the object `clonedStorageClass` is likely nil, and we will end up with a kernel panic.

drivers/scheduler/k8s/k8s.go
```diff
@@ -7517,3 +7528,29 @@ func ClusterVersion() (string, error) {
...
+		cloneStorageClass.ResourceVersion = ""
+		cloneStorageClass.UID = ""
+
+		clonedSCobj, err := k8sStorage.CreateStorageClass(cloneStorageClass)
+		if err != nil {
+			log.Infof("Error occurred while creating cloned storage class:- %v", err)
+			return "", err
+		}
+		return clonedSCobj.Name, nil
+	} else {
```
alicelyu-px, line 7553: This `else` case should resolve errors other than "not found", which I suppose is likely network issues. 
We should at least automatically retry for a few times with some wait mechanism before returning error.

alicelyu-px (changes requested):
Left a few comments. 
Additionally, please provide a detailed testing note.

drivers/scheduler/k8s/k8s.go
ragsingh-px: I printed the value of the err variable when the storage class is not present.

drivers/scheduler/k8s/k8s.go
ragsingh-px: Yes, I tested it. I have attached a jenkins job link in the description above.

drivers/scheduler/k8s/k8s.go
alicelyu-px: Could you please find the source code of that error and include a link in the comment?

drivers/scheduler/k8s/k8s.go
alicelyu-px: Removing reference to `clonedSCobj` when `err != nil` frees us from the risk of a kernel panic, however shall we still include the storage class name in this log?  Can simply use `clonedSCName`

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: Bump on this comment, I agree, hard error checks like this make me nervous for maintainability

Pure-AdamuKaapan (approved):
The core of the code LGTM, approved but please address @alicelyu-px 's comments.

drivers/scheduler/k8s/k8s.go
ragsingh-px: I have added a retry mechanism.

drivers/scheduler/k8s/k8s.go
ragsingh-px: done

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: You can use `errors.IsNotFound(err)` I believe.

ragsingh-px: Cannot use errors.IsNotFound(err) since it ss not able to catch the error.

alicelyu-px: The source of error is k8s server but the error is returned to us by k8s client. 
Wdym by "Cannot use errors.IsNotFound(err) since it is not able to catch the error"? 

Using package "k8s.io/apimachinery/pkg/api/errors" and call `errors.IsNotFound(err)` to check error is widely used in torpedo in very similar use cases. 
For example: https://github.com/portworx/torpedo/blob/6b4160c1070b796bb10309d3a57cd2776b02736a/drivers/scheduler/k8s/k8s.go#L3844

Pure-AdamuKaapan: Discussed in the sync-up meeting: the confusion was trying to use the `IsNotFound` method outside of the Retry method. If we handle it inside the retry loop, it should work properly. (The issue was `errors.IsNotFound` doesn't work with the weird combined errors that the retry loop returns).

Pure-AdamuKaapan (approved):
LGTM! Thanks for taking all the feedback

drivers/scheduler/k8s/k8s.go
```diff
@@ -7624,3 +7634,45 @@ func ClusterVersion() (string, error) {
 	}
 	return strings.TrimLeft(ver.String(), "v"), nil
 }
+
+func createClonedStorageClassIfRequired(originalStorageClass *storageapi.StorageClass) (string, error) {
+	cloneStorageClass := originalStorageClass
```
alicelyu-px, line 7639: The line cloneStorageClass := originalStorageClass doesn't actually create a new instance. Instead, both pointer variables point to the same memory location, meaning that changes to one will affect the other. This can lead to unexpected behavior. In general we should always deep copy the original object when creating a clone.

drivers/scheduler/k8s/k8s.go
```diff
@@ -7624,3 +7634,45 @@ func ClusterVersion() (string, error) {
...
+
+			cloneStorageClass.Name = clonedSCName
+			immediate := storageapi.VolumeBindingImmediate
+			cloneStorageClass.VolumeBindingMode = &immediate
+			cloneStorageClass.ResourceVersion = ""
+			cloneStorageClass.UID = ""
+
+			clonedSCobj, err := k8sStorage.CreateStorageClass(cloneStorageClass)
+			if err != nil {
+				log.Infof("Error occurred while creating cloned storage class:- %v", err)
```
alicelyu-px, line 7668: Should this be a higher severity log like `log.Errorf` since it indicates a failure in the function's main task?

drivers/scheduler/k8s/k8s.go
```diff
@@ -7624,3 +7634,45 @@ func ClusterVersion() (string, error) {
...
+func createClonedStorageClassIfRequired(originalStorageClass *storageapi.StorageClass) (string, error) {
+	cloneStorageClass := originalStorageClass
+	clonedSCName := cloneStorageClass.Name + "-clone"
+	clonedSCNotFound := false
+
+	t := func() (interface{}, bool, error) {
+		_, err := k8sStorage.GetStorageClass(clonedSCName)
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				clonedSCNotFound = true
```
alicelyu-px, line 7647: Is this `clonedSCNotFound` variable necessary?

drivers/scheduler/k8s/k8s.go
```diff
@@ -7624,3 +7634,45 @@ func ClusterVersion() (string, error) {
...
+			clonedSCobj, err := k8sStorage.CreateStorageClass(cloneStorageClass)
+			if err != nil {
+				log.Infof("Error occurred while creating cloned storage class:- %v", err)
+				return "", err
+			}
+			return clonedSCobj.Name, nil
+		} else {
+			return "", err
+		}
+	} else {
```
alicelyu-px, line 7675: This `else` seems unnecessary.
```
if err != nil {
    // handle error
    return
}

// handle success
```

alicelyu-px (changes requested):

drivers/scheduler/k8s/k8s.go
ragsingh-px: Fixed.

drivers/scheduler/k8s/k8s.go
ragsingh-px: Yes. Fixed it.

drivers/scheduler/k8s/k8s.go
ragsingh-px: fixed.

alicelyu-px (approved):
Thanks for resolving all comments!

PR #1601, opened 2023-08-08
Added a Fix to Poll TC status in control plane

Madanagopal19:
This change is required because of DS-6012
I have fixed the affected test cases to poll for the target cluster health status in control plane for 30min. 

Execution Logs 
https://jenkins.pwx.dev.purestorage.com/blue/organizations/jenkins/PDS%2Fpds-branch-build-test/detail/pds-branch-build-test/480/


aravindranath14 (approved):
LGTM

PR #1604, opened 2023-08-08
Exposing few critical Torpedo methods as APIs for Pxone

dbhatnagar-px:
Exposing 10 APIs in this PR. Will raise another PR for next 10. 

	router.DELETE("pxone/deletens/:namespace", utils.DeleteNS)
	router.POST("pxone/createns", utils.CreateNS)
	router.POST("pxone/inittorpedo", utils.InitializeDrivers)
	router.GET("pxone/getnodes", utils.GetNodes)
	router.POST("pxone/rebootnode/:nodename", utils.RebootNode)
	router.GET("pxone/storagenodes", utils.GetStorageNodes)
	router.GET("pxone/storagelessnodes", utils.GetStorageLessNodes)
	router.POST("pxone/collectsupport", utils.CollectSupport)

"Reboot Node" API is divided into 3 calls:
/rebootnode/all -> reboots all worker nodes
/rebootnode/random -> reboots a randomly selected node
/rebootnode/<nodename> -> Reboots the node whose name is provided


apiServer/pxone/utils/torpedoutils.go
```diff
@@ -0,0 +1,164 @@
...
+// pxone/rebootnode/all - Reboots all nodes in cluster
+// pxone/rebootnode/random - Reboots a randomly selected node
+// pxone/rebootnode/<nodename> - Reboots a node with specific node name
+func RebootNode(c *gin.Context) {
+	if !checkTorpedoInit(c) {
+		return
+	}
+	nodename := c.Param("nodename")
+	if nodename == "all" {
+		nodes := node.GetWorkerNodes()
```
lsrinivas-pure, line 63: since this is common you call this once after line no: 61

apiServer/pxone/utils/torpedoutils.go
```diff
@@ -0,0 +1,164 @@
...
+		for _, n := range nodes {
+			log.InfoD("reboot node: %s", n.Name)
```
lsrinivas-pure, line 65: infoD might not be apt as here you not having testcaseID to post the data to aetos
else you can leave it as it is if you have plans to set testcaseID implementation in future

apiServer/pxone/utils/torpedoutils.go
```diff
@@ -0,0 +1,164 @@
...
+				},
+			})
+			if err != nil {
+				c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+				return
+			}
+		}
+		c.JSON(http.StatusOK, gin.H{"message": "All Nodes successfully rebooted"})
+		return
+	} else if nodename == "random" {
```
lsrinivas-pure, line 80: that if ellsif else can we have switch

apiServer/pxone/utils/torpedoutils.go
```diff
@@ -0,0 +1,164 @@
...
+// RebootNode - API Call to Reboot node :
+// pxone/rebootnode/all - Reboots all nodes in cluster
+// pxone/rebootnode/random - Reboots a randomly selected node
+// pxone/rebootnode/<nodename> - Reboots a node with specific node name
+func RebootNode(c *gin.Context) {
+	if !checkTorpedoInit(c) {
+		return
+	}
+	nodename := c.Param("nodename")
+	if nodename == "all" {
```
lsrinivas-pure, line 62: Do you want to consider , separated multiple name?

apiServer/pxone/apiserver.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// We will define all API calls here.
+// Once Gin Server starts, it will initialise all APIs it contains.
+// Future work : To have segregated APIs based on need -> We will have to create multiple main calls for initialising.
+func main() {
+	router := gin.Default()
+	router.DELETE("pxone/deletens/:namespace", utils.DeleteNS)
+	router.POST("pxone/createns", utils.CreateNS)
+	router.POST("pxone/inittorpedo", utils.InitializeDrivers)
```
lsrinivas-pure, line 16: Do you want this api exposed? I think you can just call it by default when the app starts

apiServer/pxone/utils/torpedoutils.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// ScheduleAppsAndValidate : This API schedules multiple applications on the cluster and validates them
+// context is created as a global context to be accessed later in further tests
+func ScheduleAppsAndValidate(c *gin.Context) {
+	if !checkTorpedoInit(c) {
+		return
+	}
+	context = tests.ScheduleApplications(testName)
+	tests.ValidateApplications(context)
```
lsrinivas-pure, line 160: how are you planning to report the failure in case of error in validations?

apiServer/pxone/apiserver.go
dbhatnagar-px: I think it's better to expose and this will be then driven by UI Tests. Not every UI test would want driver functionality from Torpedo actually.

apiServer/pxone/utils/torpedoutils.go
dbhatnagar-px: Noted. Will put this API in next PR. Thanks

apiServer/pxone/utils/torpedoutils.go
dbhatnagar-px: Ya I was double minded regarding this. I think your suggestion is right. Let's remove it as of now and later once we reach a level where we can integrate with testcaseID, we can do these changes. Thanks

apiServer/pxone/utils/torpedoutils.go
dbhatnagar-px: Thanks for pointing this out. Made this change

apiServer/pxone/utils/torpedoutils.go
dbhatnagar-px: Noted. I will make this to switch-case in the next PR when I add API support for multiple node names

apiServer/pxone/utils/torpedoutils.go
dbhatnagar-px: I will be writing another handler to read from errChan in subsequent PRs here. If you want for now we can remove this Validation from here and bring all of that code all in. What do you suggest ?

lsrinivas-pure (approved):
LGTM

PR #1609, opened 2023-08-10
Topic/krishna px/pa 1528 v2

kphalgun-px:
This PR aims to prevent test case failures caused by changes since Px-Backup version 2.5.2.

Closes #PA-1528

To test this fix, I registered a user and added a cluster named 'source-cluster' before starting the test. We expect the InspectCluster to fail when verifying the 'source-cluster' addition using admin ctx, because there are two clusters with the same name. To fix this we need to the pass the cluster UID in the InspectClusterRequest.

Please review the Jenkins build for the test case "BasicBackupCreation" before fix using the link below

Aetos Dashboard: 

Please review the Jenkins build for the test case "BasicBackupCreation" after fix using the link below

https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2219/

Aetos Dashboard: https://aetos.pwx.purestorage.com/resultSet/testSetID/299311

I have started a run on both the S3 and NFS nightly pipelines. I will provide an update once they are completed.

### DIFF
--- a/drivers/backup/auth.go
+++ b/drivers/backup/auth.go
@@ -30,6 +30,8 @@ var PxCentralAdminPwd string
 const (
 	// PxCentralAdminUser px central admin
 	PxCentralAdminUser = "px-central-admin"
+	// PxCentralAdminGivenName is the given name of the PxCentralAdminUser
+	PxCentralAdminGivenName = "pxcentraladmin"
 	// PxCentralAdminSecretName secret for PxCentralAdminUser
 	PxCentralAdminSecretName = "px-central-admin"
 	// PxCentralAdminSecretNamespace namespace of PxCentralAdminSecretName
@@ -761,7 +763,7 @@ func GetPxCentralAdminToken() (string, error) {
 	return token, nil
 }
 
-// GetCtxWithToken getx ctx with passed token
+// GetCtxWithToken gets ctx with passed token
 func GetCtxWithToken(token string) context.Context {
 	ctx := context.Background()
 	md := metadata.New(map[string]string{
--- a/drivers/backup/backup.go
+++ b/drivers/backup/backup.go
@@ -134,6 +134,9 @@ type Cluster interface {
 	// EnumerateCluster enumerates the cluster objects
 	EnumerateCluster(ctx context.Context, req *api.ClusterEnumerateRequest) (*api.ClusterEnumerateResponse, error)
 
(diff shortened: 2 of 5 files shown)


drivers/backup/portworx/portworx.go
```diff
@@ -1535,6 +1592,97 @@ func (p *portworx) DeleteBackupSchedulePolicy(orgID string, policyList []string)
...
+	return claims, nil
+}
+
+// GetGivenNameFromCtx extracts and decodes the JWT token from the outgoing context and then returns the given name
+func GetGivenNameFromCtx(ctx context.Context) (string, error) {
+	claims, err := GetTokenClaimsFromCtx(ctx)
+	if err != nil {
+		return "", err
+	}
+	givenName, ok := claims["given_name"].(string)
```
mkoppal-px, line 1623: Why can't we use `preferred_username` instead of `given_name` ?

drivers/backup/backup.go
```diff
@@ -134,6 +134,9 @@ type Cluster interface {
 	// EnumerateCluster enumerates the cluster objects
 	EnumerateCluster(ctx context.Context, req *api.ClusterEnumerateRequest) (*api.ClusterEnumerateResponse, error)
 
+	// EnumerateAllCluster enumerates all cluster objects
```
mkoppal-px, line 137: Make the doc string more verbose and explain why we added this function

drivers/backup/portworx/portworx.go
kphalgun-px: I feel that preferred_name is optional. I am not sure what the difference is, though.

drivers/backup/auth.go
```diff
@@ -30,6 +30,8 @@ var PxCentralAdminPwd string
 const (
 	// PxCentralAdminUser px central admin
 	PxCentralAdminUser = "px-central-admin"
+	// PxCentralAdminGivenName is the given name of the PxCentralAdminUser
+	PxCentralAdminGivenName = "pxcentraladmin"
```
mkoppal-px, line 34: We can remove this since are going to use `preferred_name` now and it will be same as `PxCentralAdminUser`

kphalgun-px:
Thanks for commenting. I have raised another PR #1647 with more changes.

PR #1613, opened 2023-08-14
Refactored code, removed duplicate code,  vendoring it to Stork

gejain:
Refactored code, removed duplicate code,  vendoring it to Stork
Closes #

Result - https://aetos.pwx.purestorage.com/resultSet/testSetID/304027/testCaseID/1699261


Rohit-PX (approved):
LGTM.

Rohit-PX:
@gejain please rebase before merging.

tests/common.go
```diff
@@ -8691,3 +8702,71 @@ func UpdateCloudCredentialOwnership(cloudCredentialName string, cloudCredentialU
...
+	pods_migrated_len := len(pods_migrated.Items)
+	log.InfoD("Num of Pods on dest: %v", pods_migrated_len)
+	if pods_created_len != pods_migrated_len {
+		return fmt.Errorf("Pods migration failed as %v pods found on source and %v on destination", pods_created_len, pods_migrated_len)
+	}
+	destClusterConfigPath, err := GetDestinationClusterConfigPath()
+	if err != nil {
+		return err
+	}
+	err = asyncdr.ValidateCRD(appData.ExpectedCrdList, destClusterConfigPath)
```
diptiranjanpx, line 8753: Should not we be validating the CR itself?

gejain: This will verify CRD seperately and we CR validation will be pods on source and destination.

diptiranjanpx: Its better to do the CR validation in fact, pods are just an aftereffect of the CR being present. If it cant happen in this PR, its ok but ideally we should check for the CR migration itself.

gejain: Sure, I will create a backlog and assign it to me for this.

PR #1618, opened 2023-08-17
PWX-32982: introduce buffer to handle presence of journal device

alicelyu-px:


harsh-px (approved):
Thanks for doing this. Did you confirm if this works on a job?

lsrinivas-pure (approved):
LGTM

PR #1627, opened 2023-08-24
Adding new helm execute API

dbhatnagar-px:
Modifying Existing API to schedule apps as per review comments from last PR. Also adding API to execute Helm command.

### DIFF
--- a/apiServer/pxone/utils/torpedoutils.go
+++ b/apiServer/pxone/utils/torpedoutils.go
@@ -153,10 +153,6 @@ func ScheduleAppsAndValidate(c *gin.Context) {
 		return
 	}
 	appToRun := c.Param("appName")
-	apps := tests.Inst().AppList
-	defer func() {
-		tests.Inst().AppList = apps
-	}()
 	tests.Inst().AppList = []string{appToRun}
 	context = tests.ScheduleApplications(testName)
 	for _, ctx := range context {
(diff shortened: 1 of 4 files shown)


apiServer/pxone/utils/torpedoutils.go
```diff
@@ -152,9 +152,32 @@ func ScheduleAppsAndValidate(c *gin.Context) {
...
+	defer func() {
+		tests.Inst().AppList = apps
+	}()
+	tests.Inst().AppList = []string{appToRun}
 	context = tests.ScheduleApplications(testName)
-	tests.ValidateApplications(context)
-	c.JSON(http.StatusOK, gin.H{
-		"message": "Apps Created and Validated successfully",
-	})
+	for _, ctx := range context {
```
lsrinivas-pure, line 162: can you to errGroup to make validations faster.
https://bostonc.dev/blog/go-errgroup
also can we take boolean param whether to perform validations or not

dbhatnagar-px: My bad - the above extra code must have giving this impression that we are running multiple apps here. That wasn't meant to be there and I have removed it. This API is meant to run only one App per call. The app to run is defined in the API call itself: `router.POST("pxone/scheduleapps/:appName", utils.ScheduleAppsAndValidate) `

So, we won't require errgroups here as it's always one app. Thanks for catching it.

mkoppal-px (approved):
LGTM

lsrinivas-pure (approved):
LGTM

PR #1639, opened 2023-09-01
Added DeleteAppNamespace in cleanup to delete namespace for rancher testcases

snigdha-px:
Namespaces remain in Terminating state which cause next test cases to fail. 
Closes #

1) Changed log.Errorf to log.Warnfin cloud cred creation as it shows in log message as failed for RKE which is not correct
2)Used DeleteAppNamespaceto deleted restored namespaces from source and destination cluster

** Jenkins run : https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2724/consoleFull
https://aetos.pwx.purestorage.com/resultSet/testSetID/355868


mkoppal-px (changes requested):
Please do the following as mentioned here - https://github.com/portworx/torpedo/blob/master/tests/backup/backup_test.go#L300-L310

tests/backup/backup_rancher_test.go
kphalgun-px, line 1: I am not sure if this is applicable to the changes that have been made, but `CloneAppContextAndTransformWithMappings` does not seem to actually perform a deep copy of the scheduler context; instead, it modifies the namespace field in the context based on the provided namespace mapping. This means, If the function is first used to modify the namespaces to represent restored namespaces, the original app context will subsequently contain these modified namespaces. A subsequent call to `CloneAppContextAndTransformWithMappings` would then need to use these modified, restored namespaces as keys in the namespace mapping, rather than the original ones.

For Example:

1. Original App Context Namespace: original-namespace
2. After first call to `CloneAppContextAndTransformWithMappings`, the namespace is changed to restored-namespace.
3. If calling `CloneAppContextAndTransformWithMappings` again, the namespace mapping should now have restored-namespace as the key to account for the changes made during the first call.

snigdha-px: Since this code is no more used, marking it as resolved

kphalgun-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #1651, opened 2023-09-06
PWX-33543

alicelyu-px:


lsrinivas-pure:
@alicelyu-px can you check this
https://aetos.pwx.purestorage.com/resultSet/testSetID/333815
every test name is coming as StoragePoolExpandDiskResize

PR #1674, opened 2023-09-16
Adding ability to deploy kubevirt based apps

mkoppal-px:

- [x] Vendor-in `kubevirt.io/api v1.0.0-alpha.0`
- [x] Added a kubevirt related API structs to the scheme which will be decoded and validated during parsing of app specs
- [x] Created a function `createVirtualMachineObjects` which will deploy kubevirt VirtualMachine
- [x] Added kubevirt based specs - `kubevirt-simple-vm` (deploys kubevirt VM without PVC) and `kubevirt-vm-pvc` (deploys kubevirt VM backed by Px volume)

Closes #PA-1638 #PA-1637

[Successful jenkins run](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2628/)

Few more points to remember:

1. The creation of Virtual Machines is being done by a `kubectl apply`
2. Ideally we should use native kubevirt client which is vendored-in in sched-ops and we should be calling shced-ops functions to deploy apps
3. Making changes to sched-ops is a bigger effort and will be taken up later
4. Another risk here is, we are using kubevirt api and client version `v1.0.0-alpha.0` but the released version is `v1.0.0`
5. The reason why we are going with an older version of kubevirt is because kubevirt version `v1.0.0` has dependency on 'k8s.io' version `v0.27.1`, so when we bring in kubevirt `v1.0.0` the `k8s.io` version will be bumped to `v0.27.1`. But the other packages like stork, operator, px-backup etc., are not yet bumped to `v0.27.1` and the Torpedo compilation fails. We need to fix this in the coming days.


vilasd (commented):
LGTM

drivers/scheduler/k8s/k8s.go
```diff
@@ -627,9 +631,14 @@ func decodeSpec(specContents []byte) (runtime.Object, error) {
...
 		}
 
+		if err := kubevirtv1.AddToScheme(schemeObj); err != nil {
+			return nil, err
+		}
+
 		codecs := serializer.NewCodecFactory(schemeObj)
 		obj, _, err = codecs.UniversalDeserializer().Decode([]byte(specContents), nil, nil)
 		if err != nil {
+			log.Infof("Error occurred after adding kubevirt scheme")
```
kphalgun-px, line 641: How about we include err in the message?

mkoppal-px: Actually it is not needed. Let me remove it.

drivers/scheduler/k8s/k8s.go
```diff
@@ -4983,6 +5011,32 @@ func (k *K8s) createAdmissionRegistrationObjects(
...
+		time.Sleep(60 * time.Second)
+		specDir = flag.Lookup(SpecDirCliFlag).Value.(flag.Getter).Get().(string)
+		specPath := filepath.Join(specDir, app.Key)
+		if _, err := os.Stat(specPath); baseErrors.Is(err, os.ErrNotExist) {
+			return nil, fmt.Errorf("cannot find yaml in path %s", specPath)
+		}
+		cmdArgs := []string{"apply", "-f", specPath, "-n", ns.Name}
+		err := osutils.Kubectl(cmdArgs)
+		if err != nil {
+			return nil, fmt.Errorf("error applying spec fromt path [%s], Error: %s", specPath, err)
```
kphalgun-px, line 5031: applying spec from*

mkoppal-px: fixed

tests/common.go
kphalgun-px, line 1: How about updating CloneSpec?

kphalgun-px (approved):
LGTM

tests/common.go
mkoppal-px: Fixed it.

drivers/scheduler/k8s/specs/kubevirt-vm-pvc/kubevirt-app.yaml
```diff
@@ -0,0 +1,44 @@
...
+      domain:
+        cpu:
+          cores: 2
+        devices:
+          disks:
+            - disk:
+                bus: virtio
+              name: disk0
+            - cdrom:
+                bus: sata
```
trenukarya-px, line 26: Can we make it SCSI since it is widely used in VM world ?

drivers/scheduler/k8s/specs/kubevirt-simple-vm/kubevirt-app.yaml
```diff
@@ -0,0 +1,39 @@
...
+                bus: virtio
+            - name: cloudinitdisk
+              disk:
+                bus: virtio
+          interfaces:
+            - name: default
+              masquerade: {}
+        resources:
+          requests:
+            memory: 64M
```
trenukarya-px, line 29: Is 64M enough for the VMs ?

drivers/scheduler/k8s/specs/kubevirt-vm-pvc/kubevirt-app.yaml
mkoppal-px: @trenukarya-px These are just the first cut of kubevirt VM specs that we have introduced.
These are just plain vanilla VMs with no apps running in it.
Once @vpinisetti-px creates specs for various types of VMs with various volume types, NICs etc., we will keeo adding it to our app list in Torpedo.
I am waiting for Venkat to test them manually and qualify them as golden templates, then add those specs to Torpedo.
Let me know your thoughts.

drivers/scheduler/k8s/specs/kubevirt-simple-vm/kubevirt-app.yaml
mkoppal-px: This is just a minimal VM with no PVC attached to it and it uses memory as a disk, it is meant for quick testing which ensures that the kubevirt cluster is in a position to provision VMs. Not for real system tests.

trenukarya-px (approved):

drivers/scheduler/k8s/k8s.go
```diff
@@ -4983,6 +5010,36 @@ func (k *K8s) createAdmissionRegistrationObjects(
...
+	ns *corev1.Namespace,
+	app *spec.AppSpec,
+) (interface{}, error) {
+	var specDir string
+	if obj, ok := spec.(*kubevirtv1.VirtualMachine); ok {
+		log.Warnf("Applying VirtualMachine spec for app [%s]", app.Key)
+		// This sleep is added in order to wait for the PVC to get bound.
+		// If this is not provided we are seeing errors like this when VM is starting:
+		// preparing host-disks failed: unable to create /var/run/kubevirt-private/vmi-disks/disk0/disk.img, not enough space
+		time.Sleep(30 * time.Second)
```
kshithijiyer-px, line 5025: Why can't we add a check to make sure pvc is in bout state instead of going with this sleep directly?

lsrinivas-pure (approved):
LGTM

vsundarraj-px (approved):

vpinisetti-px (approved):
LGTM

Rohit-PX:
@mkoppal-px: If we don't use the schedops way it will be difficult to scale the code in the future, can we start with adding just basic support for kubevirt for `create` , `delete` and `validate` instead of a private method in torpedo as we are doing currently?

mkoppal-px:
> @mkoppal-px: If we don't use the schedops way it will be difficult to scale the code in the future, can we start with adding just basic support for kubevirt for `create` , `delete` and `validate` instead of a private method in torpedo as we are doing currently?

You are right @Rohit-PX and we are aware of this. @vsundarraj-px is already working on it as we speak.
This PR is just to get started and is a stop gap arrangement. The createVirtualMachineObjects func will call sched-ops for sure.

PR #1678, opened 2023-09-19
[PA-1658] TC Automation - Delete locked schedule backup and manual backup by non-admin user from admin

kphalgun-px:
What this PR does / why we need it:
This PR adds a test case named "DeleteLockedBucketUserObjectsFromAdmin" to automate the use case of deleting backups, backup schedules, restore and cluster objects created with locked bucket from the admin. This test case will increase the test coverage of Px-Backup TC Automation and ensure that the orphaned object feature functions as intended.

Which issue(s) this PR fixes (optional)
Closes #PA-1658

Special notes for your reviewer:
This test case is applicable to Px-Backup versions 2.6.0 and above. This is a locked bucket scenario and cannot be tested with existing resources.


ak-px (approved):
LGTM

PR #1681, opened 2023-09-20
PA-1630: Add addional logs to dump backup name list

kshithijiyer-px:
- [x] Change check in loop to exit if it fetches more backup then expected. 
- [x] Add additional log statements to dump all the backup names

Dashboard URL : 
https://aetos.pwx.purestorage.com/resultSet/testSetID/350107 
~https://aetos.pwx.purestorage.com/resultSet/testSetID/349427~ 

Closes # PA-1630

### DIFF
--- a/tests/backup/backup_restore_basic_test.go
+++ b/tests/backup/backup_restore_basic_test.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/blang/semver"
 
 	"github.com/portworx/sched-ops/k8s/storage"
+	"github.com/portworx/torpedo/drivers"
 	"github.com/portworx/torpedo/drivers/scheduler/k8s"
 	storageApi "k8s.io/api/storage/v1"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


snigdha-px (approved):
LGTM

mkoppal-px:
Couple of questions:

1. If we are just checking the number of backups to be more than a certain number, won't that we a problem? We may miss out on potential bugs. Is there no way to predict the exact number of synced backup even when we are using a fresh new bucket?
2. Can we add a log statement to mention the expected number of backups with the list of backup names that we are expecting and the list of backup names we get after syncing? Will help in debugging.

kshithijiyer-px:
> Couple of questions:
> 
> 1. If we are just checking the number of backups to be more than a certain number, won't that we a problem? We may miss out on potential bugs. Is there no way to predict the exact number of synced backup even when we are using a fresh new bucket?

If you see this  https://github.com/portworx/torpedo/blob/master/tests/backup/backup_restore_basic_test.go#L2018
 what we have currently done is we are using a custom bucket every time however even with this what we have observed is that once if a blue moon this testcase fails when it see more number of backups than expected. This could be due to various reasons like there are two jobs running at the same time and accidentally using the same bucket or after an older job or test which has some residue it ended reusing it.  Now there is no way of us to predict what would be the neumber of bacckup which will be synced. Will this be a bug, As of yet I don't think it will be a bug as backup is just doing what's expected and just syncing the backups which are present in bucket.   

> 2. Can we add a log statement to mention the expected number of backups with the list of backup names that we are expecting and the list of backup names we get after syncing? Will help in debugging.

This will tell you what is there and what was new which was synced but even with that we won't be able to avoid this scenario.

mkoppal-px (approved):
LGTM

PR #1689, opened 2023-09-22
Moving secret creation before pod deployment

aghodke1312:
Before the PR: https://jenkins.pwx.dev.purestorage.com/job/Users/job/Aniket/job/Mke-all/55/console Torpedo pod would not start saying the secret does not exists. 
With the change: https://jenkins.pwx.dev.purestorage.com/job/Users/job/Aniket/job/Mke-all/58/


mborodin-px (approved):
LGTM

PR #1690, opened 2023-09-24
Added new RBAC TC VerifyRBACforInfraAdmin and helper methods

ak-px:

1. Added new test VerifyRBACforInfraAdmin 
https://portworx.testrail.net/index.php?/cases/view/87886
2. Added Role interface and added new methods.
3. Changed the node from master to worker node where we mount nfs path for deletion. In some setups the master node wont have nfs utils.
4. Added code to not replace exiting px-admin-group ownership in update ownership methods.

Closes #PA-1518
#PA-1672

s3 logs:
https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2697/

NFS logs:
https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2703/


tests/common.go
```diff
@@ -4751,15 +4751,15 @@ func DeleteNfsSubPath(subPath string) {
 	creds := GetNfsInfoFromEnv()
 	mountDir := fmt.Sprintf("/tmp/nfsMount" + RandomString(4))
 
-	// Mount the NFS share to the master node.
-	masterNode := node.GetMasterNodes()[0]
```
ak-px, line 4755: The error seen if master node doesn't have nfs utils. Hence changes the node to worker

{Failed to run command on: ip-10-13-194-139.pwx.purestorage.com. Cause: unable to run cmd (mount -t nfs nfs-pxbkp.pwx.dev.purestorage.com:/mnt/nfs-share-1 /tmp/nfsMounteyml): Failed to run command on: . Cause: failed to run command. sterr: mount: /tmp/nfsMounteyml: bad option; for several filesystems (e.g. nfs, cifs) you might need a /sbin/mount.<type> helper program.

tests/backup/backup_rbac_test.go
```diff
@@ -0,0 +1,501 @@
...
+	api "github.com/portworx/px-backup-api/pkg/apis/v1"
+	"github.com/portworx/torpedo/drivers"
+	"github.com/portworx/torpedo/drivers/backup"
+	"github.com/portworx/torpedo/drivers/scheduler"
+	"github.com/portworx/torpedo/pkg/log"
+	. "github.com/portworx/torpedo/tests"
+)
+
+// VerfiyRBACforInfraAdmin Validates the RBAC opeartion for infra-admin user.
+var _ = Describe("{VerfiyRBACforInfraAdmin}", func() {
```
mkoppal-px, line 20: nit: change `Verfiy` to `Verify` everywhere

ak-px: Done

tests/backup/backup_helper.go
```diff
@@ -3644,6 +3644,7 @@ func GetVolumeMounts(AppContextsMapping *scheduler.Context) ([]string, error) {
 func UpdateBackupLocationOwnership(name string, uid string, userNames []string, groups []string, accessType OwnershipAccessType, publicAccess OwnershipAccessType, ctx context.Context) error {
 	log.Infof("UpdateBackupLocationOwnership for users %v", userNames)
 	backupDriver := Inst().Backup
+	groups = append(groups, "px-admin-group")
```
mkoppal-px, line 3647: Why do we have to do this?

ak-px: By default "px-admin-group" will have ownership to all RBAC resources, and it should not be resetted if are setting the groups list to null or any other groups.

mkoppal-px: Then in this case, you need to get the ownership group from the RBAC object first and then apply the same.
Let's say I have set another group (say `super-group`) along with `px-admin-group` as the owner group, now if we just add `px-admin-group`, we are effectively removing the `super-group` right?
Also, I feel the same has to be done with the users.

drivers/backup/auth.go
```diff
@@ -177,6 +178,13 @@ type KeycloakGroupToUser struct {
 	Realm   string `json:"realm"`
 }
 
+// AddRoleToKeyCloak representation of adding role to keycloak
+type AddRoleToKeyCloak struct {
```
mkoppal-px, line 182: Can we call this just `Role` ?

drivers/backup/auth.go
```diff
@@ -1232,3 +1240,57 @@ func GetRandomUserFromGroup(groupName string) (string, error) {
 	userName := users[rand.Intn(len(users))]
 	return userName, nil
 }
+
+// AddRoleToKeycloak adds new role to keycloak
+func AddRoleToKeycloak(role PxBackupRole, description string) error {
```
mkoppal-px, line 1245: Can we call this `CreateRole`?

drivers/backup/auth.go
```diff
@@ -1232,3 +1240,57 @@ func GetRandomUserFromGroup(groupName string) (string, error) {
...
+	}
+	_, err = processHTTPRequest(method, reqURL, headers, strings.NewReader(string(roleBytes)))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// DeleteRoleFromKeycloak deletes role from keycloak
+func DeleteRoleFromKeycloak(role PxBackupRole) error {
```
mkoppal-px, line 1274: Can we call this `DeleteRole`?

tests/backup/backup_helper.go
ak-px: There is challenge in implementing that approach in api way.
For eg: suppose the rbac object has current ownership as super-group , and now user want to cancel the ownership (nil), This method wont be useful as it will always fetch the current ownership and applies it. 
UI does this workflow by keeping a seperate list of current ownerships(users and groups), And that list is passed to px-backup for applying the ownership while updation of rbac object.

tests/backup/backup_helper.go
```diff
@@ -4427,3 +4430,86 @@ func UpdateCluster(clusterName string, clusterUid string, kubeConfigPath string,
 	}
 	return status, err
 }
+
+// CreateRole creates role with given services and apis in px-backup datastore and also add role to keycloak.
+func CreateRole(roleName backup.PxBackupRole, svcs, apis []string, ctx context.Context) error {
+	err := backup.AddRoleToKeycloak(roleName, "custom-role")
+	if err != nil {
+		return err
+	}
```
mkoppal-px, line 4439: Do we really need to do this?
When you try to create a role from UI we are not providing the role ID. We are just passing the role name along with APIs and Services right?

tests/backup/backup_helper.go
```diff
@@ -4427,3 +4430,86 @@ func UpdateCluster(clusterName string, clusterUid string, kubeConfigPath string,
 	}
 	return status, err
 }
+
+// CreateRole creates role with given services and apis in px-backup datastore and also add role to keycloak.
+func CreateRole(roleName backup.PxBackupRole, svcs, apis []string, ctx context.Context) error {
```
mkoppal-px, line 4435: Can we convert the services and apis into types and accept only those types in this function rather than accepting string? This will make the test case also more readable.

tests/backup/backup_helper.go
ak-px: Yes this is the workflow, If directly made call to px-backup, keycloak will be unware of the newly added role and we wont be able to add role later to user. Actual UI flow remains same, when we create role in UI, first entry is made in keycloak and request is given to px-backup.

drivers/backup/auth.go
ak-px: we already have role struct backup.go , hence changes this into RoleCreate.

tests/backup/backup_helper.go
ak-px: Done.

drivers/backup/auth.go
ak-px: Done.

drivers/backup/auth.go
ak-px: Done.

tests/backup/backup_helper.go
ak-px: As discussed will change this UpdateOwnership to AddOwnership and RemoveOwnership. Since the change requires changes in existing tests where it is used and all rke uses cloudCredentialUpdation. it require more time for sanity check. will take it in seperate ticket https://portworx.atlassian.net/browse/PA-1678

@mkoppal-px

mkoppal-px (approved):
LGTM

snigdha-px (approved):
LGTM

PR #1695, opened 2023-09-25
Adding helper function to delete namespaces stuck in Terminating state

mkoppal-px:
There was no helper function which could be used in case a namespace was in terminating state for a long time after issuing a delete call. We have seen this issue recently in one of our RKE pipelines.
Now `DeleteAppNamespace` function will internally call `DeleteTerminatingNamespace` function if the namespace still exists with `Terminating` status after the namespace delete timeout period.

Closes #

[Successful jenkins run](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2717/console)


tests/backup/backup_helper.go
```diff
@@ -3445,9 +3447,56 @@ func DeleteAppNamespace(namespace string) error {
 		return "", false, nil
 	}
 	_, err = task.DoRetryWithTimeout(namespaceDeleteCheck, namespaceDeleteTimeout, jobDeleteRetryTime)
+	if err != nil {
+		ns, err = k8sCore.GetNamespace(namespace)
+		if err != nil {
+			return err
```
kphalgun-px, line 3453: How about we disregard the error message indicating that the namespace does not exist?

tests/backup/backup_helper.go
```diff
@@ -3445,9 +3447,56 @@ func DeleteAppNamespace(namespace string) error {
...
+	}
+	return nil
+}
+
+// DeleteTerminatingNamespace deletes the given namespace which has been in terminating state
+func DeleteTerminatingNamespace(namespace string) error {
+	k8sCore := core.Instance()
+	ns, err := k8sCore.GetNamespace(namespace)
 	if err != nil {
 		return err
```
kphalgun-px, line 3469: How about we disregard the error message indicating that the namespace does not exist?

tests/backup/backup_helper.go
```diff
@@ -3445,9 +3447,56 @@ func DeleteAppNamespace(namespace string) error {
...
+			return err
+		}
+		namespaceDeleteCheck := func() (interface{}, bool, error) {
+			nsObj, err := core.Instance().GetNamespace(namespace)
+			if err != nil {
+				if strings.Contains(err.Error(), "not found") {
+					log.Infof("Namespace - %s is not found and hence deleted", namespace)
+					return "", false, nil
+				} else {
+					return "", false, err
```
kphalgun-px, line 3485: Is it `return "", true, err`?

tests/backup/backup_helper.go
```diff
@@ -3445,9 +3447,56 @@ func DeleteAppNamespace(namespace string) error {
...
+// DeleteTerminatingNamespace deletes the given namespace which has been in terminating state
+func DeleteTerminatingNamespace(namespace string) error {
+	k8sCore := core.Instance()
+	ns, err := k8sCore.GetNamespace(namespace)
 	if err != nil {
 		return err
 	}
+	if ns.Status.Phase == "Terminating" {
+		log.Infof("Namespace - %s is in %s phase ", namespace, ns.Status.Phase)
+		ns.Spec.Finalizers = nil
```
kphalgun-px, line 3473: How about we log `ns.Spec.Finalizers` before making it nil?

tests/backup/backup_helper.go
mkoppal-px: You are assuming that the only error that can occur is "namespace not found". But if it is failing because of some network connectivity issue, we need to catch that.

tests/backup/backup_helper.go
mkoppal-px: Same as above.

tests/backup/backup_helper.go
kphalgun-px: I mean if err contains "namespace not found" return nil else return err

tests/backup/backup_helper.go
mkoppal-px: I do not want to retry anymore here. If there is some error in getting namespace object and the error is not "namespace not found", then there is something that has gone wrong fundamentally. It is better to bail out with the error than retrying.

tests/backup/backup_helper.go
mkoppal-px: Good point. Will log it.

tests/backup/backup_helper.go
mkoppal-px: If you see, we come to this stage if and only if the namespace has not been deleted even after issuing a normal delete (See the lines above this code). That means the namespace will still be there and because of some reason it is still not deleted. So we will not get "namespace not found" error here.

kphalgun-px: I hope so. I think err returned by DoRetryWithTimeout is not logged.

dbhatnagar-px (approved):
Lgtm

snigdha-px (approved):
LGTM

kshithijiyer-px (approved):
LGTM

PR #1700, opened 2023-09-26
PB-4407: kubevirt Vendor Integration for FC automation with Torpedo

vsundarraj-px:
This is 2nd part of PB-4407 to add vendor for kubevirt support. 
Closes #
PB-4407
This change uses sched-ops changes submitted through PR https://github.com/portworx/sched-ops/pull/329


Makefile
```diff
@@ -123,6 +123,7 @@ vendor-tidy:
 
 vendor:
 	go mod vendor
+	sed -i '1 i\// +build skipcompile\n' vendor/kubevirt.io/client-go/kubecli/kubevirt_test_utils.go
```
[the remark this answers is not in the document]
vsundarraj-px: This is required if any one does make vendor. This will pull the buggy code to vendor list and build will fail.

drivers/scheduler/k8s/k8s.go
```diff
@@ -2419,6 +2429,27 @@ func (k *K8s) createCoreObject(spec interface{}, ns *corev1.Namespace, app *spec
...
+		}
+
+		if err != nil {
+			return nil, &scheduler.ErrFailedToScheduleApp{
+				App:   app,
+				Cause: fmt.Sprintf("Failed to create Virtualachine: %v, Err: %v", obj.Name, err),
+			}
+		}
+		log.Infof("[%v] Created VirtualMachine: %v", app.Key, obj.Name)
+		return vm, nil
```
mkoppal-px, line 2452: I was thinking of creating a separate method called `createVirtualMachineObjects` like this - https://github.com/portworx/torpedo/pull/1674/files#diff-79d9de3d00fb50e35f933299cfb3675d11d694ccb70e38f0f3fac6230e946d5bR5057-R5086

Let me know if this makes sense because Kubevirt VM is not a core k8s object and this function `createCoreObject` is mainly used for standard objects like deployment, pvc, statefulset, service, secret etc.,

Makefile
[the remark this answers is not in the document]
mkoppal-px: Agreed. But this may have an impact on all those pipelines which are using `go mod vendor` instead of `make vendor`

go.mod
```diff
@@ -358,7 +369,7 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.25.1
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.25.1
 	k8s.io/sample-controller => k8s.io/sample-controller v0.25.1
+	kubevirt.io/containerized-data-importer-api v1.57.0-alpha1 => kubevirt.io/containerized-data-importer-api v1.56.1
```
mkoppal-px, line 372: Any reason why we did this?

drivers/scheduler/k8s/k8s.go
vsundarraj-px: stork Rule is also in the same object which is not k8sCore.. Agree its not core, but we dont have any other kubevirt objects as of now. This can probably evolve once we integrate many kubevirt objects.. Instead of VM I think createKubevirtObject() make sense going forward.

go.mod
vsundarraj-px: This is required for dependency mapping with k8s.io/api version we currently support.

mkoppal-px (approved):
LGTM

Rohit-PX (approved):
LGTM.

Rohit-PX:
Please rebase before merging.

PR #1701, opened 2023-09-26
Initial checkin for vcluster suite

dbhatnagar-px:
This PR checks in vcluster support in Torpedo. One example test case is written to create an Nginx app using Torpedo schedulers.


drivers/scheduler/k8s/k8s.go
```diff
@@ -1794,7 +1795,14 @@ func (k *K8s) createStorageObject(spec interface{}, ns *corev1.Namespace, app *s
 				log.Infof("Setting SC %s volumebinding mode to immediate ", obj.Name)
 			}
 		}
-
+		// Change Context only in case of vCluster tests
+		if vcluster.ContextChange {
+			log.Infof("Context Change is set to %v ", vcluster.ContextChange)
+			err := vcluster.SwitchKubeContext(vcluster.UpdatedClusterContext)
```
aghodke1312, line 1801: can be written in 1 line. 
```
if err := ....; err != nil {
 
}
```

drivers/scheduler/k8s/k8s.go
```diff
@@ -1819,6 +1827,14 @@ func (k *K8s) createStorageObject(spec interface{}, ns *corev1.Namespace, app *s
 		sc.Kind = "StorageClass"
 
 		log.Infof("[%v] Created storage class: %v", app.Key, sc.Name)
+		if vcluster.ContextChange {
+			log.Infof("Changing context back to vcluster: %v", vcluster.CurrentClusterContext)
+			err := vcluster.SwitchKubeContext(vcluster.CurrentClusterContext)
+			if err != nil {
+				log.FailOnError(err, fmt.Sprintf("Could not change context to %v", vcluster.UpdatedClusterContext))
```
aghodke1312, line 1834: same as above.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1794,7 +1795,14 @@ func (k *K8s) createStorageObject(spec interface{}, ns *corev1.Namespace, app *s
 				log.Infof("Setting SC %s volumebinding mode to immediate ", obj.Name)
 			}
 		}
-
+		// Change Context only in case of vCluster tests
+		if vcluster.ContextChange {
+			log.Infof("Context Change is set to %v ", vcluster.ContextChange)
+			err := vcluster.SwitchKubeContext(vcluster.UpdatedClusterContext)
+			if err != nil {
+				log.FailOnError(err, fmt.Sprintf("Could not change context to %v", vcluster.UpdatedClusterContext))
```
aghodke1312, line 1803: log.FailOnError is executed only when there is error, no need for if condition.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+	CurrentClusterContext string
+	ContextChange         = false
+)
+
+// This method switches kube context between host and any vcluster
+func SwitchKubeContext(target string) error {
+	cmd := exec.Command("kubectl", "config", "get-contexts", "-o", "name")
+	out, err := cmd.Output()
+	if err != nil {
+		return fmt.Errorf("Failed to fetch kube contexts: %v", err)
```
aghodke1312, line 22: return err directly.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+		}
+	}
+	if desiredContext == "" {
+		return fmt.Errorf("Context for %s not found", target)
+	}
+	log.Infof("Desired Context is : %v", desiredContext)
+	cmd = exec.Command("kubectl", "config", "use-context", desiredContext)
+	_, err = cmd.CombinedOutput()
+	if err != nil {
```
aghodke1312, line 48: convert to 1 line if condition.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+		return fmt.Errorf("Failed to switch to context %s: %v", desiredContext, err)
```
aghodke1312, line 49: return err directly.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	cmd = exec.Command("kubectl", "config", "current-context")
+	out, err = cmd.Output()
+	if err != nil {
+		return fmt.Errorf("Failed to get current context: %v", err)
```
aghodke1312, line 54: return err.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+		return fmt.Errorf("Failed to switch to the desired context: %s", desiredContext)
+	}
+	return nil
+}
+
+// This method deletes a vcluster
+func DeleteVCluster(vclusterName string) error {
+	cmd := exec.Command("vcluster", "delete", vclusterName)
+	err := cmd.Run()
+	if err != nil {
```
aghodke1312, line 66: 1 line

tests/basic/vcluster_test.go
```diff
@@ -0,0 +1,149 @@
...
+				}
+			}
+		})
+	})
+	JustAfterEach(func() {
+		defer EndTorpedoTest()
+		for _, name := range vclusterNames {
+			err := vcluster.DeleteVCluster(name)
+			if err != nil {
+				log.Errorf("Error deleting Vcluster with name %v", name)
```
aghodke1312, line 51: Return error directly. In this case your error would be: 
Error deleting Vcluster with name name <name> 
But your DeleteVCluster is already returning: Failed to delete vcluster <name>

aghodke1312 (changes requested):
There are a lot of return fmt.Errorf() statements, please relpace them with return err, since the leaf node in call stack returns error string already. 
Also change all the if condition to 1 liners wherever they can be used.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+				desiredContext = ctx
+				break
+			}
+		}
+	}
+	if desiredContext == "" {
+		return fmt.Errorf("Context for %s not found", target)
+	}
+	log.Infof("Desired Context is : %v", desiredContext)
+	cmd = exec.Command("kubectl", "config", "use-context", desiredContext)
```
nikolaypopov, line 46: Same here, can we use k8s API instead?

tests/basic/vcluster_test.go
```diff
@@ -0,0 +1,149 @@
...
+	// Running the vcluster connect in the background
+	go func() {
+		cmd := exec.Command("vcluster", "connect", vclusterName)
+		cmd.Start()
+		<-killChan
+		cmd.Process.Signal(os.Interrupt)
+		cmd.Wait()
+	}()
+	isConnected := false
+	for i := 0; i < 60; i++ {
```
Santhoshairody, line 69: Can you parameterise 60 , so that caller can pass the required time

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("Failed to delete vcluster %v", vclusterName)
+	}
+	return nil
+}
+
+// This method creates a vcluster. This requires vcluster.yaml saved in a specific location.
+func CreateVCluster(vclusterName string) error {
+	cmd := exec.Command("vcluster", "create", vclusterName, "-f", "/Users/dbhatnagar/PxOne/vcluster/torpedo/vcluster.yaml", "--connect=false")
```
Santhoshairody, line 74: Pass it as a parameter. it should be able to take it from caller

Santhoshairody (commented):
Can we use dash.VerifyFatal , dash.VerifySafely wherever required.

drivers/scheduler/k8s/k8s.go
dbhatnagar-px: Moved to log.FailOnError()

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+
+var (
+	UpdatedClusterContext string
+	CurrentClusterContext string
+	ContextChange         = false
+)
+
+// This method switches kube context between host and any vcluster
+func SwitchKubeContext(target string) error {
+	cmd := exec.Command("kubectl", "config", "get-contexts", "-o", "name")
```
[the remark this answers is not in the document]
dbhatnagar-px: Hi @nikolaypopov , thanks for this suggestion. I was also exploring this on similar lines. The problem here is - there is no direct API endpoint exposed for kubectl config cli, as this works directly with kubeconfig files. So from endpoint perspective, we need to provide path to kubeconfig and then the API would read that file and use that config. However, here in vcluster context, kubeconfig is not static as these vclusters are ephemeral in nature and our test cases will be creating and destroying them on need basis. We can only get kubeconfig out of a vcluster if we have a separate IP/hostname for that vcluster and through load-balancer we are trying to connect over it. Since we are not going that route, and simply using port-forwarding mechanism of vclusters to connect we won't get kubeconfigs here. 

This is what I researched and then decided to go ahead with this way. Please let me know if you are able to figure out any other way for this usecase. I can implement that otherwise I believe this is the only way as of now.

drivers/vcluster/vcluster.go
dbhatnagar-px: Same answer as above. I don't think we can use api endpoint here for our particular use case. Please let me know if you are able to explore a way out.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,97 @@
...
+		if err != nil {
+			return err
+		}
+		if strings.Contains(string(output), vclusterName) && strings.Contains(string(output), "Running") {
+			return nil
+		}
+		time.Sleep(10 * time.Second)
+	}
+	return fmt.Errorf("vcluster %s did not reach Running status within the timeout", vclusterName)
+}
```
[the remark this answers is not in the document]
dbhatnagar-px: Done.

drivers/scheduler/k8s/k8s.go
```diff
@@ -1794,7 +1795,12 @@ func (k *K8s) createStorageObject(spec interface{}, ns *corev1.Namespace, app *s
 				log.Infof("Setting SC %s volumebinding mode to immediate ", obj.Name)
 			}
 		}
-
+		// Change Context only in case of vCluster tests
+		if vcluster.ContextChange {
+			log.Infof("Changing context to %v ", vcluster.UpdatedClusterContext)
+			err := vcluster.SwitchKubeContext(vcluster.UpdatedClusterContext)
+			log.FailOnError(err, fmt.Sprintf("Could not change context to %v", vcluster.UpdatedClusterContext))
```
lsrinivas-pure, line 1802: can we just return the err similar to other conditions in the below lines. let the test handle the error

tests/basic/vcluster_test.go
```diff
@@ -0,0 +1,139 @@
+package tests
+
+import (
+	"fmt"
+	. "github.com/onsi/ginkgo"
```
aghodke1312, line 5: sort imports.

drivers/vcluster/vcluster.go
```diff
@@ -0,0 +1,105 @@
+package vcluster
+
+import (
+	"fmt"
+	"github.com/portworx/sched-ops/task"
```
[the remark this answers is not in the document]
dbhatnagar-px: Goland itself is sorting it in this manner in this file (alphabetically).

lsrinivas-pure (approved):
LGTM

aghodke1312 (approved):

PR #1708, opened 2023-09-27
Removed the logic of appending px-admin-group in update ownership

mkoppal-px:
Reverting partial code from https://github.com/portworx/torpedo/pull/1690
If we append `px-admin-group` to the `groups` in any update ownership function and we send the access type as invalid, the `px-admin-group` which has full access to it may lose the access and it is also resulting in the following error

```
failed to share the cloud cred with error failed to update CloudCredential ownership : rpc error: code = Unknown desc = group[4a3d3297-052d-4e37-9aa3-5b06e5aee857] has invalid access type[Invalid]
```

Closes #


kphalgun-px (approved):
LGTM

ak-px (approved):
LGTM

snigdha-px (approved):
lgtm

dbhatnagar-px (approved):
Lgtm

PR #1712, opened 2023-09-28
fix for zones in gke

lsrinivas-pure:
With recent change in spawn to use only one zone for torpedo pool for GKE, number of zones giving incorrect value to GetZones in torpedo. This PR fixes that.

Closes #PTX-20378


dbhatnagar-px (approved):
Lgtm

PR #1720, opened 2023-10-03
Adding few more Gin APIs

dbhatnagar-px:
This PR adds few more GIn APIs:
1. Fetches and Parses Pxctl status output
2. Validates if Px is installed or not
3. Fetches Px Version


lsrinivas-pure (approved):
LGTM

apiServer/pxone/utils/torpedoutils.go
```diff
@@ -177,3 +178,110 @@ func ScheduleAppsAndValidate(c *gin.Context) {
 		})
 	}
 }
+
+// GetPxVersion This function returns the current Px Version in the Target Cluster
+func GetPxVersion(c *gin.Context) {
+	if !checkTorpedoInit(c) {
```
mkoppal-px, line 184: What if there is an error in the function `checkTorpedoInit`?
Also how does gin handle when there is nothing returned? As in, here we are not returning either `StatusOK` or `StatusInternalServerError` ?

dbhatnagar-px: So, checkTorpedoInit() is InitInstance(). I was also having similar doubt but nowhere in our other test code we are checking for any errors in Initinstance(). That's the reason this method is also similar. 

Regarding not returning anything here, you are right. This seems to be a miss from my end. I should be adding a StatusInternalServerError to gin Context and then doing a null return. I will do this - thanks for catching this.

dbhatnagar-px: Done. Thanks for catching this.

mkoppal-px (approved):
LGTM

PR #1725, opened 2023-10-04
PA-1719:  Fix upgrade case to run with IKS + PX

kshithijiyer-px:
- [x] Add refresh for ibm driver.  
- [x] Add master node check in `CollectLogsFromPods()` to not attempt to collect pods logs in case of clouds. 
- [x] Fix remove icr  custom registry to allow to upgrade to stork version not pushed to icr.   

Job link where testcase has passed: https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/upgrade-tests/job/Px-Backup-with-Stork-Upgrade/job/Px-Backup+Stork-Upgrade-iks-px/23/

Closes #PA-1719


tests/backup/backup_helper.go
```diff
@@ -2672,6 +2672,13 @@ func upgradeStorkVersion(storkImageToUpgrade string) error {
 			return err
 		}
 		storageSpec.Spec.Stork.Image = finalImageToUpgrade
+
+		// Check to reset customImageRegistry to blank as in case of ibm it'll be icr.io/ext/ and not
+		// docker.io/ which causes issues when we try to install stork which is not pushed to icr.io/ext
+		if os.Getenv("CLUSTER_PROVIDER") == "ibm" {
```
kphalgun-px, line 2678: Is it possible to use GetClusterProviders()?

tests/backup/backup_helper.go
```diff
@@ -2672,6 +2672,13 @@ func upgradeStorkVersion(storkImageToUpgrade string) error {
...
+			storageSpec.Spec.CustomImageRegistry = ""
```
kphalgun-px, line 2679: How about we log the value of storageSpec.Spec.CustomImageRegistry before making it empty?

kshithijiyer-px: How will that be helpful?

kphalgun-px (approved):
LGTM

tests/backup/backup_helper.go
kphalgun-px: I am not sure, but for debugging purposes, logging the original value of storageSpec.Spec.CustomImageRegistry before overriding it could be helpful to verify it was not already empty.

tests/backup/backup_helper.go
kshithijiyer-px: Done

tests/backup/backup_helper.go
kshithijiyer-px: When you deploy from IBM catalog it will always have the value and hence it won't be empty. Hence this case won't occur.

mkoppal-px (approved):
LGTM
@lsrinivas-pure Can you please review this since it involves changes to `ibm.go` and you are using it in your pipelines

tests/common.go
```diff
@@ -6040,6 +6046,13 @@ func collectAndCopyDiagsOnWorkerNodes(issueKey string) {
 
 // CollectLogsFromPods collects logs from specified pods and stores them in a directory named after the test case
 func CollectLogsFromPods(testCaseName string, podLabel map[string]string, namespace string, logLabel string) {
+
+	// Check to handle cloud based deployment with 0 master nodes
+	if len(node.GetMasterNodes()) == 0 {
```
lsrinivas-pure, line 6051: is this failing for only IBM or all the clouds?

kshithijiyer-px: This will fail for all clouds.

lsrinivas-pure (approved):
LGTM

PR #1740, opened 2023-10-10
Added new resiliency test

Madanagopal19:
New Resiliency Test - RestoreDuringNodesAreRebooted
This tests below scenarios

1. Trigger restore during all worker nodes are getting rebooted
2. Trigger restore during all ds nodes are getting rebooted (ds not in quorum)
3. Trigger restore and reboot the node on which restored ds is hosted

Jenkins logs - https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/582/console


aravindranath14 (approved):
LGM

PR #1743, opened 2023-10-10
Kubevirt app deployment fixes

mkoppal-px:

- [x] Created a separate function `createVirtualMachineObjects` to create Kubevirt VM
- [x] Added Virtual Machine details to be printed when `Describe` function is called
- [x] Added specs for `kubevirt-simple-vm` which creates a Kubevirt VM without any volumes. This creates volume in-memory
- [x] Added specs for `kubevirt-vm-pvc` which creates a Kubevirt VM with PVC as volume source
- [x] Made a small change in `validateMountsInPods`- In some cases the container path is sym linked to a local directory in the pod "/proc/mounts" contains this sym link path instead of the actual container path fetched. To make sure it doesn't fail in validation, we will be appending the sym link path to "paths" and removing the container path
- [x] Added Kubevirt VM type in `GetSpecNameKindNamepace`, `CloneSpec` and `UpdateNamespace`

Closes #PA-1638 #PA-1637

[Jenkins Run with simple VM](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2921)
[Jenkins Run with Px PVC based VM](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/2919)

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -5176,7 +5176,7 @@ func (k *K8s) createVirtualMachineObjects(
 						}
 						return "", true, fmt.Errorf("waiting for import to be completed for pvc [%s] in namespace [%s] for virtual machine [%s]", pvcName, ns.Name, obj.Name)
 					}
-					_, err := task.DoRetryWithTimeout(t, 5*time.Minute, 5*time.Second)
+					_, err := task.DoRetryWithTimeout(t, 5*time.Minute, 30*time.Second)
 					if err != nil {
 						return nil, err
 					}
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -5211,7 +5211,7 @@ func isPVCType(source kubevirtv1.VolumeSource) bool {
 		log.Infof("Field - %s", t.Field(i).Name)
 		fieldType := t.Field(i)
 
-		if fieldType.Type == reflect.TypeOf(kubevirtv1.PersistentVolumeClaimVolumeSource{}) {
+		if fieldType.Type == reflect.TypeOf(&kubevirtv1.PersistentVolumeClaimVolumeSource{}) {
 			log.Infof("returning true")
 			return true
 		}
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -1229,6 +1229,22 @@ func (k *K8s) CreateSpecObjects(app *spec.AppSpec, namespace string, options sch
 		}
 	}
 
+	for _, appSpec := range app.SpecList {
+		t := func() (interface{}, bool, error) {
+			obj, err := k.createVirtualMachineObjects(appSpec, ns, app)
+			if err != nil {
+				return nil, true, err
+			}
+			return obj, false, nil
+		}
+		obj, err := task.DoRetryWithTimeout(t, k8sObjectCreateTimeout, DefaultRetryInterval)
+		if err != nil {
+			return nil, err
+		}
+		if obj != nil {
+			specObjects = append(specObjects, obj)
+		}
+	}
 	return specObjects, nil
 }
 
@@ -2485,27 +2501,6 @@ func (k *K8s) createCoreObject(spec interface{}, ns *corev1.Namespace, app *spec
 		}
 		log.Infof("[%v] Created Rule: %v", app.Key, rule.GetName())
 		return rule, nil
-	} else if obj, ok := spec.(*kubevirtv1.VirtualMachine); ok {
-		// Create VirtualMachine Spec
-		if obj.Namespace != "kube-system" {
-			obj.Namespace = ns.Name
-		}
-		vm, err := k8sKubevirt.CreateVirtualMachine(obj)
-		if k8serrors.IsAlreadyExists(err) {
-			if vm, err = k8sKubevirt.GetVirtualMachine(obj.Name, obj.Namespace); err == nil {
-				log.Infof("[%v] Found existing VirtualMachine: %v", app.Key, obj.Name)
-				return vm, nil
-			}
-		}
-
-		if err != nil {
-			return nil, &scheduler.ErrFailedToScheduleApp{
-				App:   app,
-				Cause: fmt.Sprintf("Failed to create Virtualachine: %v, Err: %v", obj.Name, err),
-			}
-		}
-		log.Infof("[%v] Created VirtualMachine: %v", app.Key, obj.Name)
-		return vm, nil
 	} else if obj, ok := spec.(*corev1.Pod); ok {
 		obj.Namespace = ns.Name
 		if options.Scheduler != "" {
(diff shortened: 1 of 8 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -4879,6 +4876,19 @@ func (k *K8s) Describe(ctx *scheduler.Context) (string, error) {
...
+			buf.WriteString(insertLineBreak(fmt.Sprintf("VirtualMachine: [%s] %s", obj.Namespace, obj.Name)))
+			var virtualMachine *kubevirtv1.VirtualMachine
+			if virtualMachine, err = k8sKubevirt.GetVirtualMachine(obj.Name, obj.Namespace); err != nil {
+				buf.WriteString(fmt.Sprintf("%v", &scheduler.ErrFailedToGetCustomSpec{
+					Name:  obj.Name,
+					Cause: fmt.Sprintf("Failed to get VirtualMachine: %v. Err: %v", obj.Name, err),
+					Type:  obj,
+				}))
+			}
+			buf.WriteString(fmt.Sprintf("%v\n", virtualMachine))
```
kphalgun-px, line 4889: How about %+v?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5134,6 +5144,73 @@ func (k *K8s) createAdmissionRegistrationObjects(
...
+							return "", false, err
+						}
+						log.Infof("Events for pvc [%s] in namespace [%s] for virtual machine [%s] \n%v\n", pvcName, ns.Name, obj.Name, events)
+						for _, event := range events.Items {
+							if strings.Contains(event.Message, cdiImportSuccessEvent) {
+								pvc, err := k8sCore.GetPersistentVolumeClaim(pvcName, ns.Name)
+								if err != nil {
+									return "", false, err
+								}
+								log.Infof("cdi.kubevirt.io/storage.condition.running.message status - [%s]", pvc.Annotations[cdiPvcRunningMessageAnnotationKey])
```
kphalgun-px, line 5175: How about %s, cdiPvcRunningMessageAnnotationKey?

drivers/scheduler/k8s/k8s.go
```diff
@@ -5134,6 +5144,73 @@ func (k *K8s) createAdmissionRegistrationObjects(
...
+		// Create VirtualMachine Spec
+		if obj.Namespace != "kube-system" {
+			obj.Namespace = ns.Name
+		}
+		vm, err := k8sKubevirt.CreateVirtualMachine(obj)
+		if k8serrors.IsAlreadyExists(err) {
+			if vm, err = k8sKubevirt.GetVirtualMachine(obj.Name, obj.Namespace); err == nil {
+				log.Infof("[%v] Found existing VirtualMachine: %v", app.Key, obj.Name)
+				return vm, nil
+			}
```
kphalgun-px, line 5198: How about else return nil, err?

kphalgun-px (approved):
LGTM

drivers/scheduler/k8s/k8s.go
```diff
@@ -5134,6 +5144,73 @@ func (k *K8s) createAdmissionRegistrationObjects(
...
+	ns *corev1.Namespace,
+	app *spec.AppSpec,
+) (interface{}, error) {
+	if obj, ok := spec.(*kubevirtv1.VirtualMachine); ok {
+		// Validating to make sure the desired images are imported by CDI and volumes are ready to be used by Kubevirt VM
+		virtualMachineVolumes := obj.Spec.Template.Spec.Volumes
+		if len(virtualMachineVolumes) > 0 {
+			for _, v := range virtualMachineVolumes {
+				// Validating Volume Source of type PersistentVolumeClaim is ready to be used by Kubevirt VM
+				if v.VolumeSource.PersistentVolumeClaim != nil {
```
dbhatnagar-px, line 5159: Does it make sense to make this a separate method - for better and future usability

drivers/scheduler/k8s/k8s.go
```diff
@@ -5134,6 +5144,73 @@ func (k *K8s) createAdmissionRegistrationObjects(
...
+					pvcName := v.VolumeSource.PersistentVolumeClaim.ClaimName
+					t := func() (interface{}, bool, error) {
+						events, err := k8sCore.ListEvents(ns.Name, metav1.ListOptions{
+							FieldSelector: fmt.Sprintf("involvedObject.kind=PersistentVolumeClaim,involvedObject.name=%s", pvcName),
+						})
+						if err != nil {
+							return "", false, err
+						}
+						log.Infof("Events for pvc [%s] in namespace [%s] for virtual machine [%s] \n%v\n", pvcName, ns.Name, obj.Name, events)
+						for _, event := range events.Items {
```
dbhatnagar-px, line 5169: I believe this block can be a separate method as well. If you believe this may be required by other future tests, then it would make sense to do so.

drivers/scheduler/k8s/k8s.go
mkoppal-px: I just tried to follow the convention based on how other have done. For your happiness I have added it. 😄

drivers/scheduler/k8s/k8s.go
mkoppal-px: Done

drivers/scheduler/k8s/k8s.go
mkoppal-px: We have to return the VM object if there is already one existing

drivers/scheduler/k8s/k8s.go
mkoppal-px: I agree. I will convert this whole block to a function called `WaitForImageImportForVM`

drivers/scheduler/k8s/k8s.go
mkoppal-px: We can but the function will need a lot of input parameters like FieldSelector, namespace, object name, key string etc.,
So I felt the function will become lengthy.

drivers/scheduler/k8s/k8s.go
kphalgun-px: err from k8sKubevirt.GetVirtualMachine(obj.Name, obj.Namespace)

drivers/scheduler/k8s/k8s.go
kphalgun-px: My suggestion was based on the convention as well. XD

dbhatnagar-px (approved):
Lgtm

kshithijiyer-px (approved):
LGTM

trenukarya-px (approved):

drivers/scheduler/k8s/k8s.go
```diff
@@ -5134,6 +5144,81 @@ func (k *K8s) createAdmissionRegistrationObjects(
...
+			}
+		}
+		log.Infof("[%v] Created VirtualMachine: %v", app.Key, obj.Name)
+		return vm, nil
+	}
+
+	return nil, nil
+}
+
+func (k *K8s) WaitForImageImportForVM(vmName string, namespace string, v kubevirtv1.Volume) error {
```
vpinisetti-px, line 5189: This import pod will be created only incase PVC has below annotation. Meaning for plane PVC's ` cdiImportSuccessEvent` will not be generated. Need to handle this case.

vpinisetti-px (changes requested):

drivers/scheduler/k8s/k8s.go
mkoppal-px: Fixed it

vpinisetti-px (approved):
LGTM

PR #1746, opened 2023-10-11
funcs to fetch details from cluster provision status

vprabhakar-px:
changes to provision staus functions to pick details on the pool


tests/common.go
```diff
@@ -9198,6 +9191,38 @@ func GetClusterProvisionStatus() ([]ProvisionStatus, error) {
...
+	// Using Node which is up and running
+	var selectedNode []node.Node
+	for _, eachNode := range node.GetStorageDriverNodes() {
+		status, err := IsPxRunningOnNode(&eachNode)
+		if err != nil {
+			log.InfoD("Px is not running on the Node.. searching for other node")
+			continue
+		}
+		if status {
+			selectedNode = append(selectedNode, eachNode)
```
lsrinivas-pure, line 9209: can we select one node and break this loop? any particular reason you want all nodes and then select random node?

tests/common.go
```diff
@@ -9295,6 +9322,39 @@ func WaitForSnapShotToReady(snapshotScheduleName, snapshotName, appNamespace str
 
 }
 
+// WaitTillPoolExpanded waits till pool expansion gets completed
+func WaitTillPoolExpanded(poolUUID string, poolCurrSize uint64) error {
```
lsrinivas-pure, line 9326: How are you making pool expansion is completed here? Sometimes pool size changes but rebalance might be in-progress

tests/common.go
vprabhakar-px: oh !! I thought I was breaking it , let me reupdate it similar to above ones

vprabhakar-px: the intension here is to pick up random node , for specific node we already have a function GetClusterProvisionStatusOnSpecificNode which can be used

tests/common.go
vprabhakar-px: yes, you are right , I missed that part , will update it . currently I am checking only till pool size got increased

vprabhakar-px: removing this function as we already have waitForPoolToBeResized

tests/common.go
```diff
@@ -9200,6 +9193,39 @@ func GetClusterProvisionStatus() ([]ProvisionStatus, error) {
...
+	// Select Random Volumes for pool Expand
+	randomIndex := rand.Intn(len(selectedNode))
+	randomNode := selectedNode[randomIndex]
+
+	clusterProvision, err := GetClusterProvisionStatusOnSpecificNode(randomNode)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Infof("Cluster provision status [%v]", clusterProvision)
```
lsrinivas-pure, line 9228: this is getting logged twice once here and once in  GetClusterProvisionStatusOnSpecificNode

lsrinivas-pure (approved):
LGTM , one minor comment

PR #1747, opened 2023-10-12
handle KubeVirt VM spec in k8s.GetVolumes() method

pureneelesh:
KubeVirt VM specs don't have PVCs. So, GetVolumes() was not finding any volumes for the VM.

This patch looks for the PVCs with VM as owner in the same namespace.

Closes #

No testing done yet. Will merge this only after testing this with new KubeVirt tests in Stork repo.

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -4124,6 +4124,31 @@ func (k *K8s) DeleteVolumes(ctx *scheduler.Context, options *scheduler.VolumeOpt
 	return vols, nil
 }
 
+func (k *K8s) appendVolForPVC(vols []*volume.Volume, pvc *v1.PersistentVolumeClaim) ([]*volume.Volume, error) {
+	shouldAdd, err := k.filterPureVolumesIfEnabled(pvc)
+	if err != nil {
+		return nil, err
+	}
+	if !shouldAdd {
+		return vols, nil
+	}
+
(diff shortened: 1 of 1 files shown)


pp511 (approved):
LGTM

drivers/scheduler/k8s/k8s.go
```diff
@@ -4112,6 +4112,31 @@ func (k *K8s) DeleteVolumes(ctx *scheduler.Context, options *scheduler.VolumeOpt
...
+func (k *K8s) appendVolForPVC(vols []*volume.Volume, pvc *v1.PersistentVolumeClaim) ([]*volume.Volume, error) {
+	shouldAdd, err := k.filterPureVolumesIfEnabled(pvc)
+	if err != nil {
+		return nil, err
+	}
+	if !shouldAdd {
+		return vols, nil
+	}
+
+	pvcSizeObj := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
```
Rohit-PX, line 4124: Add a check if `pvcSizeObj` is nil, else the next line might throw a nil pointer exception.

pureneelesh: pvcSizeObj is of type resource.Quantity which is a struct. So, should be ok?

```
type Quantity struct {
    // i is the quantity in int64 scaled form, if d.Dec == nil
    i int64Amount
 ...
}
```

PR #1754, opened 2023-10-13
PA-1712: Add test to check if  kube and portworx namespaces skipped during all namespace backup

kshithijiyer-px:
- [x] Add test to check if  kube-system and portworx namespace skipped during all namespace backup
- [x] Add helper function `checkBackupObjectForNonExpectedNS()`
- [x] Add helper function `getNamespaceAge()`
- [x] Add helper function `checkStatusOfNamespaces()`

The testcase failed due to: https://portworx.atlassian.net/browse/PB-4609

Closes #PA-1712


tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,327 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
 		CleanupCloudSettingsAndClusters(backupLocationMap, s3CloudCredName, s3CloudCredUID, ctx)
 	})
 })
+
+// checkBackupObjectForNonExpectedNS checks if namespaces like kube-system and px namespace
+// is backed up or not
+func checkBackupObjectForNonExpectedNS(ctx context.Context, backupName string) {
```
kphalgun-px, line 3719: I feel we should take services as input rather than switching clusters internally.

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,327 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+		for _, namespacetoskip := range namespacesToSkip {
+			dash.VerifyFatal(namespacetoskip == namespace, false, fmt.Sprintf("Expected namespace %s shouldn't be present in backup", namespace))
+		}
+	}
+	err = SetSourceKubeConfig()
+	log.FailOnError(err, "Switching context to source cluster failed")
+}
+
+// getNamespaceAge gets the namespace age of all the namespaces on the cluster
+func getNamespaceAge() map[string]time.Time {
```
kphalgun-px, line 3755: I feel we should take namespaces as input rather than switching clusters internally.

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,327 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+	for _, namespace := range allNamespaces.Items {
+		namespaceAge[namespace.ObjectMeta.GetName()] = namespace.ObjectMeta.GetCreationTimestamp().Time
+	}
+	err = SetSourceKubeConfig()
+	log.FailOnError(err, "Switching context to source cluster failed")
+	return namespaceAge
+}
+
+// checkStatusOfNamespaces checks the status of namespaces on clusters where the restore was done
+func checkStatusOfNamespaces(oldNamespaceAge map[string]time.Time) {
```
kphalgun-px, line 3770: I feel we should take services, namespaces as input rather than switching clusters internally.

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,327 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+		}
+	}
+
+	allNamespaces, err := k8sCore.ListNamespaces(make(map[string]string))
+	for _, namespace := range allNamespaces.Items {
+		for _, skipcase := range namespacesToSkip {
+			if skipcase == namespace.GetName() {
+				dash.VerifyFatal(namespaceNamesAge[namespace.GetName()] == oldNamespaceAge[namespace.GetName()], false, fmt.Sprintf("Namespace [%s] was restored", skipcase))
+			} else {
+				dash.VerifyFatal(namespaceNamesAge[namespace.GetName()].After(oldNamespaceAge[namespace.GetName()]), true, fmt.Sprintf("Namespace[%s] not restored", namespace.GetName()))
```
kphalgun-px, line 3791: How about we return error?

tests/backup/backup_restore_basic_test.go
kphalgun-px: How about we return error?

tests/backup/backup_restore_basic_test.go
kphalgun-px: How about we return error?

tests/backup/backup_restore_basic_test.go
kphalgun-px: How about we return error?

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,327 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+					appName := Inst().AppList[index]
+					ctx.ReadinessTimeout = appReadinessTimeout
+					namespace := GetAppNamespace(ctx, taskName)
+					log.InfoD("Scheduled application [%s] in destination cluster in namespace [%s]", appName, namespace)
+					appNamespaces = append(appNamespaces, namespace)
+					scheduledAppContexts = append(scheduledAppContexts, ctx)
+				}
+			}
+			err = SetSourceKubeConfig()
+			log.FailOnError(err, "Switching context to source cluster failed")
```
kphalgun-px, line 3852: This might be redundant.

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,327 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+
+			dstClusterStatus, err := Inst().Backup.GetClusterStatus(orgID, destinationClusterName, ctx)
+			log.FailOnError(err, fmt.Sprintf("Fetching [%s] cluster status", destinationClusterName))
+			dash.VerifyFatal(dstClusterStatus, api.ClusterInfo_StatusInfo_Online, fmt.Sprintf("Verifying if [%s] cluster is online", destinationClusterName))
+
+			dstClusterUid, err := Inst().Backup.GetClusterUID(ctx, orgID, destinationClusterName)
+			log.FailOnError(err, fmt.Sprintf("Fetching [%s] cluster uid", destinationClusterName))
+			log.Infof("Cluster [%s] uid: [%s]", destinationClusterName, dstClusterUid)
+		})
+		Step("Create a schedule policy", func() {
```
kphalgun-px, line 3907: log.InfoD is missing

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,327 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+			log.Infof("Schedule policy [%s] uid: [%s]", schedulePolicyName, schedulePolicyUid)
+		})
+
+		Step("Create a manual backup", func() {
+			log.InfoD("Creating a manual backup")
+			backupName := fmt.Sprintf("%s-all-%v", BackupNamePrefix, time.Now().Unix())
+
+			namespaces := []string{"*"}
+			labelSelectors := make(map[string]string)
+			sourceClusterUid, err := Inst().Backup.GetClusterUID(ctx, orgID, destinationClusterName)
```
kphalgun-px, line 3927: how about destinationClusterUid?

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Check line 3909

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: The intend of these functions is to reduce code duplication in case of the specific case, returning error and again adding dash.verifyfatel will come back to the same problem of redundant code. Yes it would be 3-4 redundant lines then some 10-20 lines but I feel this way it's much more cleaner.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: The intend of these functions is to reduce code duplication in case of the specific case, returning error and again adding dash.verifyfatel will come back to the same problem of redundant code. Yes it would be 3-4 redundant lines then some 10-20 lines but I feel this way it's much more cleaner.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: The intend of these functions is to reduce code duplication in case of the specific case, returning error and again adding dash.verifyfatel will come back to the same problem of redundant code. Yes it would be 3-4 redundant lines then some 10-20 lines but I feel this way it's much more cleaner.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: I have just added it as a fail safe to avoid subsequent cases to fail due to switched kubeconfigs.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Thanks fixing it.

tests/backup/backup_restore_basic_test.go
kphalgun-px: Makes sense but AfterSuite will take care of it, right?

kphalgun-px (approved):
LGTM

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: if we add it to the ValidateBackup func the thing this logic is only needed in case of all namespace backup and an in-place restore of all the namespaces with replace set to true. It will just make it slow for us to use the same.

tests/backup/backup_restore_basic_test.go
kshithijiyer-px: Done

tests/backup/backup_restore_basic_test.go
kphalgun-px: Better not because if we switch in ValidateBackup many tests will fail. Thanks.

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,324 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
 		CleanupCloudSettingsAndClusters(backupLocationMap, s3CloudCredName, s3CloudCredUID, ctx)
 	})
 })
+
+// checkBackupObjectForNonExpectedNS checks if namespaces like kube-system and px namespace
+// is backed up or not
+func checkBackupObjectForNonExpectedNS(ctx context.Context, backupName string) {
```
mkoppal-px, line 3719: Return error here

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,324 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+		for _, namespacetoskip := range namespacesToSkip {
+			dash.VerifyFatal(namespacetoskip == namespace, false, fmt.Sprintf("Expected namespace %s shouldn't be present in backup", namespace))
+		}
+	}
+	err = SetSourceKubeConfig()
+	log.FailOnError(err, "Switching context to source cluster failed")
+}
+
+// getNamespaceAge gets the namespace age of all the namespaces on the cluster
+func getNamespaceAge() map[string]time.Time {
```
mkoppal-px, line 3755: This function is actually helpful in general. Can you please move this to backup_helper or common.go?
Make sure we accept namespace input and return the time with error.

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,324 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+	for _, namespace := range allNamespaces.Items {
+		namespaceAge[namespace.ObjectMeta.GetName()] = namespace.ObjectMeta.GetCreationTimestamp().Time
+	}
+	err = SetSourceKubeConfig()
+	log.FailOnError(err, "Switching context to source cluster failed")
+	return namespaceAge
+}
+
+// checkStatusOfNamespaces checks the status of namespaces on clusters where the restore was done
+func checkStatusOfNamespaces(oldNamespaceAge map[string]time.Time) {
```
mkoppal-px, line 3770: Same here. Move to helper and return error

tests/backup/backup_restore_basic_test.go
```diff
@@ -3712,3 +3713,324 @@ var _ = Describe("{DeleteS3ScheduleAndCreateNfsSchedule}", func() {
...
+		ctx, err = backup.GetAdminCtxFromSecret()
+		log.FailOnError(err, "Fetching px-central-admin ctx")
+
+		numDeployments = Inst().GlobalScaleFactor
+		if len(Inst().AppList) == 1 && numDeployments < 3 {
+			numDeployments = 3
+		}
+	})
+
+	It("Verify if Kube-system and Px Namespace is skipped on all namespace backup", func() {
```
mkoppal-px, line 3833: Now it is not just `kube-system` right?

tests/backup/backup_restore_basic_test.go
mkoppal-px, line 3770: func `CompareNamespaceAge`

mkoppal-px (approved):
LGTM. As discussed we will be running this test once the fix is in px-backup

sn-px (approved):
LGTM

PR #1760, opened 2023-10-16
Enabling Vcluster NodePort framework and one FIO Testcase

dbhatnagar-px:
This PR enables NodePort changes to vcluster framework and one FIO testcase for these changes. These changes will be what we will be using going further.


drivers/vcluster/vcluster.go
```diff
@@ -76,9 +126,19 @@ func DeleteVCluster(vclusterName string) error {
...
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+	return nil
+}
+
 // CreateVCluster This method creates a vcluster. This requires vcluster.yaml saved in a specific location.
 func CreateVCluster(vclusterName string, absPath string) error {
 	cmd := exec.Command("vcluster", "create", vclusterName, "-f", absPath, "--connect=false")
+	//cmd := exec.Command("vcluster", "create", vclusterName, "--connect=false")
```
jainsmit, line 141: Remove the commented cmd

drivers/vcluster/vcluster.go
```diff
@@ -76,9 +126,19 @@ func DeleteVCluster(vclusterName string) error {
 	return nil
 }
 
+// TerminateVCluster Terminates Vcluster and runs it in it's own context
+func (v *VCluster) TerminateVCluster() error {
+	cmd := exec.Command("vcluster", "delete", v.Name)
```
jainsmit, line 131: Can we have a common method for executing vCluster commands ?

drivers/vcluster/vcluster.go
```diff
@@ -103,3 +163,373 @@ func WaitForVClusterRunning(vclusterName string, timeout time.Duration) error {
...
+		return "", err
+	}
+	_, err := v.Clientset.CoreV1().PersistentVolumeClaims(appNs).Create(context.TODO(), createOpts, metav1.CreateOptions{})
+	if err != nil {
+		return "", err
+	}
+	return pvcName, nil
+}
+
+// int32Ptr converts integre to pointer
```
jainsmit, line 397: Correct  "integer" spelling in a comment

drivers/vcluster/vcluster.go
```diff
@@ -103,3 +163,373 @@ func WaitForVClusterRunning(vclusterName string, timeout time.Duration) error {
...
+func int32Ptr(i int32) *int32 { return &i }
+
+// CreateFIODeployment creates a FIO Batch Job on given PVC
+func (v *VCluster) CreateFIODeployment(pvcName string, appNS string, fioOpts FIOOptions) error {
```
jainsmit, line 401: Is there any reason for adding this method? Can't we use existing way to deploy the FIO pods or other applications specs in vCluster?

drivers/vcluster/vcluster.go
dbhatnagar-px: Done. Thanks for suggesting this optimization.

drivers/vcluster/vcluster.go
dbhatnagar-px: Done

drivers/vcluster/vcluster.go
dbhatnagar-px: Hi Smit, yes I considered this. The problem with trying to use existing specs is that they are quite tightly coupled to their own constructs. Now there are 2 things we need to consider : 
1. With vcluster few things need to be created at Host layer and few at vcluster layer
2. Torpedo doesn't support multiple k8s constructs. For every k8s construct, we need to do an InitInstance() and this initinstance will need to be modified when we are trying to do a vcluster init. It's much better, quicker and flexible if we have vcluster constructs separately rather than trying to intrusively modify current Torpedo.

jainsmit: Can't we have vCluster scheduler driver which have it own node instances and when we deploying these app specs using vCluster scheduler instance that time it will deploy in vClutser.  

If required vCluster provisioning part can be moved to spawn.

tests/basic/vcluster_test.go
```diff
@@ -136,5 +169,36 @@ func RunNginxAppOnVcluster(args []interface{}) []interface{} {
...
+	}
 	return []interface{}{nil}
 }
+
+// CreateStorageClass method creates a storageclass using host's k8s clientset on host cluster
+func CreateStorageClass(scName string) error {
+	params := make(map[string]string)
+	params["repl"] = "2"
+	params["priority_io"] = "high"
+	params["io_profile"] = "db_remote"
```
Santhoshairody, line 186: Can you use auto instead of db_remote

drivers/vcluster/vcluster.go
```diff
@@ -76,10 +126,25 @@ func DeleteVCluster(vclusterName string) error {
...
+func (v *VCluster) TerminateVCluster() error {
+	_, err := ExecuteVClusterCommand("delete", v.Name)
+	return err
+}
+
 // CreateVCluster This method creates a vcluster. This requires vcluster.yaml saved in a specific location.
 func CreateVCluster(vclusterName string, absPath string) error {
-	cmd := exec.Command("vcluster", "create", vclusterName, "-f", absPath, "--connect=false")
-	err := cmd.Run()
+	_, err := ExecuteVClusterCommand("create", vclusterName, "-f", absPath, "--connect=false")
```
Santhoshairody, line 147: If we use vcluster on top of Managed kubernetes cluster like `EKS anywhere` then by default storage class would be not portworx storage class, during this time you need to change the default storageclass as mentioned here
`kubectl patch storageclass px-db -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'`
Otherwise we cannot create vCluster.
https://portworx.atlassian.net/wiki/spaces/TA/pages/2754543945/Aera+-+vCluster+Use+case#vCluster-Installation

This comment is not related to this changes, but we should have this fix in future changes.

dbhatnagar-px: Hi Santosh, noted and thanks. I am aware of this and will be taking care of this in my next PR. Along with downloading vcluster binary during Torpedo installation.

drivers/vcluster/vcluster.go
dbhatnagar-px: Discussed this - pasting summary here: 
I did consider this approach as I mentioned in the comment as well. The problem with this approach is it will cause intrusive changes in Torpedo. I don't believe that's the right thing to do in Torpedo when we have an equally great way out. The ROI in the other approach is only that we will be able to utilise existing app specs - which is not very flexible in vcluster context. We will get ROI when we do a lot of intrusive changes. Rest everything is anyways happening as is using Torpedo.
Regarding node instances of vcluster, initinstance will not work as is. We will have to make changes to make it work in vcluster context
Another thing is : Torpedo doesn't support more than one context at one time.

tests/basic/vcluster_test.go
dbhatnagar-px: Done

lsrinivas-pure (approved):
LGTM

mkoppal-px (approved):
LGTM

jainsmit (approved):
LGTM !!

PR #1766, opened 2023-10-17
Removed SetMissingClusterUID function from inspect cluster function

aks-px:
Removed the addition of UID incase if it is not present in the request from inspect cluster function so that  px-backup integration test negative scenarios would go through.

Closes #


kphalgun-px (approved):
LGTM

mkoppal-px (changes requested):
Run a few tests which are using these functions. Especially orphaned backup tests.

ak-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #1764, opened 2023-10-16
PWX-33942: restart px after expansion

alicelyu-px:
Added new test case "Restart PX after pool expansion" in pool_functional_test.go
Renamed poolToBeResized -> poolToResize

Successful test run at 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/alyu/job/pxdv-7-dev/77/console

PR #1767, opened 2023-10-17
PWX-33948: Try to resize multiple pools on one node, expects only one request is making progress as a given time

alicelyu-px:
Successful runs: 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/alyu/job/pxdv-7-dev/85
https://jenkins.pwx.dev.purestorage.com/job/Users/job/alyu/job/pxdv-7-dev/86


tests/basic/pool_function_test.go
```diff
@@ -84,6 +85,33 @@ var _ = Describe("{PoolExpandSmoky}", func() {
...
 		verifyPoolSizeEqualOrLargerThanExpected(poolIDToResize, targetSizeGiB)
 	})
+
+	// test resizing all pools at the same time and ensure only the first one makes progress
+	It("Select all pools and expand them concurrently. ", func() {
+		StartTorpedoTest("PoolExpandRejectConcurrent",
+			"Validate storage pool expansion rejects concurrent requests", nil, 0)
+		// collect all pools available
+		var pools []*api.StoragePool
+		nodes := node.GetNodes()
```
lsrinivas-pure, line 95: Please use node.GetStorageDriverNodes, node.GetNodes will get all the nodes including master node and nodes where portworx is not enabled. This will cause failures while running in cloud environments where we deploy a node pool for running torpedo

tests/basic/pool_function_test.go
```diff
@@ -84,6 +85,33 @@ var _ = Describe("{PoolExpandSmoky}", func() {
 		dash.VerifyFatal(resizeErr, nil, "Pool expansion does not result in error")
 		verifyPoolSizeEqualOrLargerThanExpected(poolIDToResize, targetSizeGiB)
 	})
+
+	// test resizing all pools at the same time and ensure only the first one makes progress
+	It("Select all pools and expand them concurrently. ", func() {
+		StartTorpedoTest("PoolExpandRejectConcurrent",
```
lsrinivas-pure, line 91: Having StartTorpedoTest here will discard pushing all the logs prior to this like in JustBeforeEach to aetos. Logs will printed in jenkins but not in aetos

alicelyu-px: Sorry what do you mean? 

I have a similar setup in https://github.com/portworx/torpedo/pull/1764, 
logs on aetos seem correct: 
https://aetos.pwx.purestorage.com/resultSet/testSetID/402969

lsrinivas-pure: it works fine when there is no error or log.InfoD in JustBeforeEach, for example you can change lof.Infof to log.InfoD in JustBeforeEach, you will notice the log statement missing in aetos

alicelyu-px: hmm this sounds like a bug that we should fix on the infra side..?

lsrinivas-pure: no, StartTorpedoTest will create a testsetid in aetos for a test and pushes all logs to the testsetid. Here testsetid is created after JustBeforeEach executed and logging in JustBeforeEach will not be pushed to aetos as it doesn't have testsetId at that time

alicelyu-px: Well in fact I think this is the correct behavior. JustBeforeEach contains test setup, and we do not want those logs to be present on aetos as part of the test per testsetID

pp511: What happens if test fails before StartTorpedoTest is called?
That failure doesn't get logged to Aetos at all? 
If that is true, we can be in a situation where tests fail but we continue to report failures to Aetos.

tests/basic/pool_function_test.go
```diff
@@ -143,6 +144,72 @@ var _ = Describe("{PoolExpandSmoky}", func() {
...
 		verifyPoolSizeEqualOrLargerThanExpected(poolIDToResize, targetSizeGiB)
 	})
+
+})
+
+var _ = Describe("{PoolExpandRejectConcurrent}", func() {
+	BeforeEach(func() {
+		contexts = scheduleApps()
+	})
+
```
pp511, line 154: Why are we not initializing a testRailID and not reporting to Aetos?

alicelyu-px: Added!

tests/basic/pool_function_test.go
```diff
@@ -143,6 +144,75 @@ var _ = Describe("{PoolExpandSmoky}", func() {
...
+		})
+
+		Step("Verify only one expansion is making progress at any given time", func() {
+			inProgressCount := 0
+			startTime := time.Now()
+			for time.Since(startTime) < 1*time.Minute {
+				inProgressCount = 0
+				time.Sleep(5)
+				storageNode, err = GetNodeWithGivenPoolID(poolIDToResize)
+				for _, p := range storageNode.Pools {
```
pp511, line 207: nit.. I think you could use for _, p := range pools to be consistent with the above for loop

tests/basic/pool_function_test.go
```diff
@@ -143,6 +144,75 @@ var _ = Describe("{PoolExpandSmoky}", func() {
...
+			// collect all pools available
+			for _, p := range storageNode.Pools {
+				pools = append(pools, p)
+			}
+			dash.VerifyFatal(len(pools) > 1, true, "This test requires more than 1 pool.")
+		})
+
+		Step("Expand all pools concurrently. ", func() {
+			expandType := api.SdkStoragePool_RESIZE_TYPE_ADD_DISK
+			var wg sync.WaitGroup
```
pp511, line 190: We are using a WaitGroup here because we want to wait for all the pool expand ops to be triggered before we move ahead. Shouldn't we be waiting on waitgroup. 	wg.Wait()  after the for loop?

tests/basic/pool_function_test.go
```diff
@@ -143,6 +144,75 @@ var _ = Describe("{PoolExpandSmoky}", func() {
...
+					defer wg.Done()
+					err = Inst().V.ExpandPool(p.Uuid, expandType, p.TotalSize/units.GiB+100, true)
+				}(p)
+			}
+		})
+
+		Step("Verify only one expansion is making progress at any given time", func() {
+			inProgressCount := 0
+			startTime := time.Now()
+			for time.Since(startTime) < 1*time.Minute {
```
pp511, line 203: I don't understand the logic here. Why do we need this for loop? Can you elaborate?

tests/basic/pool_function_test.go
```diff
@@ -143,6 +144,75 @@ var _ = Describe("{PoolExpandSmoky}", func() {
...
+				inProgressCount = 0
+				time.Sleep(5)
```
pp511, line 205: Explain the random sleep here. Why do we need it. Add comment in the code too if it's essentially needed.

tests/basic/pool_function_test.go
```diff
@@ -143,6 +144,75 @@ var _ = Describe("{PoolExpandSmoky}", func() {
...
+				storageNode, err = GetNodeWithGivenPoolID(poolIDToResize)
+				for _, p := range storageNode.Pools {
+					if p.LastOperation.Status == api.SdkStoragePool_OPERATION_IN_PROGRESS {
+						inProgressCount++
+					}
+					dash.VerifyFatal(inProgressCount <= 1, true, "Only one pool expansion should be in progress at any given time.")
```
pp511, line 211: It should be  inProgressCount == 1

pp511 (changes requested):

tests/basic/pool_function_test.go
alicelyu-px: We can no use `_, p := range pools` --
Here the storagePool is refreshed from the line above. 
This is needed before there was pool operation performed earlier.

tests/basic/pool_function_test.go
alicelyu-px: Sorry my bad!

tests/basic/pool_function_test.go
alicelyu-px: We check how many pools have expansion "in progress" every 5 seconds, and do that for 1 minute (12 checks in total). The test objective is to verify there can only be one pool "in progress" at any given time.

tests/basic/pool_function_test.go
alicelyu-px: It should be inProgressCount <= 1, 0 is feasible. This is because it is possible that none is "in progress". It can be the one making progress is still in "pending" stage(haven't moved to "in progress"), or it has completed(success).

tests/basic/pool_function_test.go
alicelyu-px: The test objective is to verify that only one pool is "in progress" at any given time. The test verifies it every 5 seconds.

pp511: We should use DoRetryWithTimeout to stay consistent with the rest of the code base https://github.com/portworx/torpedo/blob/7c88f114d3c3400be1e0baa65e98ee79d1dd9963/drivers/backup/portworx/portworx.go#L485C2-L485C2

pp511 (approved):
Changes look ok to me. 
Only suggestion I have is if we use DoRetryWithTimeout, it would be cleaner I feel. Other than that from logical standpoint the code looks good.

tests/basic/pool_function_test.go
lsrinivas-pure: yes aetos will have the any log regard the test if any failure occurring in JustBeforeEach. Jenkins run will be failed but in aetos it will be all green

alicelyu-px:
> Changes look ok to me. Only suggestion I have is if we use DoRetryWithTimeout, it would be cleaner I feel. Other than that from logical standpoint the code looks good.


Copying my responses on Slack here for future reference: 

> DoRetryWithTimeout aims to get an operation done within a certain timeout limit
> [6:50](https://purestorage.slack.com/archives/D04FT951KKJ/p1698976245017849)
> In our test case, we try to verify some logic is true within a said period
> [6:51](https://purestorage.slack.com/archives/D04FT951KKJ/p1698976265685199)
> These two use cases are distinctive
> [6:54](https://purestorage.slack.com/archives/D04FT951KKJ/p1698976453466089)
> Therefore I wouldn’t use DoRetryWithTimeout.. the time limit we set here is not a “timeout” but a hubristic time interval we perform the check over. 
> 
> we “could” use the `DoRetryWithTimeout` utility, but I feel like our use case here does not fit in the designed purpose of that function, and hope to avoid potential confusion in code reading in future.

pp511 (approved):

PR #1773, opened 2023-10-18
PWX-33997, PWX-34074: Added checks around local Pure mapper paths

Pure-AdamuKaapan:
This adds two extra functions to the Volume driver interface:

1. First, it adds a method to set the "baseline" of what Pure devices are attached to each node. This way, if the node has an FA boot drive or is using FA cloud drives for example, we don't get confused over those drives and think they are extra.

2. Second, it adds a method to list all the local Pure mapper paths on each node (excluding baseline devices), and validate that they are all corresponding to a volume that should be attached on that node. It will then raise an error if any of the mapper paths left on the device should not be attached, or if some of the baseline devices are no longer present then we also raise an error as we may have accidentally touched a disk we weren't supposed to.

**Note**: This will encounter some issues with FACD failover tests, as it will then be a different mapper device than originally and will trigger the baseline device error. I have left todos for where we can improve this in the future.

**Testing notes**:
Ran locally on my own cluster fully through both Pure tests, and confirmed that a) the tests passed, and b) the checks that it was running were making sense and passing as well. I also confirmed with a few failures while testing that it does indeed catch volumes that should have been deleted but were not.

Confirmed the same above for FADA + FBDA, with multiple apps at once, and confirmed it still passes

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -7007,7 +7007,29 @@ func (k *K8s) snapshotAndVerify(size resource.Quantity, data, snapName, namespac
 			return fmt.Errorf("restored volume does NOT contain data from original volume: expected to contain '%s', got '%s'", data, string(fileContent))
 		}
 	}
-	log.Info("Validation complete")
+
+	log.Info("Validation complete, deleting restored pods")
+	err = clientset.CoreV1().Pods(namespace).Delete(context.TODO(), restoredPod.Name, metav1.DeleteOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return fmt.Errorf("error deleting restored pod: %s", err)
+	}
+
+	// Wait for the pod to be actually gone
+	t := func() (interface{}, bool, error) {
+		var err error
+		if _, err = clientset.CoreV1().Pods(namespace).Get(context.TODO(), restoredPod.Name, metav1.GetOptions{}); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return nil, false, nil // Pod is now successfully gone
+			}
+			return nil, true, fmt.Errorf("failed to check that pod %s in ns %s is gone. Err: %v", restoredPod.Name, restoredPod.Namespace, err)
+		}
+		return nil, true, fmt.Errorf("pod %s in ns %s still exists", restoredPod.Name, restoredPod.Namespace)
+	}
+
+	if _, err := task.DoRetryWithTimeout(t, time.Minute*2, DefaultRetryInterval); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -7067,7 +7089,29 @@ func (k *K8s) cloneAndVerify(size resource.Quantity, data, namespace, storageCla
 			return fmt.Errorf("cloned volume does NOT contain data from original volume: expected to contain '%s', got '%s'", data, string(fileContent))
 		}
 	}
-	log.Info("Validation complete")
+
+	log.Info("Validation complete, deleting restored pods")
+	err = clientset.CoreV1().Pods(namespace).Delete(context.TODO(), restoredPod.Name, metav1.DeleteOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return fmt.Errorf("error deleting restored pod: %s", err)
+	}
+
+	// Wait for the pod to be actually gone
+	t := func() (interface{}, bool, error) {
+		var err error
+		if _, err = clientset.CoreV1().Pods(namespace).Get(context.TODO(), restoredPod.Name, metav1.GetOptions{}); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return nil, false, nil // Pod is now successfully gone
+			}
+			return nil, true, fmt.Errorf("failed to check that pod %s in ns %s is gone. Err: %v", restoredPod.Name, restoredPod.Namespace, err)
+		}
+		return nil, true, fmt.Errorf("pod %s in ns %s still exists", restoredPod.Name, restoredPod.Namespace)
+	}
+
+	if _, err := task.DoRetryWithTimeout(t, time.Minute*2, DefaultRetryInterval); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -6996,7 +6996,7 @@ func (k *K8s) snapshotAndVerify(size resource.Quantity, data, snapName, namespac
 			return fmt.Errorf("error checking content of cloned PVC: %s. Output: %s", err, string(fileContent))
 		}
 		if data != fileContent {
-			return fmt.Errorf("Compared data of text file & data copied to device path is not same")
+			return fmt.Errorf("compared data of text file & data copied to device path is not same")
 		}
 	} else {
 		fileContent, err := k.readDataFromPod(restoredPod.Name, namespace, "/mnt/volume1/aaaa.txt")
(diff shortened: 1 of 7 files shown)


Pure-AdamuKaapan:
I've encountered a bug regarding multiple apps and this code - I'm fixing it up and have also included the size check code. This will be ready to review again on Monday.

drivers/volume/portworx/portworx.go
```diff
@@ -1776,6 +1779,287 @@ func (d *portworx) ValidatePureVolumesNoReplicaSets(volumeName string, params ma
...
+		fadaVolumes := []*api.Volume{}
+		for _, volName := range allVolNames {
+			v, err := d.InspectVolume(volName)
+			if err != nil {
+				return err
+			}
+			if !v.Spec.IsPureVolume() {
+				continue
+			}
+			if v.Spec.ProxySpec.PureBlockSpec == nil {
```
pnookala-px, line 1992: Use new interface v.Spec.IsPureBlockVolume().

drivers/scheduler/k8s/k8s.go
```diff
@@ -6951,7 +6951,29 @@ func (k *K8s) snapshotAndVerify(size resource.Quantity, data, snapName, namespac
...
+				return nil, false, nil // Pod is now successfully gone
+			}
+			return nil, true, fmt.Errorf("failed to check that pod %s in ns %s is gone. Err: %v", restoredPod.Name, restoredPod.Namespace, err)
+		}
+		return nil, true, fmt.Errorf("pod %s in ns %s still exists", restoredPod.Name, restoredPod.Namespace)
+	}
+
+	if _, err := task.DoRetryWithTimeout(t, time.Minute*2, DefaultRetryInterval); err != nil {
+		return err
+	}
```
zoxpx, line 6975: I think this code is a bit "racey" -- it may happen that the K8s is very quick with restarting the PODs after they got deleted -- and we'd get "still exists" error.
I think a better approach would be to collect the POD-IDs before the deletion, then inside `t()` check if those specific POD-IDs are now gone.

Pure-AdamuKaapan: @zoxpx in this case the pods are not managed by a deployment or anything (they are one-off pods), so they won't be restarted

zoxpx: Well.. first off, do we actually need to wait until the pods are gone?
* note that the `kubectl delete pod -Al name=portworx` will actually hang until all portworx PODs will be in `Terminated` state
* not sure if K8s API behaves the same way -- but if it does, probably no need to have the extra code that waits for the PODs to actually disappear

Other than that, I don't think there's a big danger that PODs get restarted w/ the exact same UID --
* if the POD executable dies/crashes and POD gets restarted, then yes
* but I haven't actually seen the PODs getting restarted w/ the exact same name + UID after the successful 'pod delete' function..  (they _will_ recycle POD name/UID once restarted).

Finally, my apologies, I haven't initially noticed that you are dealing with a single POD, and you're deleting it "by name" (I thought you were deleing a group of PODs).   My initial feedback was based on this assumption -- but I think I was wrong, and I think this code actually works OK.

To speed things up, you might want to also check for the status of the POD -- and break out immediately when the status switches to `Terminated`  (AFAIK Terminated PODs are like zombie-processes, just need to be cleaned up from process-table, which could take a minute on a busy K8s cluster)

drivers/volume/portworx/portworx.go
```diff
@@ -1778,6 +1781,287 @@ func (d *portworx) ValidatePureVolumesNoReplicaSets(volumeName string, params ma
...
 }
 
+type pureLocalPathEntry struct {
+	WWID        string
+	SinglePaths []string
+	Size        uint64
+}
+
+func GetSerialFromWWID(wwid string) (string, error) {
+	if !strings.Contains(wwid, "24a937") {
```
zoxpx, line 1791: nit: you might want was to extract `"24a937"` as a constant, since you keep using it on multiple places.

Just curious, I assume the WWID is coming from SCSI-inquiry, pages 0x80/0x83 ?  How do we "get it out" (ioctl()'s or external commands?)

zoxpx (approved):
I think this looks OK -- but I didn't follow what you are doing w/ the volumes + lsblk

You still might want to get a review from someone more knowledgeable w/ the volume-handling specifics

drivers/scheduler/k8s/k8s.go
Pure-AdamuKaapan: We actually specifically want to wait for pods to be fully terminated: we've had a lot of bugs where FlashArray/FlashBlade pods get stuck in terminating, and we want this to fail loudly on it.

drivers/volume/portworx/portworx.go
Pure-AdamuKaapan: This WWID is actually produced from device-mapper multipath: it's `/dev/mapper/<WWID>` which contains the serial.

Good point, let me see if there's an existing constant

tests/basic/pure_test.go
```diff
@@ -84,6 +84,9 @@ var _ = Describe("{PureVolumeCRUDWithSDK}", func() {
 		Step("setup credential necessary for cloudsnap", createCloudsnapCredential)
 		contexts = make([]*scheduler.Context, 0)
 
+		err := Inst().V.InitializePureLocalVolumePaths() // Initialize our "baseline" of Pure devices, such as FACD devices or other local FA disks
```
adityadani, line 87: Does it make sense to move this into `JustBeforeEach` ?

tests/common.go
```diff
@@ -1059,6 +1059,11 @@ func ValidatePureSnapshotsSDK(ctx *scheduler.Context, errChan ...*chan error) {
 			processError(err, errChan...)
 		})
 
+		Step("validate Pure local volume paths", func() {
```
adityadani, line 1062: Why do this check here, instead of at the end once the snapshot is successful? Also does it make sense to move this validation into `AfterEach` ?

drivers/volume/portworx/portworx.go
```diff
@@ -1778,6 +1781,287 @@ func (d *portworx) ValidatePureVolumesNoReplicaSets(volumeName string, params ma
...
+		lsblkEntry, exists := lsblkParsed[mapper]
+		if !exists {
+			return nil, fmt.Errorf("found mapper %s in dmsetup but not lsblk on node %s, inconsistent disk state (we didn't clean something up right?)", mapper, n.MgmtIp)
+		}
+		// Ensure there are no WWIDs with empty single paths, as that is a malformed device
+		if len(lsblkEntry.SinglePaths) == 0 {
+			return nil, fmt.Errorf("found mapper %s with no single paths on node %s, this should not happen (we didn't clean something up right?)", mapper, n.MgmtIp)
+		}
+	}
+
```
adityadani, line 1925: Add a check for `len(lsblkParsed) == len(dmsetupFoundMappers)`, if for some reason dmsetup did not find a mapper path but lsblk found one.

adityadani: Actually this might not work since in `parseLsblkOutput` you are parsing all devices, even if they are not multipath. May be a keep count of mapper devices found in `lsblk` output and compare it with that.

drivers/volume/portworx/portworx.go
```diff
@@ -1778,6 +1781,287 @@ func (d *portworx) ValidatePureVolumesNoReplicaSets(volumeName string, params ma
...
+			if err != nil {
+				return err
+			}
+			if !v.Spec.IsPureVolume() {
+				continue
+			}
+			if v.Spec.ProxySpec.PureBlockSpec == nil {
+				// TODO: handle FBDA volumes properly as well
+				continue
+			}
```
adityadani, line 1997: No need of this since you are checking for IsPureVolume() ?

drivers/volume/portworx/portworx.go
```diff
@@ -1778,6 +1781,287 @@ func (d *portworx) ValidatePureVolumesNoReplicaSets(volumeName string, params ma
...
+			return err
+		}
+		logrus.Infof("Current pure devices: %+v", currentDevices)
+
+		// TODO: handle FACD drive failovers properly (these will show up as new devices but they're actually fine)
+		// Remove all devices that are in the baseline (complex set math time!). Also warn if any baseline devices are now missing?
+		for node, baselineDevices := range d.pureDeviceBaseline {
+			currentNodeDevices := currentDevices[node]
+			for wwid := range baselineDevices {
+				if _, exists := currentNodeDevices[wwid]; exists {
```
adityadani, line 1972: Should we compare the single paths here too? We want to make sure the number of singlePaths match before and after.

drivers/volume/portworx/portworx.go
```diff
@@ -1778,6 +1781,287 @@ func (d *portworx) ValidatePureVolumesNoReplicaSets(volumeName string, params ma
...
+				}
+			}
+			if foundNode == nil {
+				logrus.Infof("Volume %s is not attached to any node, skipping", v.Locator.Name)
+				continue
+			}
+
+			// Find the device for this volume
+			var device *pureLocalPathEntry
+			for _, deviceEntry := range currentDevices[foundNode.MgmtIp] {
```
adityadani, line 2020: same here: for FADA volumes we should check if singlePaths exist.

adityadani (approved):
lgtm - I feel the logic is a bit complicated and we might end up with failing jobs if the platform/OS/kernel changes and the lsblk or dmsetup outputs change slightly. But this is a great start!

Also it would be nice to have functions in portworx.go driver, where given a PVC name/ID it goes and checks on the node its corresponding multipath device exists or does not exist. In this way we can modify our `ValidateXXX` function to ensure that a pod is really healthy and using an actual device path.

drivers/volume/portworx/portworx.go
pnookala-px: IsPureVolume() returns true for both FA/FB.

Pure-AdamuKaapan: What Phani said, yeah, I want FADA specifically

drivers/volume/portworx/portworx.go
Pure-AdamuKaapan: We can but I think we tried this in PSO tests in the pass and we got pretty often false alarms due to temporary path failures - maybe if what we have so far proves to be very consistent we can add it later?

drivers/volume/portworx/portworx.go
Pure-AdamuKaapan: See my comment on the last one but yeah we could at least check that we have one single path, good point

tests/basic/pure_test.go
Pure-AdamuKaapan: Yes, honestly as well as the cloudsnap credentials.

tests/common.go
Pure-AdamuKaapan: I suppose that does make sense, yeah

harsh-px (approved):

PR #1783, opened 2023-10-20
Update utils.go status change

jkagliwal-px:
status changed from : 
Down status -> Partially Available
For PDS pods


aravindranath14 (approved):
LGTM

PR #1793, opened 2023-10-22
PA-1712: Update logic to use pod status instead of ns

kshithijiyer-px:

- [x] Rename `checkBackupObjectForNonExpectedNS` to `checkBackupObjectForUnexpectedNS`
- [x] Update logic to use pod status instead of ns

https://portworx.atlassian.net/browse/PB-4632 needs to be fixed to run this fix. 

Closes #PA-1712


trenukarya-px:
@kshithijiyer-px Please add more details why to change to POD instead of NS. Can't make out from the ticket number provided.

kshithijiyer-px:
> @kshithijiyer-px Please add more details why to change to POD instead of NS. Can't make out from the ticket number provided.

When we do a replace it usually cleans up the namespace but doesn't delete the namespace itself. The final output of this would be that the pods and pvcs would be recreated. Hence we need to check for pods age instead of ns age.

tests/backup/backup_helper.go
```diff
@@ -5140,37 +5140,43 @@ func checkBackupObjectForNonExpectedNS(ctx context.Context, backupName string) e
 	return nil
 }
 
-// getNamespaceAge gets the namespace age of all the namespaces on the cluster
-func getNamespaceAge() (map[string]time.Time, error) {
-	var namespaceAge = make(map[string]time.Time)
+// getPodAge gets the pod age of all pods on all the namespaces on the cluster
+func getPodAge() (map[string]time.Time, error) {
```
kphalgun-px, line 5144: I feel map[string][string]time.Time (map[namespace][pod]time.Time) would be more useful.

kphalgun-px (approved):
LGTM

tests/backup/backup_helper.go
kshithijiyer-px: Nice catch it would be much cleaner. It would slightly increase complexity but approach is much cleaner.

kshithijiyer-px:
> LGTM. Except for one minor nit. Can we rename the func `checkBackupObjectForNonExpectedNS` to `checkBackupObjectForUnexpectedNS`

Done

PR #1794, opened 2023-10-23
FA Volume token timeout scenario

lsrinivas-pure:
Test that validates FADA volumes token timeout when multiple requests hit same node at same time

Closes #PTX-20616

Result:
https://aetos.pwx.purestorage.com/resultSet/testSetID/391904

### DIFF
--- a/tests/basic/pure_test.go
+++ b/tests/basic/pure_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+	"math/rand"
 	"strings"
 	"sync"
 	"time"
@@ -162,7 +163,6 @@ var _ = Describe("{BringUpLargePodsVerifyNoPanic}", func() {
 				"FADA/Generic Volumes while kvdb failover in progress", nil, testrailID)
 		runID = testrailuttils.AddRunsToMilestone(testrailID)
 	})
-	var contexts []*scheduler.Context
 
 	stepLog := "Validate no panics when creating more number of pods on FADA/Generic " +
 		"Volumes while kvdb failover in progress"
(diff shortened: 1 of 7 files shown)


tests/basic/pure_test.go
```diff
@@ -339,3 +339,88 @@ var _ = Describe("{BringUpLargePodsVerifyNoPanic}", func() {
...
+
+		var wg sync.WaitGroup
+
+		stepLog = "Attaching 40 volumes at same time"
+		scheduleCount := 20 //here it is 20 as each app deploys 2 volumes
+		Step(stepLog, func() {
+			scheduleAppParallel := func() {
+				defer wg.Done()
+				defer GinkgoRecover()
+				contexts = append(contexts, ScheduleApplications(fmt.Sprintf(fmt.Sprintf("fadavoltkn-%d", i)))...)
```
spai-px, line 402: What's the logic for using two fmt.Sprintf(fmt.Sprintf("fadavoltkn-%d", i))?

tests/basic/pure_test.go
```diff
@@ -339,3 +339,88 @@ var _ = Describe("{BringUpLargePodsVerifyNoPanic}", func() {
...
+			}
+			ValidateApplications(contexts)
+		})
+
+		var wg sync.WaitGroup
+
+		stepLog = "Attaching 40 volumes at same time"
+		scheduleCount := 5 //20 //here it is 20 as each app deploys 2 volumes
+		Step(stepLog, func() {
+			scheduleAppParallel := func() {
```
kphalgun-px, line 399: How about taking count as param?

lsrinivas-pure: we already have a param GlobalScaleFactor but it is global and will be applied for all tests.
Do not want to create new param specific to this test

tests/basic/pure_test.go
lsrinivas-pure: it is just to create to have different namespaces

tests/basic/pure_test.go
kphalgun-px: `i` is accessed by multiple goroutines. How about we pass it (or count) as a parameter to prevent data races.

lsrinivas-pure: fixed it

aravindranath14 (approved):
LGTM

PR #1824, opened 2023-10-31
remove mount check in case of FBDA

vprabhakar-px:
Comment the mount validation as it is removed from FBDA

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -858,27 +858,28 @@ func ValidateContextForPureVolumesSDK(ctx *scheduler.Context, errChan ...*chan e
 			}
 		})
 
-		driverVersion, err := Inst().V.GetDriverVersion()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


Pure-AdamuKaapan (changes requested):
@vprabhakar-px I have a few concerns here:
1. How does checking the mount path relate to PWX-34377 and RWX volumes? As far as I can tell these two shouldn't be related
2. It seems like we're commenting out more than just the mount path validation, we're also removing the create options check and the entire mount options check (not just the path part). Can we narrow what we're commenting out?
3. These functions apply for both FADA and FBDA, so commenting it here disables it for all tests, not just FBDA, which I don't think is what we want
4. Can we leave a more in-depth comment than just linking to an issue? Just a bit more details on the specific reason why, as we may lose context on tickets over time

PR #1828, opened 2023-10-31
Fixed error handling in CreateRestoreWithValidation function

mkoppal-px:

- [x] `CreateRestoreWithValidation` function was never handling the error returned by `ValidateRestore` function. Fixed that by returning the proper error.
- [x] Removed the receiver type for `GetUpdatedSpec`
- [x] Added a case to handle the restored volume status in `ValidateRestore` if the status was `Retain` incase of in-place restore

Closes #PB-4759

[Successful Jenkins Run on RKE](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/basic-system-test/job/px/job/s3/job/px-backup-system-test-rke/349/)
[Aetos link](https://aetos.pwx.purestorage.com/resultSet/testSetID/403906)
There is one failed test `MultipleProjectsAndNamespacesBackupAndRestore`
This has been failing from the beginning but we were unable to catch it because the `CreateRestoreWithValidation` function was not returning the error.
Created a ticket to track this and assigned it to @snigdha-px - https://portworx.atlassian.net/browse/PB-4760

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -1460,7 +1460,7 @@ func (k *K8s) UpdateTasksID(ctx *scheduler.Context, id string) error {
 }
 
 // GetUpdatedSpec gets the updated spec of a K8s Object. Just `name`, and if required `namespace` must be specified on `spec` in order to GET the spec from K8s (if it exists).
-func (k *K8s) GetUpdatedSpec(spec interface{}) (interface{}, error) {
+func GetUpdatedSpec(spec interface{}) (interface{}, error) {
 	if specObj, ok := spec.(*appsapi.Deployment); ok {
 		dep, err := k8sApps.GetDeployment(specObj.Name, specObj.Namespace)
 		if err == nil {
--- a/tests/backup/backup_helper.go
+++ b/tests/backup/backup_helper.go
@@ -1680,12 +1680,7 @@ func ValidateBackup(ctx context.Context, backupName string, orgID string, schedu
 					continue volloop
 				}
 
-				sched, ok := Inst().S.(*k8s.K8s)
-				if !ok {
-					continue volloop
-				}
-
-				updatedSpec, err := sched.GetUpdatedSpec(pvcSpecObj)
+				updatedSpec, err := k8s.GetUpdatedSpec(pvcSpecObj)
 				if err != nil {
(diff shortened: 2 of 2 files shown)


snigdha-px (approved):
LGTM

tests/backup/backup_helper.go
```diff
@@ -946,25 +946,28 @@ func CreateRestoreWithoutCheck(restoreName string, backupName string,
...
-			return
+			return err
 		}
 	}
 	defer func() {
 		log.InfoD("Switching cluster context back to cluster path [%s]", originalClusterConfigPath)
-		err = SetClusterContext(originalClusterConfigPath)
+		err := SetClusterContext(originalClusterConfigPath)
+		if err != nil {
+			log.Errorf("Failed switching cluster context back to cluster path [%s].\nError - %s", originalClusterConfigPath, err.Error())
```
kphalgun-px, line 969: How about log.FailOnError?

kphalgun-px (approved):
LGTM

tests/backup/backup_helper.go
mkoppal-px: I can do `log.FailOnError`
But not sure how will it work if I just append to the error that is being returned by the function.

kphalgun-px: The test case will fail right if err != nil?

mkoppal-px: Added `log.FailOnError`

dbhatnagar-px (approved):
Lgtm

drivers/scheduler/k8s/k8s.go
```diff
@@ -1460,7 +1460,7 @@ func (k *K8s) UpdateTasksID(ctx *scheduler.Context, id string) error {
 }
 
 // GetUpdatedSpec gets the updated spec of a K8s Object. Just `name`, and if required `namespace` must be specified on `spec` in order to GET the spec from K8s (if it exists).
-func (k *K8s) GetUpdatedSpec(spec interface{}) (interface{}, error) {
+func GetUpdatedSpec(spec interface{}) (interface{}, error) {
```
kshithijiyer-px, line 1463: @lsrinivas-pure Are we sure this will not created issues on the px side of things for you folks?

mkoppal-px: I checked this. This function was created and called only by backup tests.
Moreover the receiver type is redundant here.

kshithijiyer-px: Okay cool

kshithijiyer-px (approved):
LGTM

PR #1843, opened 2023-11-04
PB-4812: Fix pod age compare logic

kshithijiyer-px:
The current comparePodAge logic breaks because it was dependent on actual name for comparing instead of generated name due to which it will fails as the hash at the end of the  end of pod name like in  `postgres-7fb9569c6b-4mkzv` it is `4mkzv` which fails. 

To fix I am doing the following fixes in this patch: 

- [x] Change logic to use `pod.ObjectMeta.GetGenerateName()` instead of `pod.ObjectMeta.GetName()`
- [x] Use `IsPresent()` to check if namespace in skip list
- [x] Remove redundant kubeconfig switches
- [x] Minor string fixes

Dashboard URL: 
https://aetos.pwx.purestorage.com/resultSet/testSetID/413330 (Latest one with IsPresent)
https://aetos.pwx.purestorage.com/resultSet/testSetID/412342

Closes #PB-4812


tests/backup/backup_helper.go
```diff
@@ -5241,20 +5229,21 @@ func comparePodAge(oldPodAge map[string]nsPodAge) error {
...
-			if skipCase == namespace.GetName() {
-				pods, err := k8sCore.GetPods(namespace.ObjectMeta.GetName(), make(map[string]string))
-				if err != nil {
-					return fmt.Errorf("failed to get pods for namespace")
+		pods, err := k8sCore.GetPods(namespace.ObjectMeta.GetName(), make(map[string]string))
+		if err != nil {
+			return fmt.Errorf("failed to get pods for namespace")
+		}
+		log.InfoD("slice utils %v", sliceutil.Contains(namespacesToSkip, namespace.ObjectMeta.GetName()))
+		if sliceutil.Contains(namespacesToSkip, namespace.ObjectMeta.GetName()) {
```
KrishnaPhalgun28, line 5237: How about IsPresent?

kshithijiyer-px: I felt it was better to use inbuil function instead. What's the advantage of using IsPresent() instead?

sn-px (approved):
LGTM

tests/backup/backup_helper.go
kphalgun-px: I felt there is no need to import a new package.

KrishnaPhalgun28 (approved):
LGTM

kphalgun-px (approved):
LGTM

tests/backup/backup_helper.go
```diff
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"reflect"
 	"regexp"
+	"sigs.k8s.io/kustomize/kyaml/sliceutil"
```
mkoppal-px, line 12: Why is this package needed?

kshithijiyer-px: It's used in line 5236 for checking if a namespace is in a list of namespaces or not

mkoppal-px: We already have a function like this in `backup_helper.go`

```
// IsPresent verifies if the given data is present in slice of data
func IsPresent(dataSlice interface{}, data interface{}) bool {
	s := reflect.ValueOf(dataSlice)
	for i := 0; i < s.Len(); i++ {
		if s.Index(i).Interface() == data {
			return true
		}
	}
	return false
}
```

kshithijiyer-px: What's the benefit of using this over the one which I have used?

mkoppal-px: I do not see any difference between these two functions and it will avoid an import.

kshithijiyer-px: Done, I still don't understand why we need it.

mkoppal-px (approved):
LGTM

PR #1851, opened 2023-11-07
PWX-34918, PWX-34919: Supports FADA and FBDA volumes in SetupTeardown and AppScaleUpAndDown test

Pure-AdamuKaapan:
As part of an initiative to ensure that Direct Access volumes work with the standard Torpedo tests, we're starting with the SetupTeardown test. No change was actually required for this one, though I was able to also re-enable a check that the proper mount path is present.

More PRs will be coming in the future to further expand functionality - for now I'm just doing what is required to get the tests to pass with relevant checks.

**Testing notes**:
Tested with both an FADA and FBDA app (elasticsearch,mysql-without-enc) and validated the proper checks showed up in the output.

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -34,6 +34,7 @@ import (
 	volsnapv1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	snapv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	apapi "github.com/libopenstorage/autopilot-api/pkg/apis/autopilot/v1alpha1"
+	osapi "github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/pkg/units"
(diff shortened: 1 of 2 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -4137,6 +4146,32 @@ func (k *K8s) appendVolForPVC(vols []*volume.Volume, pvc *v1.PersistentVolumeCla
...
+		labels = make(map[string]string)
+	}
+	if inspectedVol.Spec.IsPureVolume() {
+		switch inspectedVol.Spec.ProxySpec.ProxyProtocol {
+		case osapi.ProxyProtocol_PROXY_PROTOCOL_PURE_BLOCK:
+			labels[PureDAVolumeLabel] = PureDAVolumeLabelValueFA
+			labels[FADAVolumeSerialLabel] = inspectedVol.Spec.ProxySpec.PureBlockSpec.SerialNum
+		case osapi.ProxyProtocol_PROXY_PROTOCOL_PURE_FILE:
+			labels[PureDAVolumeLabel] = PureDAVolumeLabelValueFB
+		}
```
alicelyu-px, line 4172: Don't we need a default clause?

alicelyu-px (approved):

PR #1862, opened 2023-11-08
Skip Anthos admin cluster upgrade if upgrade hops contains only one v…

jainsmit:
…ersion

[Anthos] Skip Anthos admin cluster upgrade if upgrade hops contains only one version

Closes # PTX-21038

Minor changes
```
11:53:33 2023-11-09 19:53:46 +0000:[INFO] [anthos.(*anthos).Init:#210] - Skip admin cluster upgrade is: [true]
```
Upgrade without admin cluster upgrade: https://jenkins.pwx.dev.purestorage.com/job/Torpedo/job/tp-anthos-upgrade/265/console

### DIFF
--- a/drivers/scheduler/anthos/anthos.go
+++ b/drivers/scheduler/anthos/anthos.go
@@ -161,12 +161,13 @@ type AnthosInstance struct {
 type anthos struct {
 	version string
 	kube.K8s
-	adminWsSSHInstance *ssh.SSH
-	instances          []AnthosInstance
-	adminWsNode        *node.Node
-	adminWsKeyPath     string
-	instPath           string
-	confPath           string
+	adminWsSSHInstance  *ssh.SSH
+	instances           []AnthosInstance
+	adminWsNode         *node.Node
+	adminWsKeyPath      string
+	instPath            string
+	confPath            string
+	adminClusterUpgrade bool
 }
 
 // Init Initialize the driver
@@ -200,6 +201,10 @@ func (anth *anthos) Init(schedOpts scheduler.InitOptions) error {
 	if err := anth.getVersion(); err != nil {
 		return err
 	}
+	if len(schedOpts.UpgradeHops) > 0 && len(strings.Split(schedOpts.UpgradeHops, ",")) > 1 {
+		anth.adminClusterUpgrade = true
+	}
+	log.Infof("Skip admin cluster upgrade is: [%t]", anth.adminClusterUpgrade)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


drivers/scheduler/anthos/anthos.go
```diff
@@ -200,6 +201,13 @@ func (anth *anthos) Init(schedOpts scheduler.InitOptions) error {
 	if err := anth.getVersion(); err != nil {
 		return err
 	}
+	anth.skipAdminClusterUpgrade = true
```
Rohit-PX, line 204: instead of `skipAdminClusterUpgrade` if you define `adminClusterUpgrade` and then you would not need this line
```
anth.skipAdminClusterUpgrade = true
```

Then you can set anth.adminClusterUpgrade = true on line: 207

Rohit-PX (changes requested):
LGTM.

drivers/scheduler/anthos/anthos.go
jainsmit: Good catch. Incorporated this comment.

Rohit-PX (approved):

drivers/scheduler/anthos/anthos.go
```diff
@@ -200,6 +201,12 @@ func (anth *anthos) Init(schedOpts scheduler.InitOptions) error {
 	if err := anth.getVersion(); err != nil {
 		return err
 	}
+	if len(schedOpts.UpgradeHops) > 0 {
```
nikolaypopov, line 204: nit: you can also do this instead
```suggestion
	if len(schedOpts.UpgradeHops) > 0 && if len(strings.Split(schedOpts.UpgradeHops, ",")) > 1 {
		anth.adminClusterUpgrade = true
	}
```

drivers/scheduler/anthos/anthos.go
```diff
@@ -200,6 +201,12 @@ func (anth *anthos) Init(schedOpts scheduler.InitOptions) error {
...
+		if len(strings.Split(schedOpts.UpgradeHops, ",")) > 1 {
+			anth.adminClusterUpgrade = true
+		}
+	}
```
nikolaypopov, line 208: nit: you can also do this
```suggestion
	if len(schedOpts.UpgradeHops) > 0 && if len(strings.Split(schedOpts.UpgradeHops, ",")) > 1 {
		anth.adminClusterUpgrade = true
	}
```

jainsmit: Thanks!! Resolved.

PR #1863, opened 2023-11-09
Fix Validation for Resource, Storage Configuration

Madanagopal19:
This PR has the fix for validating the storage and resource configuration of a data service

I have added the support to validate the resource and storage validation from db object. Initially we were using the values from the last applied configuration annotation of a storage class.
Going Forward this last applied configuration will be removed and also db object will be our source of truth to validate the configurations

Pipeline run - https://jenkins.pwx.dev.purestorage.com/job/PDS/job/pds-branch-build-test/649/


aravindranath14 (approved):
LGTM

PR #1868, opened 2023-11-09
PTX-20400 Create Torpedo SCC for OCP cluster to allow app provisioning

nikolaypopov:
Create Torpedo SCC for OCP cluster to allow app provisioning by copying and renaming default privileged SCC

**NOTE:** Current method to update SCC doesn't work for ROSA, because updating default SCC is not allowed there


mborodin-px (approved):
LGTM

PR #1873, opened 2023-11-13
Fixed a recent Pure volume change for IBM drivers

Pure-AdamuKaapan:
My recent change #1851 accidentally broke tests for the IBM platform driver as the InspectVolume method is not supported. This adds an extra error check for if the driver does not support the operation, then it will treat it as not a Pure volume.

Many thanks to @mkoppal-px for root-causing the original issue.

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -4158,21 +4158,26 @@ func (k *K8s) appendVolForPVC(vols []*volume.Volume, pvc *v1.PersistentVolumeCla
 
 		inspectedVol, err := driver.InspectVolume(pvc.Spec.VolumeName)
 		if err != nil {
-			return nil, fmt.Errorf("failed to inspect volume %s: %v", pvc.Name, err)
-		}
-
-		if labels == nil {
-			labels = make(map[string]string)
-		}
-		if inspectedVol.Spec.IsPureVolume() {
-			switch inspectedVol.Spec.ProxySpec.ProxyProtocol {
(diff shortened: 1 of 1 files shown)


drivers/scheduler/k8s/k8s.go
```diff
@@ -4158,21 +4158,26 @@ func (k *K8s) appendVolForPVC(vols []*volume.Volume, pvc *v1.PersistentVolumeCla
...
-			switch inspectedVol.Spec.ProxySpec.ProxyProtocol {
-			case osapi.ProxyProtocol_PROXY_PROTOCOL_PURE_BLOCK:
-				labels[PureDAVolumeLabel] = PureDAVolumeLabelValueFA
-				labels[FADAVolumeSerialLabel] = inspectedVol.Spec.ProxySpec.PureBlockSpec.SerialNum
-			case osapi.ProxyProtocol_PROXY_PROTOCOL_PURE_FILE:
-				labels[PureDAVolumeLabel] = PureDAVolumeLabelValueFB
-			default:
-				return nil, fmt.Errorf("unknown proxy type %v for Pure volume", inspectedVol.Spec.ProxySpec.ProxyProtocol)
+			if _, ok := err.(*errors.ErrNotSupported); !ok {
+				return nil, fmt.Errorf("failed to inspect volume %s: %v", pvc.Name, err)
```
mkoppal-px, line 4162: Can we return error as nil here?

kshithijiyer-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #1874, opened 2023-11-13
PWX-34305: added autopilot functional test suite

ezhang-px:
It added a functional test suite which includes three autopilot test cases:
1. running pool expansion and rebalance at the same time
2. resize large quantity of volumes (100) at the same time
3. resize sharedv4 volume

job link at: 
https://jenkins.pwx.dev.purestorage.com/job/Autopilot/job/tp-autopilot-functional-tests/ 

Closes #


tests/basic/autopilot_test.go
```diff
@@ -1848,6 +1848,204 @@ var _ = Describe(fmt.Sprintf("{%sRebalanceWithApproval}", testSuiteName), func()
...
+	JustBeforeEach(func() {
+		StartTorpedoTest(fmt.Sprintf("{%sFunctionalTests}", testSuiteName), "Perform several autopilot functional tests", nil, testrailID)
+		runID = testrailuttils.AddRunsToMilestone(testrailID)
+	})
+
+	It("has to run rebalance and resize pools, validate rebalance, validate pools and teardown apps", func() {
+		var contexts []*scheduler.Context
+		testName := strings.ToLower(fmt.Sprintf("%srebalance", testSuiteName))
+		poolLabel := map[string]string{"autopilot": "resizedisk"}
+		storageNodes := node.GetStorageDriverNodes()
```
aghodke1312, line 1866: This will return all the nodes with storageDriver installed. Do we want that? If you need only storagenodes, there is a method: `node.GetStorageNodes()` will return only nodes which are part of storage pool

ezhang-px: I think it's needed because the job is to expand pool that uses cloud drive. 
Tried to change to GetStorageNodes in my test branch (https://github.com/portworx/torpedo/blob/ezhang/test/tests/basic/autopilot_test.go#L1866) and rerun the pipeline, it failed due to errors like 

> device path is not present for volume: pvc-c3ec5c90-1205-4c1a-bdac-1630604d11e0

https://jenkins.pwx.dev.purestorage.com/job/Autopilot/job/tp-autopilot-functional-tests/16/consoleFull

aghodke1312 (approved):

PR #1875, opened 2023-11-14
Chiachun/pwx 34955 tp pool delete variations

chiachunlin:
Migrate pool delete px-test to torpedo:
PoolDeleteVariations,  PoolDeleteServiceDisruption

Tests PASSED: 
vsphere
https://jenkins.pwx.dev.purestorage.com/job/Users/job/ChiaChun/job/dmthin-clouddrive/44/
AKS
https://jenkins.pwx.dev.purestorage.com/job/Users/job/ChiaChun/job/chialin-aks-dmthin-tp-test/6/
GKE 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/ChiaChun/job/chialin-gcp-dmthin-tp-test/44/

Closes #

### DIFF
--- a/tests/basic/storage_pool_test.go
+++ b/tests/basic/storage_pool_test.go
@@ -10462,10 +10462,10 @@ var _ = Describe("{PoolDeleteVariations}", func() {
 		stepLog = fmt.Sprintf("2. Verify pool deletion after creating vols and each with %v snaps %v", numVols, numVolSnaps)
 
 		Step(stepLog, func() {
-			deletablePools := make (map[string]string) // poolUUID to ID
-			for _, p := range testNode.Pools {
-				deletablePools[p.Uuid] = fmt.Sprintf("%v",p.GetID())
-			}
+			deletablePools, err := Inst().V.GetNodePools(*testNode)
+			log.FailOnError(err, "failed to get node pool info")
+			log.Infof("Deletable pools %+v", deletablePools)
+
 			volIDs := []string{}
 			for i:=0; i<numVols; i++ {
 				uuidObj := uuid.New()
@@ -10490,14 +10490,13 @@ var _ = Describe("{PoolDeleteVariations}", func() {
 				}
 			}
 
-			log.Infof("Deletable pools %+v", deletablePools)
 			// select a pool with volume on it to delete
 			poolIDToDelete := ""
 			for _, vol := range snapshotList {
 				appVol, err := Inst().V.InspectVolume(vol)
 				log.FailOnError(err, fmt.Sprintf("err inspecting vol : %s", vol))
 				replPools := appVol.ReplicaSets[0].PoolUuids 
-				log.Infof("vol %+v, replPools %+v", vol, appVol, replPools)
+				log.Infof("vol %+v, replPools %+v", vol, replPools)
 				for _, p := range replPools {
 					if id, ok := deletablePools[p]; ok {
 						poolIDToDelete = id
--- a/tests/basic/storage_pool_test.go
+++ b/tests/basic/storage_pool_test.go
@@ -10458,7 +10458,7 @@ var _ = Describe("{PoolDeleteVariations}", func() {
 		})
 
 		numVolSnaps := 3
-		numVols := 20
+		numVols := 25
 		stepLog = fmt.Sprintf("2. Verify pool deletion after creating vols and each with %v snaps %v", numVols, numVolSnaps)
 
 		Step(stepLog, func() {
@@ -10490,12 +10490,14 @@ var _ = Describe("{PoolDeleteVariations}", func() {
 				}
 			}
 
+			log.Infof("Deletable pools %+v", deletablePools)
 			// select a pool with volume on it to delete
 			poolIDToDelete := ""
 			for _, vol := range snapshotList {
 				appVol, err := Inst().V.InspectVolume(vol)
 				log.FailOnError(err, fmt.Sprintf("err inspecting vol : %s", vol))
 				replPools := appVol.ReplicaSets[0].PoolUuids 
+				log.Infof("vol %+v, replPools %+v", vol, appVol, replPools)
 				for _, p := range replPools {
 					if id, ok := deletablePools[p]; ok {
 						poolIDToDelete = id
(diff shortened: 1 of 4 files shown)


tests/basic/storage_pool_test.go
```diff
@@ -9904,3 +9904,255 @@ var _ = Describe("{AddDriveWithKernelPanic}", func() {
...
+		testNode = selectPoolDeletableNode()
+		dash.VerifyFatal(testNode != nil, true, "very if select test node ok")
+	  })
+	var contexts []*scheduler.Context
+	
+ 	numVolCreate := 50
+	itLog := fmt.Sprintf("1. Verify pool deletion after creating %v volumes", numVolCreate)
+	It(itLog, func() {
+		log.InfoD(itLog)
+		poolsMap, err := Inst().V.GetPoolDrives(testNode)
```
dgoel-px, line 9970: `testNode.StoragePools[0]` can also work

tests/basic/storage_pool_test.go
```diff
@@ -9904,3 +9904,255 @@ var _ = Describe("{AddDriveWithKernelPanic}", func() {
 	})
 
 })
+
+func selectPoolDeletableNode () (*node.Node){
```
dgoel-px, line 9908: why only check for kvdb nodes? What about metadata / journal on the pool that we are selecting?

tests/basic/storage_pool_test.go
chiachunlin: mmm, it will use a cached value, since I create and delete the pool, the StoragePools will be obsoleted.

tests/basic/storage_pool_test.go
```diff
@@ -9904,3 +9904,255 @@ var _ = Describe("{AddDriveWithKernelPanic}", func() {
...
+		for i:=0; i< numVolCreate; i++{
+			uuidObj := uuid.New()
+			volName := fmt.Sprintf("volume_%s", uuidObj.String())
+			size := uint64(rand.Intn(5) + 1)   // Size of the Volume between 1G to 5G
+			ha := int64(rand.Intn(3) + 1) // Size of the HA between 1 and 3
+	
+			volId, err := Inst().V.CreateVolume(volName, size, int64(ha))
+			log.FailOnError(err, "volume creation failed on the cluster with volume name [%s]", volName)
+			volumesCreated[volId]=true
+		}
```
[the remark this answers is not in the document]
chiachunlin: I created 10 or 50 vols with random replicas. It's unlikely that the pool has no replicas on it. but of course, there is a tiny chance for that to happen.

tests/basic/storage_pool_test.go
```diff
@@ -9904,3 +9904,255 @@ var _ = Describe("{AddDriveWithKernelPanic}", func() {
...
+				snapshotName := fmt.Sprintf("snapshot_%s_%s", volID, uuidCreated.String())
+				snapshotResponse, err := Inst().V.CreateSnapshot(volID, snapshotName)
+				log.FailOnError(err, "error identifying volume [%s]", volID)
+				snapID := snapshotResponse.GetSnapshotId()
+				snapshotList = append(snapshotList, snapID)
+				log.InfoD("Snapshot [%s] created with ID [%s]", snapshotName, snapID)
+			}
+		}
+		
+		// delete should fail
```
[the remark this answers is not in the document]
chiachunlin: same as above.

tests/basic/storage_pool_test.go
chiachunlin: because internal kvdb nodes cannot become storageless node. we cannot delete pool on this node if there is only one. So, if all nodes are internal kvdb nodes, we have to add a pool to run test.

tests/basic/storage_pool_test.go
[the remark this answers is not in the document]
chiachunlin: if this happens, then I think we have problem in provisioning replicas. it is unbalanced.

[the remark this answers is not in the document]
chiachunlin: I updates the vol number to 20 and ha to 3. This should be good enough I think.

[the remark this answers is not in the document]
dgoel-px: I agree it'll likely not be a problem but we can avoid any flakiness here. Maybe use `GetPoolIDsFromVolName` and `GetPoolIDFromPoolUUID` to get the pool ID from the volume name that you already have and then delete it.

tests/basic/storage_pool_test.go
dgoel-px: But then in the `PoolDeleteServiceDisruption` test you are using `StoragePools` to get the pool to delete. Maybe change it there? Or we can call `Inst().S.RefreshNodeRegistry()` in `JustBeforeEach` for the tests you and I add.

dgoel-px: I would prefer the latter approach so that the node registry is always updated after each test.

dgoel-px (approved):

tests/basic/storage_pool_test.go
dgoel-px: If auto journal is enabled on the node, even then we can't delete all the pools. That's also a good check to add so that if the job is ever run with a different setup we'll know exactly why it fails or we can even skip the test if no suitable node is found.

tests/basic/storage_pool_test.go
```diff
@@ -10032,3 +10033,287 @@ func exitPoolMaintenance(poolId string) {
...
+					uuidCreated := uuid.New()
+					snapshotName := fmt.Sprintf("snapshot_%s_%s", volID, uuidCreated.String())
+					snapshotResponse, err := Inst().V.CreateSnapshot(volID, snapshotName)
+					log.FailOnError(err, "error creating snapshot [%s]", volID)
+					snapID := snapshotResponse.GetSnapshotId()
+					snapshotList = append(snapshotList, snapID)
+					log.InfoD("Snapshot [%s] created with ID [%s]", snapshotName, snapID)
+				}
+			}
+
```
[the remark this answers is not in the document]
chiachunlin: @lsrinivas-pure sure, I updated the test

PR #1878, opened 2023-11-14
[PWX-33955]:  [resize-disk]Check for alerts when pool resize is triggered and done

ragsingh-px:
Automation test for to check for alerts when pool resize is triggered and done using resize disk
Closes # PWX-33955

Testing link:- https://jenkins.pwx.dev.purestorage.com/user/ragsingh@purestorage.com/my-views/view/all/job/Users/job/Raghvendra/job/raghav_pxdv-7-dev/50


tests/basic/pool_function_test.go
```diff
@@ -859,3 +860,87 @@ var _ = Describe("{PoolExpandAddDiskInMaintenanceMode}", func() {
...
+					// break
+				} else {
+					return false, fmt.Errorf("Current pool size after expansion %s GiB is less than expected size %d GiB", matchedSize, targetSizeGiB)
+				}
+			}
+			// return false, fmt.Errorf("Alert does not mention the Size of the Pool")
+
+		}
+	}
+	return true, nil
```
vinayakshnd, line 945: If control comes here, means no alert found for specific pool for specific size. Return false and error msg here.

tests/basic/pool_function_test.go
```diff
@@ -859,3 +860,87 @@ var _ = Describe("{PoolExpandAddDiskInMaintenanceMode}", func() {
...
+			if re.MatchString(line) {
+				matchedSize := re.FindStringSubmatch(line)[0]
+				poolSize := matchedSize[:len(matchedSize)-4]
+				poolSizeUint, _ := strconv.ParseUint(poolSize, 10, 64)
+				if poolSizeUint >= targetSizeGiB {
+					// alertExist = true
+					log.Infof("The Alert generated is %s", line)
+					return true, nil
+					// break
+				} else {
```
vinayakshnd, line 937: This else block is not required. 
If `PoolExpandSuccessful` alert is seen for given `poolIDToResize` and for `targetSizeGiB`, only then return true(ie. Alert found), In all rest of the cases, (ie. when for loop exits), just return false, error (line no. 945) that expected alert not found.

tests/basic/pool_function_test.go
```diff
@@ -859,3 +860,87 @@ var _ = Describe("{PoolExpandAddDiskInMaintenanceMode}", func() {
...
+		dash.VerifyFatal(resizeErr, nil, "Pool expansion does not result in error")
+
+		log.Infof("Check the alert for pool expand for pool uuid %s", poolIDToResize)
+		alertExists, _ := checkAlertsForPoolExpansion(poolIDToResize, targetSizeGiB)
+		dash.VerifyFatal(alertExists, true, "Verify Alert is Present")
+	})
+
+})
+
+func checkAlertsForPoolExpansion(poolIDToResize string, targetSizeGiB uint64) (bool, error) {
```
vinayakshnd, line 908: Move this function to utils package as Public method so other tests can also use it.

vinayakshnd (changes requested):

tests/basic/pool_function_test.go
ragsingh-px: fixed.

tests/basic/pool_function_test.go
ragsingh-px: fixed.

vinayakshnd (approved):
LGTM

tests/basic/pool_function_test.go
ragsingh-px: not moved as the common/utils.go file has k8s related. functions

PR #1881, opened 2023-11-14
PWX-33944 

alicelyu-px:
This test is based on the https://github.com/portworx/torpedo/blob/acf4287ba1222d0385fdbadbc9366e2e80f202fd/tests/basic/storage_pool_test.go#L4630 

Main changes: 
* Use fio-fastpath-repl1 instead of `fio-fastpath` as the test app.`fio-fastpath` uses 5 replicas, which is excessive for this test. Using `fio-fastpath-repl1` reduces test run time and risk of failures due to infra issues.
* Removed the logic around introducing and managing a `secondReplNode`. This is confirmed feasible @lsrinivas-pure  
* Removed journal related logic. 
* Various test structure improvements  

Test passed at https://jenkins.pwx.dev.purestorage.com/job/pool-functional-0/8

### DIFF
--- a/tests/basic/pool_function_test.go
+++ b/tests/basic/pool_function_test.go
@@ -885,7 +885,43 @@ var _ = Describe("{StorageFullPoolExpansion}", func() {
 		_ = Inst().S.RemoveLabelOnNode(*selectedNode, k8s.NodeType)
 	})
 
-	It(stepLog, func() {
+	It("Expand pool with add-disk type after pool is down due to storage full", func() {
+		// https://portworx.testrail.net/index.php?/cases/view/51280
+		StartTorpedoTest("StorageFullPoolExpandAddDiskType", "Expand pool with add-disk type after pool is down due to storage full", nil, 51280)
+		Step("Prepare a full pool to expand", func() {
+			err = WaitForPoolOffline(*selectedNode)
+			log.FailOnError(err, "Timed out waiting to load a pool and bring node %s storage down", selectedNode.Name)
+			poolsStatus, err := Inst().V.GetNodePoolsStatus(*selectedNode)
+			log.FailOnError(err, "error getting pool status on node %s", selectedNode.Name)
+			for i, s := range poolsStatus {
+				if s == "Offline" {
+					poolIDToResize = i
+					poolToResize, err = GetStoragePoolByUUID(poolIDToResize)
+					log.FailOnError(err, "error getting pool with UUID [%s]", poolIDToResize)
+					break
+				}
+			}
+		})
+
+		Step("Expand the full pool in type add-disk", func() {
+			targetSizeGiB = (poolToResize.TotalSize / units.GiB) * 2
+			log.InfoD("Current Size of the pool %s is %d, trying to expand it to double the size", poolToResize.Uuid, poolToResize.TotalSize/units.GiB)
+			err = Inst().V.ExpandPool(poolToResize.Uuid, api.SdkStoragePool_RESIZE_TYPE_ADD_DISK, targetSizeGiB, true)
+			dash.VerifyFatal(err, nil, "Pool expansion init should be successful.")
+		})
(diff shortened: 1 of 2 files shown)


tests/basic/pool_function_test.go
```diff
@@ -859,3 +860,63 @@ var _ = Describe("{PoolExpandAddDiskInMaintenanceMode}", func() {
...
+	})
+
+	It("Expand pool with resize-disk type after pool is down due to storage full", func() {
+		// https://portworx.testrail.net/index.php?/cases/view/51280
+		StartTorpedoTest("StorageFullPoolExpandResizeType", "Expand pool with resize-disk type after pool is down due to storage full", nil, 51280)
+		Step("Prepare a full pool to expand", func() {
+			err = WaitForPoolOffline(*selectedNode)
+			log.FailOnError(err, fmt.Sprintf("Timed out waiting to load a pool and bring node %s storage down", selectedNode.Name))
+			poolsStatus, err := Inst().V.GetNodePoolsStatus(*selectedNode)
+			log.FailOnError(err, "error getting pool status on node %s", selectedNode.Name)
```
pp511, line 895: nit...Curious why fmt.Sprintf in one log and not the orher?

tests/basic/pool_function_test.go
```diff
@@ -859,3 +860,63 @@ var _ = Describe("{PoolExpandAddDiskInMaintenanceMode}", func() {
...
+
+	BeforeEach(func() {
+		Inst().AppList = []string{"fio-fastpath-repl1"}
+		contexts = ScheduleApplications("storagefull-resize")
+		appList = Inst().AppList
+	})
+
+	JustBeforeEach(func() {
+		selectedNode = GetNodeWithLeastSize()
+		_ = Inst().S.AddLabelOnNode(*selectedNode, k8s.NodeType, k8s.FastpathNodeType)
```
pp511, line 878: I am not aware. What happens with this label?

tests/basic/pool_function_test.go
```diff
@@ -859,3 +860,63 @@ var _ = Describe("{PoolExpandAddDiskInMaintenanceMode}", func() {
...
+					log.FailOnError(err, "error getting pool with UUID [%s]", poolIDToResize)
+					break
+				}
+			}
+		})
+
+		Step("Expand the full pool in type resize-disk", func() {
+			targetSizeGiB = (poolToResize.TotalSize / units.GiB) * 2
+			log.InfoD("Current Size of the pool %s is %d, trying to expand it to double the size", poolToResize.Uuid, poolToResize.TotalSize/units.GiB)
+			err = Inst().V.ExpandPool(poolToResize.Uuid, api.SdkStoragePool_RESIZE_TYPE_RESIZE_DISK, targetSizeGiB, true)
```
pp511, line 909: We are using repl1 and skipWaitForCleanVolumes is true. This way the code that gets triggered is not an actual customer scenario. Is it? Do customers use forced option option when doing the  expand? For repl1 volumes if pool needs to be put to maintenance then they will need to stop IOs and detach the volumes, put pool explicitly in maintenance mode before doing the pool expand and this won't be an online operation right? 

Instead why don't we test with a repl2 app which triggers an actual prod scenario?

tests/basic/pool_function_test.go
alicelyu-px: This makes the testing app(fio-fastpath) to write to a pool on this specific node

tests/basic/pool_function_test.go
alicelyu-px: This is a tradeoff we decided to take a few months ago. 
The functionality of `skipWaitForCleanVolumes` param is fairly simple and independent, we would test it working separately. In the meantime, for all other functional tests, we set it to true to avoid unnecessary complexity.

pp511 (approved):

PR #1884, opened 2023-11-15
PTX-21105 Fix/improve ProfileOnlyDiags test

nikolaypopov:
Fix/improve ProfileOnlyDiags test

* Improved overall code readability
* Simplified and improved some methods/functions
* Enforced Telemetry check to fail test, if test is expecting to validate diags on s3
* Fixed the part where we looking for `[{ProfileOnlyDiags}] Validating diag file [.] on s3` file by trimming space from `newDiags` string
* Increase retry interval to wait for s3 diags and make it a `const`
* Looking for diags in latest and previous folders (if exist), in same loop to save time, due to race condition of how diags can be potentially uploaded
* Renamed func `GetClusterID()` to `GetClusterUUID()` as this is what we getting there
* Renamed func `collectAsyncDiags` to `collectDiagsSdk` 
* Improved some logging and errors to be more verbose
* Removed some commented blocks of code
* Some other minor improvements of code, comments and messages

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -99,7 +99,7 @@ const (
 	pxVersionsConfigmapName                   = "px-versions"
 	pureKey                                   = "backend"
 	pureBlockValue                            = "pure_block"
-	clusterIDFile                             = "/etc/pwx/cluster_uuid"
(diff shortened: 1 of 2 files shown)


mborodin-px (approved):
LGTM

mborodin-px (approved):
LGTM

PR #1893, opened 2023-11-16
Assigning kube-system as debug pod namespace for non-px environments

mkoppal-px:
During node driver initialisation when we run torpedo on cloud clusters where the nodes are not accessible through SSH, we deploy a debug pod which will help us run commands on the node. This debug pod always assumes that Portworx is deployed and tries to get the Portworx namespace where we deploy the debug pod. But in case of non-px environments (which is needed for Px-Backup) there is no px installed and hence it kept failing with error
```
Error while creating debug daemonset. Err: an empty namespace may not be set during creation
```
Now we will deploy the debug pod in kube-system namespace if we dont find px namespace.

Closes #PB-4894

[Jenkins Run](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/basic-system-test/job/nonpx/job/ibm/job/s3/job/ibm-s3-direct-kdmp/50/)

### DIFF
--- a/drivers/node/ssh/ssh.go
+++ b/drivers/node/ssh/ssh.go
@@ -165,6 +165,10 @@ func (s *SSH) updateDriver() error {
 	if err != nil {
 		return err
 	}
+	if execPodNamespace == "" {
+		log.Warnf("Unable to find portworx namespace. Setting debug pod namespace to kube-system")
+		execPodNamespace = "kube-system"
+	}
 	s.execPodNamespace = execPodNamespace
 	log.InfoD("UpdateDriver: ssh driver's namespace (the 'portworx-service' namespace) is updated to [%s]", execPodNamespace)
 


kphalgun-px (approved):
LGTM

lsrinivas-pure:
> > @mkoppal-px can you run regression also once
> 
> I thought CBT jobs will take care of that

I think CBT uses direct ssh not debug pods

kshithijiyer-px (approved):
LGTM

mkoppal-px:
@lsrinivas-pure I the IBM pipeline with Px where we are deploying debug pod and it is passing - https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/basic-system-test/job/nonpx/job/ibm/job/s3/job/ibm-s3-direct-kdmp/51/

nikolaypopov (approved):

PR #1901, opened 2023-11-17
Fix torpedo pod logs getting printed on console

mkoppal-px:
We have been observing this from a very long time in long torpedo runs. The logs we get on the jenkins console is a result of `kubectl logs -f torpedo`
But for some reason this log stream exits after every 4 hours and the whole log is printed again.
The cumulative logs get printed on the console page and this is resulting in slowness while loading the jenkins console page and also crashing at times.
To fix this we will be getting only recent logs if `kubectl logs -f torpedo` exits during the first iteration by using `kubectl logs -f --since=1m -f torpedo`

Closes #

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -747,15 +747,45 @@ function describe_pod_then_exit {
   exit 1
 }
 
-for i in $(seq 1 900) ; do
+#for i in $(seq 1 900) ; do
+#  printf .
+#  state=`kubectl get pod torpedo | grep -v NAME | awk '{print $3}'`
+#  if [ "$state" == "Error" ]; then
+#    echo "Error: Torpedo finished with $state state"
+#    describe_pod_then_exit
+#  elif [ "$state" == "Running" ]; then
+#    echo ""
+#
+#    kubectl logs -f torpedo
+#  elif [ "$state" == "Completed" ]; then
+#    echo "Success: Torpedo finished with $state state"
+#    exit 0
+#  fi
+#
+#  sleep 1
+#done
+
+first_iteration=true
+
+for i in $(seq 1 900); do
   printf .
-  state=`kubectl get pod torpedo | grep -v NAME | awk '{print $3}'`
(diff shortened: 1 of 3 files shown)


deployments/deploy-ssh.sh
```diff
@@ -747,15 +747,25 @@ function describe_pod_then_exit {
...
-    echo ""
-    kubectl logs -f torpedo
+    # For the first iteration, display all logs. Later, only from 1 minute ago
+    if [ "$first_iteration" = true ]; then
+      echo "Logs from first iteration"
+      kubectl logs -f torpedo
+      first_iteration=false
+    else
+      echo -n "Logs from iteration: $i"
+      kubectl logs -f --since=1m -f torpedo
```
lsrinivas-pure, line 767: -f twice is required here?

lsrinivas-pure, line 767: is there any chance of missing logging in console due to time here?

mkoppal-px: No. We will not miss anything. For this reason I have kept an overlap of 1 min.

mkoppal-px: Fixed it

kphalgun-px (approved):
LGTM

mkoppal-px:
@lsrinivas-pure Successful jenkins run for full pipeline - https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/basic-system-test/job/px/job/s3/job/px-backup-system-test-vanilla/520/

lsrinivas-pure (approved):
LGTM

kshithijiyer-px (approved):
LGTM

kshithijiyer-px:
@mkoppal-px Should we do the same change in https://github.com/portworx/torpedo/blob/master/deployments/deploy-pxb-cloud.sh as well?

mkoppal-px:
> @mkoppal-px Should we do the same change in https://github.com/portworx/torpedo/blob/master/deployments/deploy-pxb-cloud.sh as well?

@kshithijiyer-px Did the same for other sh files

kshithijiyer-px (approved):
LGTM

PR #1911, opened 2023-11-21
PB-4968: Fix issue with `LicensingCountBeforeAndAfterBackupPodRestart`

kshithijiyer-px:
While running the IBM jobs we saw testcase `LicensingCountBeforeAndAfterBackupPodRestart` fail intermittently with the below error:
```
error: pods "quick-maintenance-repo-6f03bd2-28337041-778fs" not found, Description: Restart all the backup pods
```
This is mostly because the `quick-maintenance-repo-6f03bd2-28337041-778fs` job pod is a short lived job pod and is deleted in some cases between the line 1480 and 1518.

Closes #PB-4968

### DIFF
--- a/tests/backup/backup_helper.go
+++ b/tests/backup/backup_helper.go
@@ -1473,7 +1473,7 @@ func GetAllRestoresNonAdminCtx(ctx context.Context) ([]string, error) {
 }
 
 // DeletePodWithLabelInNamespace kills pod with the given label in the given namespace
-func DeletePodWithLabelInNamespace(namespace string, label map[string]string) error {
+func DeletePodWithLabelInNamespace(namespace string, label map[string]string, ignoreLabel bool) error {
 	var pods *corev1.PodList
 	var err error
 	// TODO: Revisit this function and remove the below code if not needed
@@ -1501,8 +1501,19 @@ func DeletePodWithLabelInNamespace(namespace string, label map[string]string) er
 	if err != nil {
 		return err
 	}
-
 	for _, pod := range pods.Items {
+		flag := false
+		if ignoreLabel {
+			for _, podlabel := range label {
+				if IsPresent(pod.GetLabels(), podlabel) {
+					flag = true
+					break
+				}
+			}
+		}
+		if flag {
+			continue
+		}
 		log.Infof("Deleting pod %s with label %v", pod.GetName(), label)
 		err = core.Instance().DeletePod(pod.GetName(), namespace, false)
 		if err != nil {
@@ -4517,7 +4528,7 @@ func DeletePodWhileBackupInProgress(ctx context.Context, orgId string, backupNam
 	if err != nil {
 		return err
 	}
-	err = DeletePodWithLabelInNamespace(namespace, label)
+	err = DeletePodWithLabelInNamespace(namespace, label, false)
 	if err != nil {
 		return err
 	}
@@ -4552,7 +4563,7 @@ func DeletePodWhileRestoreInProgress(ctx context.Context, orgId string, restoreN
 	if err != nil {
 		return err
 	}
-	err = DeletePodWithLabelInNamespace(namespace, label)
+	err = DeletePodWithLabelInNamespace(namespace, label, false)
 	if err != nil {
 		return err
 	}
--- a/tests/backup/backup_license_test.go
+++ b/tests/backup/backup_license_test.go
@@ -315,7 +315,6 @@ var _ = Describe("{LicensingCountWithNodeLabelledBeforeClusterAddition}", func()
 	})
 })
 
-// LicensingCountBeforeAndAfterBackupPodRestart verifies the license count before and after the backup pods restart
 var _ = Describe("{LicensingCountBeforeAndAfterBackupPodRestart}", func() {
 	var (
 		pxbNamespace                  string
@@ -378,7 +377,9 @@ var _ = Describe("{LicensingCountBeforeAndAfterBackupPodRestart}", func() {
 		Step("Restart all the backup pod and wait for it to come up", func() {
 			pxbNamespace, err = backup.GetPxBackupNamespace()
 			log.FailOnError(err, "Getting px-backup namespace")
-			err = DeletePodWithLabelInNamespace(pxbNamespace, nil)
+			var labelselector = make(map[string]string)
+			labelselector["kdmp.portworx.com/driver-name"] = "kopiamaintenance"
+			err = DeletePodWithLabelInNamespace(pxbNamespace, labelselector, true)
 			dash.VerifyFatal(err, nil, "Restart all the backup pods")
 			log.InfoD("Validate if all the backup pods are up")
 			err = ValidateAllPodsInPxBackupNamespace()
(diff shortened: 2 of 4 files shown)


tests/backup/backup_helper.go
```diff
@@ -1497,12 +1502,34 @@ func DeletePodWithLabelInNamespace(namespace string, label map[string]string) er
...
+						break
+					}
+				}
+				if flag {
+					break
+				}
+			}
+			if flag {
+				continue
+			}
```
kphalgun-px, line 1530:
```go
flag := false
for key, podlabel := range pod.GetLabels() {
    if podlabel2, exists := label[key]; exists && podlabel == podlabel2 {
        flag = true
        break
    }
}
if flag {
    continue
}
```
?

tests/backup/backup_helper.go
```diff
@@ -1497,12 +1502,34 @@ func DeletePodWithLabelInNamespace(namespace string, label map[string]string) er
...
+		pods, err = core.Instance().GetPods(namespace, nolabel)
+	} else {
+		pods, err = core.Instance().GetPods(namespace, label)
+	}
 	if err != nil {
 		return err
 	}
-
 	for _, pod := range pods.Items {
+		flag := false
```
kphalgun-px, line 1515: skipPod instead of flag?

tests/backup/backup_helper.go
```diff
@@ -1473,12 +1473,17 @@ func GetAllRestoresNonAdminCtx(ctx context.Context) ([]string, error) {
 }
 
 // DeletePodWithLabelInNamespace kills pod with the given label in the given namespace
-func DeletePodWithLabelInNamespace(namespace string, label map[string]string) error {
+func DeletePodWithLabelInNamespace(namespace string, label map[string]string, ignoreLabel bool) error {
```
kphalgun-px, line 1476:
```go
if ignoreLabel {
    label = make(map[string]string)
}
```
?

tests/backup/backup_helper.go
kphalgun-px: if possible log pod name while skipping?

kphalgun-px (approved):
LGTM

tests/backup/backup_helper.go
```diff
@@ -1497,12 +1502,34 @@ func DeletePodWithLabelInNamespace(namespace string, label map[string]string) er
 	}
 
 	// fetch the newest set of pods post wait for pods to come up
-	pods, err = core.Instance().GetPods(namespace, label)
+	if ignoreLabel {
```
atrivedi-px, line 1505: Add this as part of **podList** instead of a separate call

tests/backup/backup_helper.go
```diff
@@ -1473,12 +1473,17 @@ func GetAllRestoresNonAdminCtx(ctx context.Context) ([]string, error) {
...
 
 // DeletePodWithLabelInNamespace kills pod with the given label in the given namespace
-func DeletePodWithLabelInNamespace(namespace string, label map[string]string) error {
+func DeletePodWithLabelInNamespace(namespace string, label map[string]string, ignoreLabel bool) error {
 	var pods *corev1.PodList
 	var err error
 	// TODO: Revisit this function and remove the below code if not needed
 	podList := func() (interface{}, bool, error) {
-		pods, err = core.Instance().GetPods(namespace, label)
+		if ignoreLabel {
```
atrivedi-px, line 1481: Instead of making this if check at multiple places could you calculate the podList based on ignoreLabel and use it later

tests/backup/backup_helper.go
```diff
@@ -1497,12 +1502,34 @@ func DeletePodWithLabelInNamespace(namespace string, label map[string]string) er
...
+	} else {
+		pods, err = core.Instance().GetPods(namespace, label)
+	}
 	if err != nil {
 		return err
 	}
-
 	for _, pod := range pods.Items {
+		flag := false
+		if ignoreLabel {
```
atrivedi-px, line 1516: Could you implement this logic as part of podList and keep the deletion for all pods fetched using same method irrespective of whether ignoreLabel is set or not

tests/backup/backup_helper.go
kshithijiyer-px: Based on offline discussion we decided to not go with this as it will be more time complex as we would need to run through the pod list twice hence. We'll revisit this code as the logic of the function isn't very clear. When we revist the code we'll fix the logic. 

For now going with the current logic until we revisit again

tests/backup/backup_helper.go
kshithijiyer-px: Make sense will fix it in the next logic

tests/backup/backup_helper.go
kshithijiyer-px: Yes this looks better we'll cleanup

tests/backup/backup_helper.go
kshithijiyer-px: Can't do this as we still need label at the end to do filter

atrivedi-px (approved):
As per offline discussion we will revisit the logic later. LGTM for now

tests/backup/backup_helper.go
kshithijiyer-px: We need the label later to not delete the other pods

tests/backup/backup_helper.go
kshithijiyer-px: Done

kshithijiyer-px:
Run post changes https://aetos.pwx.purestorage.com/resultSet/testSetID/449688

mkoppal-px (changes requested):
I am not clear here. We are calling this function `DeletePodWithLabelInNamespace` which means this function expects namespace and label. But if we are ignoring label here, then we might as well have a new function called `DeleteAllPodsInNamespace`

kshithijiyer-px:
> I am not clear here. We are calling this function `DeletePodWithLabelInNamespace` which means this function expects namespace and label. But if we are ignoring label here, then we might as well have a new function called `DeleteAllPodsInNamespace`

The reason why I went to update the same existing code is because the current code deletes all pods or all pods which have the mentioned label and I have updated it to run and delete pods which don't have the label. If we create a new function it'll create be redundant code here.

tests/backup/backup_helper.go
mkoppal-px, line 1476: Include/Exclude labels

kshithijiyer-px: Done

tests/backup/backup_helper.go
```diff
@@ -1472,13 +1472,18 @@ func GetAllRestoresNonAdminCtx(ctx context.Context) ([]string, error) {
 	return restoreNames, nil
 }
 
-// DeletePodWithLabelInNamespace kills pod with the given label in the given namespace
-func DeletePodWithLabelInNamespace(namespace string, label map[string]string) error {
+// DeletePodWithWithoutLabelInNamespace kills pod with the given label in the given namespace
```
mkoppal-px, line 1475: Change the doc string accordingly

kshithijiyer-px: Done

kshithijiyer-px:
Successful run after rename https://aetos.pwx.purestorage.com/resultSet/testSetID/449931

mkoppal-px (approved):
LGTM

PR #1920, opened 2023-11-23
Fixing autopilot cleanup

aghodke1312:
Test: 
Disabled autopilot from STC. 
`~/aetos on  master! ⌚ 15:54:36$  kfunc get pods | grep -i autopilot
users-aniket-autopilotall-px-int-5ngk9                        2/2     Running   0             21h
users-aniket-autopilotall-px-int-8nb69                        2/2     Running   0             21h
users-aniket-autopilotall-px-int-btpf8                        2/2     Running   0             21h
users-aniket-autopilotall-px-int-bvf9g                        2/2     Running   0             21h
users-aniket-autopilotall-px-int-pchnz                        2/2     Running   0             21h`

Log says no rule found:

`
2023-11-23 17:34:29 +0000:[INFO] [{SetupTeardown}] [k8s.(*K8s).ValidateVolumes:#3840] - [nginx] Autopilot is not enabled, skipping PVC: px-nginx-pvc-enc-legacy-shared size validation
` 
Test passed: https://aetos.pwx.purestorage.com/resultSet/testSetID/442020

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -3153,16 +3153,20 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
 	var podList []corev1.Pod
 
 	allRules, err := k8sAutopilot.ListAutopilotRules()
-	if err != nil {
-		return err
-	}
-
-	for _, rule := range allRules.Items {
-		log.InfoD("Cleaning up autopilot rule: %s", rule.Name)
-		err := k8sAutopilot.DeleteAutopilotRule(rule.Name)
-		if err != nil {
-			return err
+	if err == nil {
+		if len(allRules.Items) > 0 {
+			for _, rule := range allRules.Items {
+				log.InfoD("Cleaning up autopilot rule: %s", rule.Name)
+				err := k8sAutopilot.DeleteAutopilotRule(rule.Name)
+				if err != nil {
+					return err
+				}
+			}
+		} else {
+			log.InfoD("No autopilot rules found, skipping cleanup")
 		}
+	} else {
+		log.InfoD("Error in getting autopilot rules, skipping cleanup: %v", err)
 	}
 
 	// destruction of CustomResourceObjects must most likely be done *first*,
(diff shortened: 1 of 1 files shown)


CharudathGopal (approved):

drivers/scheduler/k8s/k8s.go
```diff
@@ -3153,16 +3153,16 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
...
-		return err
-	}
-
-	for _, rule := range allRules.Items {
-		log.InfoD("Cleaning up autopilot rule: %s", rule.Name)
-		err := k8sAutopilot.DeleteAutopilotRule(rule.Name)
-		if err != nil {
-			return err
+	if err == nil {
+		for _, rule := range allRules.Items {
```
nikolaypopov, line 3157: Error could be nil, but rules could also be empty, I would check if its empty before going through loop and maybe even say no rules were found if it is empty

drivers/scheduler/k8s/k8s.go
```diff
@@ -3153,16 +3153,20 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
...
+				if err != nil {
+					return err
+				}
+			}
+		} else {
+			log.InfoD("No autopilot rules found, skipping cleanup")
 		}
+	} else {
+		log.InfoD("Error in getting autopilot rules, skipping cleanup: %v", err)
 	}
```
nikolaypopov, line 3170:
```suggestion
	if err != nil {
		return fmt.Errorf("failed to List autopilot rules, Err: %v")
	}
	
	if len(allRules.Items) > 0 {
		for _, rule := range allRules.Items {
			log.InfoD("Cleaning up autopilot rule: %s", rule.Name)
			if err := k8sAutopilot.DeleteAutopilotRule(rule.Name): err != nil {
				return err
			}
		}
	} else{
		log.InfoD("No autopilot rules found, skipping cleanup")
	}
```

nikolaypopov (approved):

drivers/scheduler/k8s/k8s.go
```diff
@@ -3153,16 +3153,20 @@ func (k *K8s) Destroy(ctx *scheduler.Context, opts map[string]bool) error {
...
-	if err != nil {
-		return err
-	}
-
-	for _, rule := range allRules.Items {
-		log.InfoD("Cleaning up autopilot rule: %s", rule.Name)
-		err := k8sAutopilot.DeleteAutopilotRule(rule.Name)
-		if err != nil {
-			return err
+	if err == nil {
```
mkoppal-px, line 3156: Few comments here:

- This function `Destroy` is meant only to delete the kubernetes objects which got deployed during the app creation
- As you can see in all the lines below, we are looping over SpecList and destroying them by explicitly calling the relevant destroy function
- Since autopilot rule is not getting created as part of any spec, it should be a separate function altogether. I would suggest you create a function explicitly in `common.go` to delete the autopilot rules and call it in the test case wherever needed
- The whole logic of deleting autopilot rule does not fit here because we are not using the receiver type `k *K8s` neither are we using the input arguments `(ctx *scheduler.Context, opts map[string]bool)`

mkoppal-px (changes requested):
Please consider creating a separate function for autopilot rule deletion

PR #1943, opened 2023-12-04
DS-0000: Added fix for killdbmasternodetest

jkagliwal-px:
Fix to run test only for postgres

### DIFF
--- a/tests/pds/pds_resiliency_test.go
+++ b/tests/pds/pds_resiliency_test.go
@@ -1681,7 +1681,7 @@ var _ = Describe("{KillDbMasterNodeDuringStorageResize}", func() {
 
 		Step("Deploy Data Services", func() {
 			for _, ds := range params.DataServiceToTest {
-				if ds.Name == postgresql {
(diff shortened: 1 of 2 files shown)


tests/pds/pds_resiliency_test.go
```diff
@@ -1681,12 +1681,17 @@ var _ = Describe("{KillDbMasterNodeDuringStorageResize}", func() {
 
 		Step("Deploy Data Services", func() {
 			for _, ds := range params.DataServiceToTest {
-				Step("Deploy and validate data service", func() {
-					isDeploymentsDeleted = false
-					deployment, _, _, err = DeployandValidateDataServices(ds, params.InfraToTest.Namespace, tenantID, projectID)
-					log.FailOnError(err, "Error while deploying data services")
-					deployments[ds] = deployment
-				})
+				if ds.Name == postgresql {
```
Madanagopal19, line 1684: it also runs for MYSQL ?

Madanagopal19 (changes requested):

PR #1979, opened 2023-12-16
AllNSBackupWithIncludeNewNSOption: Suspending schedule before deleting it

snigdha-px:
Currently we are deleting schedule before suspending the schedule with DeleteBackups: true, flag
So backup is in creating state while deleting the schedule. Hence its in DeletePending
Code change: Suspend the schedule before deleting it

Closes #https://portworx.atlassian.net/browse/PB-5220

log : https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/3806/consoleFull


tests/backup/backup_restore_basic_test.go
```diff
@@ -954,10 +954,13 @@ var _ = Describe("{AllNSBackupWithIncludeNewNSOption}", func() {
...
 		log.InfoD("Deleting deployed namespaces - %v", allAppNamespaces)
 		ValidateAndDestroy(allContexts, opts)
-
 		err := SetSourceKubeConfig()
 		log.FailOnError(err, "failed to switch context to source cluster")
-
+		// TO DO: Remove suspendBackupSchedule after PB-5221 is fixed
+		ctx, err := backup.GetAdminCtxFromSecret()
+		log.FailOnError(err, "Fetching px-central-admin ctx")
+		err = suspendBackupSchedule(scheduleName, schedulePolicyName, orgID, ctx)
```
mkoppal-px, line 962: Can we add suspend inside the delete schedule function itself. So that we are making sure the schedule is suspended before deletion? Later we can remove it from that function.
Now we don't know in which other test cases we are deleting the schedule without suspending it.

mkoppal-px (changes requested):
Commented

tests/backup/backup_restore_basic_test.go
snigdha-px: I also thought so , but I tried doing it manually, if its already in suspended state, there will be no option  of suspend again , only resume option
So will api also not throw error if the testcase has already suspended it?
Also its only this testcase in last many runs that has  been failing because of this issue

mkoppal-px: Then we can add a check before we suspend it. If it is suspended then we will directly delete. If not, we will suspend and delete.
If someone else uses delete in future they should not hit this issue. That's the intention here.

If you feel we should not be changing the existing delete schedule function, then you can create a new function called SuspendAndDeleteSchedule and use that in the test case. It will be a useful function.

snigdha-px: Done

ak-px (approved):
LGTM

tests/backup/backup_helper.go
```diff
@@ -6457,3 +6459,61 @@ func validateCRCleanup(resourceInterface interface{},
...
+			Name: schedulePolicyName,
+			Uid:  schPolicyUID,
+		},
+	}
+	log.InfoD("Suspending backup schedule %s", backupScheduleName)
+	_, err = backupDriver.UpdateBackupSchedule(ctx, bkpScheduleSuspendRequest)
+	if err != nil {
+		return err
+	}
+	log.InfoD("Deleting backup schedule %s", backupScheduleName)
```
mkoppal-px, line 6487: Make sure we are checking for suspended status before moving to delete

mkoppal-px (approved):
LGTM except for one comment

tests/backup/backup_helper.go
snigdha-px: done

snigdha-px:
> LGTM except for one comment

Log after change :https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/3862/consoleFull

PR #2002, opened 2024-01-11
PTX-21279 Add more verbosity when validating mounts and devices

nikolaypopov:
Add more verbosity when validating mounts and devices. There was a bug I was chasing and this helped me get more info and understand the issue


mborodin-px (approved):
LGTM

PR #2004, opened 2024-01-12
[PA-1538] New BYOC

kphalgun-px:
This PR re-implements Dockerfile and SSH deployment to reduce build time and enhance flexibility. It also includes a script for collecting test reports and data from /var/cores using a debug pod.

Closes #PA-1538

I have a implemented a Jenkins pipeline on top of these changes and here is an example build:

https://jenkins.pwx.dev.purestorage.com/job/Users/job/krishna/job/tp-pxe-byoc/1140/

To make it easy for review, I have modified the Dockerfile and deploy-ssh.sh file in place. I will create new files during the merging process.


mkoppal-px:
Code looks good. But create separate files so that it does not affect the current scripts.

PR #2018, opened 2024-01-19
PTX-21474 Install gcompat package to unblock OCP 4.14 CLI

nikolaypopov:
Install gcompat package to unblock OCP 4.14 CLI


mborodin-px (approved):
LGTM

PR #2020, opened 2024-01-19
PWX-35639:Adding test for fuse driver failure

lsrinivas-pure:
Adding test for fuse driver failure

Closes #PTX-21805,PTX-21950

Result : https://aetos.pwx.purestorage.com/resultSet/testSetID/511842

### DIFF
--- a/tests/basic/volume_ops_test.go
+++ b/tests/basic/volume_ops_test.go
@@ -2780,7 +2780,7 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
 			contexts = make([]*scheduler.Context, 0)
 
 			for i := 0; i < Inst().GlobalScaleFactor; i++ {
-				contexts = append(contexts, ScheduleApplications(fmt.Sprintf("purevolumestest-%d", i))...)
+				contexts = append(contexts, ScheduleApplications(fmt.Sprintf("nfsproxytest-%d", i))...)
 			}
 
 			for _, ctx := range contexts {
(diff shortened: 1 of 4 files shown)


tests/basic/volume_ops_test.go
```diff
@@ -2843,6 +2843,111 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
...
+						StopVolDriverAndWait([]node.Node{*attachedNode})
+						log.Infof("waiting for 1 min before starting PX for volume corordinator to failover")
+						time.Sleep(1 * time.Second)
+						StartVolDriverAndWait([]node.Node{*attachedNode})
+						newAttachedNode, err := Inst().V.GetNodeForVolume(vol, 1*time.Minute, 5*time.Second)
+						log.FailOnError(err, "error getting attached node for volume [%s]", vol.Name)
+						log.Infof("vol [%s] is now attached to node [%s]", vol.Name, newAttachedNode.Name)
+						ValidateContext(ctx)
+						for _, eachNode := range node.GetStorageDriverNodes() {
+							status, err := IsPxRunningOnNode(&eachNode)
```
gpal1, line 2915: After all the iterations are completed, verify that the PX is NOT stuck in an initialization loop, i.e. the PX should NOT keep transitioning between 'PX Down' and 'PX Initialization state

tests/basic/volume_ops_test.go
```diff
@@ -2843,6 +2843,111 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
...
+				vols, err := Inst().S.GetVolumes(ctx)
+				log.FailOnError(err, "failed to get volumes from the contexts")
+
+				for _, vol := range vols {
+					log.InfoD("Initiating failover iteration for volume [%s] of app [%s]", vol.Name, ctx.App.Key)
+					for i := 1; i <= 5; i++ {
+						log.Infof("Iteration: #%d for volume [%s]", i, vol.Name)
+						attachedNode, err := Inst().V.GetNodeForVolume(vol, 1*time.Minute, 5*time.Second)
+						log.FailOnError(err, "error getting attached node for volume [%s]", vol.Name)
+						log.Infof("vol [%s] is attached to node [%s]", vol.Name, attachedNode.Name)
```
Santhoshairody, line 2905: How do you make sure, active IO happening on the device. Options would be 
root@ip-10-13-232-218:~# iostat -xm 5 | grep pxd585034700205937734
pxd/pxd585034700205937734    0.00      0.00     0.00  20.01    2.00     5.47    0.74      0.00     0.35  32.03    6.15     6.40    0.00      0.00     0.00   0.00    0.00     0.00    0.00   0.09


pxd/pxd585034700205937734    0.00      0.00     0.00   0.00    0.00     0.00    5.40      0.03     2.60  32.50    0.63     5.63    0.00      0.00     0.00   0.00    0.00     0.00    0.00   0.56
pxd/pxd585034700205937734    0.00      0.00     0.00   0.00    0.00     0.00    1.60      0.01     1.60  50.00    0.38     8.00    0.00      0.00     0.00   0.00    0.00     0.00    0.00   0.16

OR 
root@ip-10-13-232-218:~# cat /sys/block/pxd\!pxd585034700205937734/inflight 
       0        0

If there is any other way in PX-Test  you can use that as well.

tests/basic/volume_ops_test.go
```diff
@@ -2843,6 +2843,111 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
 	})
 })
 
+var _ = Describe("{SharedVolFuseTest}", func() {
+	/*
+
+				https://portworx.atlassian.net/browse/PWX-35639
+			   https://portworx.atlassian.net/browse/PTX-21805
```
Santhoshairody, line 2850: Can we add testrail id in the code
	var testrailID = 12133434
	// testrailID corresponds to: https://portworx.testrail.net/index.php?/cases/view/35259
	var runID int
	JustBeforeEach(func() {
		StartTorpedoTest("VolumeDriverDown", "Validate volume driver down", nil, testrailID)
	})

tests/basic/volume_ops_test.go
```diff
@@ -2843,6 +2843,111 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
...
+				contexts = append(contexts, ScheduleApplications(fmt.Sprintf("pxfusetest-%d", i))...)
+			}
+
+			for _, ctx := range contexts {
+				log.InfoD("Validating application [%s]", ctx.App.Key)
+				ValidateContext(ctx)
+			}
+		})
+
+		stepLog = "restart PX on coordinator nodes"
```
Santhoshairody, line 2891: IMO, this test should be time based, until time expires, test should run in a loop. Can you modify the test accordingly?

Santhoshairody: Also, bug QA notes say it is on different kernel versions, make sure we are running this test on different kernel versions
< 5.13
>= 5.13 and < 5.15.25
>= 5.15.25

tests/basic/volume_ops_test.go
```diff
@@ -2843,6 +2843,111 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
...
+
+				https://portworx.atlassian.net/browse/PWX-35639
+			   https://portworx.atlassian.net/browse/PTX-21805
+
+
+		       	1. Deploy apps  with mix of sv4 service and sv4  using k8s
+		  		2.Get the list of coordinator nodes have sv4 volume attached
+				3.Stop/Start PX on the coordinator node
+				4.Check the failover has happened i.e. we  have new  coordinator node assigned after step 3
+		   		5.Repeat this in a loop for 10 iterations
```
Rohit-PX, line 2857: The comment says the loop repeats for 10 iterations but on line  2901 it's doing only 5 iterations.

tests/basic/volume_ops_test.go
```diff
@@ -2843,6 +2843,111 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
...
+		Step(stepLog, func() {
+			log.InfoD(stepLog)
+
+			for _, ctx := range contexts {
+				vols, err := Inst().S.GetVolumes(ctx)
+				log.FailOnError(err, "failed to get volumes from the contexts")
+
+				for _, vol := range vols {
+					log.InfoD("Initiating failover iteration for volume [%s] of app [%s]", vol.Name, ctx.App.Key)
+					for i := 1; i <= 5; i++ {
```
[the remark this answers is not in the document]
lsrinivas-pure: Updated the iteration count to 10 and changed the logic to restart px on every node  
number of volume can be increase by using GlobalScaleFactor option

tests/basic/volume_ops_test.go
lsrinivas-pure: IsPxRunningOnNode will validate the PX and returns error if it is not in running state

tests/basic/volume_ops_test.go
lsrinivas-pure: @Santhoshairody 
Currently we are running vdbench and fio apps which does continuous IOs. Also since we are restarting all nodes now it should cover the data path. I think adding this check will introduce unnecessary complexity to the test and make test slower. Please let me know you thoughts

tests/basic/volume_ops_test.go
lsrinivas-pure: fixed

tests/basic/volume_ops_test.go
lsrinivas-pure: fixed

tests/basic/volume_ops_test.go
```diff
@@ -2843,6 +2843,111 @@ var _ = Describe("{NFSProxyVolumeValidation}", func() {
...
+		stepLog = "create sharedv4 and sharedv4_svc apps "
+		Step(stepLog, func() {
+			log.InfoD(stepLog)
+			appList := Inst().AppList
+
+			defer func() {
+				Inst().AppList = appList
+			}()
+
+			Inst().AppList = []string{"vdbench-sv4-svc", "fio-sharedv4"}
```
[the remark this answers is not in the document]
lsrinivas-pure: @nikolaypopov this test is more specific to sharedv4 and sharedv4_svc volumes, hence defaulting the apps to "vdbench-sv4-svc", "fio-sharedv4".

Here Im taking the user passed app-list to variable appList, overriding  Inst().AppList and reverting to original value once test is done.

tests/basic/volume_ops_test.go
lsrinivas-pure: > Also, bug QA notes say it is on different kernel versions, make sure we are running this test on different kernel versions < 5.13
> 
> > = 5.13 and < 5.15.25
> > = 5.15.25

yes, running this test in different platforms should cover it.

lsrinivas-pure: > IMO, this test should be time based, until time expires, test should run in a loop. Can you modify the test accordingly?

spoke to @gpal1 and he mentioned , he is able to reproduce the issue by restarting all nodes hence changed the logic to reboot all nodes for 10 iterations each

tests/basic/volume_ops_test.go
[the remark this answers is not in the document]
nikolaypopov: Should we just go through the list of passed apps and run only sharedv4 apps and skip any other none suitable for this test apps, this way you don't have to default to only these 2 apps and you can pass more sharedv4 apps, in case we add them in the future? It looks like if we need to add/delete apps, it becomes a code change, rather than a simple app list edit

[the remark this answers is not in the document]
lsrinivas-pure: fixed it

dbhatnagar-px (approved):
Lgtm

PR #2022, opened 2024-01-19
PTX-21497 Fix for PKS diag directory

nikolaypopov:
Fix for PKS diag directory


tests/basic/telemetry_test.go
```diff
@@ -71,6 +72,22 @@ func TelemetryEnabled(currNode node.Node) bool {
...
 
+// IsPKS checks if this is PKS cluster based on annotation in the StorageCluster
+func IsPKS() bool {
+	// Check if this is PX Operator based install and see if this PKS annotation is set to true
+	log.Info("Checking if this is PX Operator based install...")
+	isOpBased, _ := Inst().V.IsOperatorBasedInstall()
+	if isOpBased {
+		log.Info("This is PX Operator based install, checking if StorageCluster has PKS annotation set to true...")
+		spec, err := Inst().V.GetDriver()
+		Expect(err).ToNot(HaveOccurred())
```
lsrinivas-pure, line 83: please use log.FailOnError

tests/basic/telemetry_test.go
```diff
@@ -71,6 +72,22 @@ func TelemetryEnabled(currNode node.Node) bool {
 	return status
 }
 
+// IsPKS checks if this is PKS cluster based on annotation in the StorageCluster
+func IsPKS() bool {
+	// Check if this is PX Operator based install and see if this PKS annotation is set to true
+	log.Info("Checking if this is PX Operator based install...")
+	isOpBased, _ := Inst().V.IsOperatorBasedInstall()
```
jainsmit, line 79: Better to check for error here because `isOpBased` could be nil in error condition and next if condition can fail with NPE..

nikolaypopov: `isOpBased` will never return `nil`, its a type `bool`, so it will always be either `true` or `false`, which is what I need here, see https://github.com/portworx/torpedo/blob/b74cb656ee8511c889391ac6f70e71afb368bf86/drivers/volume/portworx/portworx.go#L5032

jainsmit (approved):
LGTM !!

mborodin-px (approved):
LGTM

PR #2027, opened 2024-01-24
PWX-34920: Supported the VolumeUpdateForAttachedNode for Pure DA volumes

Pure-AdamuKaapan:
As part of the ongoing initiative to support Pure Direct Access volumes with the standard torpedo tests (instead of separate custom ones), this PR adds support for the `VolumeUpdateForAttachedNode` test. Specifically, if a Pure Direct Access volume is found, instead of changing the repl factor up and down, it instead validates it fails with the correct error and then moves on.

PWX-34920

**Testing notes**:
Validated this works using the `elasticsearch` app (FA) and `nginx-without-enc` app (FB).

### DIFF
--- a/tests/basic/volume_ops_test.go
+++ b/tests/basic/volume_ops_test.go
@@ -3,6 +3,13 @@ package tests
 import (
 	"errors"
 	"fmt"
+	"math"
+	"math/rand"
+	"reflect"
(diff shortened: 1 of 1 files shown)


alicelyu-px (approved):
lgtm

PR #2031, opened 2024-01-24
PWX-34921: Supports the VolumeDriverDown test for Pure volumes

Pure-AdamuKaapan:
Ensures the VolumeDriverDown test supports Pure Direct Access volumes as well as normal pxd volumes. This test normally runs some very px-specific commands, most of which does not apply to DA volumes and we instead have different invariants we want to test.

For Pure volumes, this test now validates that **IOs continue while PX is down**. It does this by repeatedly ensuring the FS is usable by touching files in the data directory of all pods, until the test is complete. If any of these touches fail, it will be reported at the end of the test.

PWX-34921

**Testing notes**:
Tested this using the `elasticsearch` and `nginx-without-enc` apps (FADA and FBDA respectively). Also ensured the same still works for pxd volumes (the original test).

Passing jenkins run: https://jenkins.pwx.dev.purestorage.com/job/Users/job/akrpan/job/akrpan-FADA/9/

### DIFF
--- a/tests/basic/misc_test.go
+++ b/tests/basic/misc_test.go
@@ -2,14 +2,19 @@ package tests
 
 import (
 	"fmt"
-	"github.com/portworx/torpedo/drivers/volume"
 	"math/rand"
+	"path"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/portworx/torpedo/drivers/volume"
+	"github.com/portworx/torpedo/drivers/volume/portworx"
+
 	opsapi "github.com/libopenstorage/openstorage/api"
 	"github.com/portworx/torpedo/pkg/log"
+	"github.com/portworx/torpedo/pkg/osutils"
+	"github.com/portworx/torpedo/pkg/pureutils"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/portworx/sched-ops/k8s/apps"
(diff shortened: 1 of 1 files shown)


tests/basic/misc_test.go
```diff
@@ -105,6 +165,22 @@ var _ = Describe("{VolumeDriverDown}", func() {
...
+				var err error
+				select {
+				case err = <-pureErrOutChan:
+					log.FailOnError(err, "Error writing to Pure volume")
+				default:
+					log.Infof("No errors found in error channel for Pure volume write validation")
+				}
+				// Then, close all the routines out so they stop writing
+				pureStopWriteRoutine = true
+				return
```
alicelyu-px, line 181: Do we not need line 184 in case of FADA?

alicelyu-px (approved):
lgtm. Could we share a link to the jenkins job the runs this test to success? 
Would be helpful for future reference. Thanks in advance!

tests/basic/misc_test.go
Pure-AdamuKaapan: We do not - that runs tests that are specific to pxd volumes (it runs commands like pxctl sv md5sum or something like that?)

Pure-AdamuKaapan:
Ah, I had run it on my own cluster without using a Jenkins run. Yes, I'll try to do that (will likely combine this and my next PR then)

tests/basic/misc_test.go
```diff
@@ -232,6 +321,11 @@ var _ = Describe("{VolumeDriverCrash}", func() {
 			}
 		})
 
+		if pureCleanupFunction != nil {
```
alicelyu-px, line 324: nit: why is it necessary to introduce this pureCleanupFunction variable?
For readability purpose I guess we could introduce a boolean like `isDirectAccessVolume`.. not seeing clear benefit of `pureCleanupFunction` here since looks like Inst().V.String() == portworx.PureDriverName is equivalent to pureCleanupFunction != nil and is arguably easier to follow.

tests/basic/misc_test.go
```diff
@@ -75,6 +148,11 @@ var _ = Describe("{VolumeDriverDown}", func() {
 
 		ValidateApplications(contexts)
 
+		var pureCleanupFunction func()
+		if Inst().V.String() == portworx.PureDriverName {
+			pureCleanupFunction = StartPureBackgroundWriteRoutines()
```
alicelyu-px, line 153: I guess this is the same readability nitpick: the variable name "cleanupFunction" seems not quite aligned with what the function StartPureBackgroundWriteRoutines() does? The purpose of pureCleanupFunction isn't very obvious to me

Pure-AdamuKaapan: `pureCleanupFunction` is basically what stops the goroutines and checks for errors. So the "start" function returns that cleanup function so we can then stop everything, without needing a global state variable

alicelyu-px: hmm I found it a little unintuitive cuz `pureCleanupFunction()` doesn't do any "cleanup" work.

alicelyu-px: This question is really for my education: what are some reasons we prefer defining the procedure to stop writing as a function returned from StartPureBackgroundWriteRoutines() over using a global state variable?

Pure-AdamuKaapan: Good point on the name, I guess "cleanup" for me was that it was "cleaning up" (stopping) the routines. `stopBackgroundWrites` would be more clear, I agree.

In my opinion, I opted for this because there's multiple tests using the same logic. If it was just one test, I was going to keep it as a boolean, but since both VDDown and VDCrash will use this function, I was concerned about bugs from not resetting the state properly. This way, we can just use local variables and closure them, so that they never leave the context of the test.

PR #2035, opened 2024-01-26
PTX-21291 Update libopenstorage/operator vendor package and go version

nikolaypopov:
Update libopenstorage/operator vendor package and go version

NOTE: One more major change is we will be using `"github.com/libopenstorage/operator/api/px"` instead of `"github.com/portworx/torpedo/porx/px/api"` now to fix the vendor issue, until the porx APIs will be changed to public package


mborodin-px (approved):
LGTM

PR #2037, opened 2024-01-29
Longevity Support for PxBackup

atrivedi-px:
Longevity support to create backup

Closes #PB-5368, PB-5375


tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,1567 @@
...
+		CreatePxBackup: TriggerCreateBackup,
+	}
+	//Creating a distinct trigger to make sure email triggers at regular intervals
+	emailTriggerFunction = map[string]func(){
+		EmailReporter: TriggerEmailReporter,
+	}
+
+	BeforeEach(func() {
+		if !populateDone {
+			StartPxBackupTorpedoTest("BackupLongevityTest",
```
kphalgun-px, line 69: If I recall correctly, adding the tag "longevity": "true" is needed for Aetos to skip performing cleanup.

*BackupLongevity~Test~

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,1567 @@
...
+	It("has to schedule app and introduce test triggers", func() {
+		log.InfoD("schedule apps and start test triggers")
+		watchLog := fmt.Sprintf("Start watch on K8S configMap [%s/%s]",
+			configMapNS, testTriggersConfigMap)
+
+		Step(watchLog, func() {
+			log.InfoD(watchLog)
+			err := watchConfigMap()
+			if err != nil {
+				log.Fatalf(fmt.Sprintf("%v", err))
```
kphalgun-px, line 86: better fail if there is an error

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,1567 @@
...
+	err = core.Instance().WatchConfigMap(cm, fn)
+	if err != nil {
+		return fmt.Errorf("Failed to watch on config map: %s due to: %v", testTriggersConfigMap, err)
+	}
+	return nil
+}
+
+func populateDisruptiveTriggers() {
+	disruptiveTriggers = map[string]bool{
+		HAIncrease:                      false,
```
kphalgun-px, line 270: Need PxE events?

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,1567 @@
...
+		if err := Inst().V.WaitDriverUpOnNode(n, nodeUpTimeout); err != nil {
+			return nil, fmt.Errorf("Volume Driver not coming up in a node [%s]. Error:[%v]", n.Name, err)
+		}
+	}
+
+	return topologyLabels, nil
+}
+
+func populateIntervals() {
+	triggerInterval = map[string]map[int]time.Duration{}
```
kphalgun-px, line 473: Since isTriggerEnabled uses GetWaitTime we need not triggerInterval.

kphalgun-px:
Please remove the PxE Longevity-specific code for easier review.

tests/backup/backup_longevity_test.go
atrivedi-px: Removed from backup tests

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,452 @@
...
+	. "github.com/onsi/ginkgo"
+	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/portworx/torpedo/drivers/scheduler"
+	. "github.com/portworx/torpedo/tests"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	testTriggersConfigMap = "longevity-triggers"
```
mkoppal-px, line 22: Should we have different config map for backup related triggers?

atrivedi-px: Sure, will change this to pxb-longevity-triggers

tests/backuptestTriggers.go
```diff
@@ -0,0 +1,634 @@
...
+	EventValidateScheduleApplication:    eventValidateScheduleApplication,
+	EventAddCredentialandBackupLocation: eventAddCredentialandBackupLocation,
+	EventAddSourceAndDestinationCluster: eventAddSourceAndDestinationCluster,
+	EventCreateBackup:                   eventCreateBackup,
+	EventRestore:                        eventRestore,
+}
+
+var ctx, _ = backup.GetAdminCtxFromSecret()
+
+type PxBackupEventBuilder func(*PxBackupLongevity) (error, string, EventData)
```
mkoppal-px, line 111: Why do we need a string output here?
Also, error should be always in the end as per convention.

mkoppal-px:
Remove all the unused variables and functions from `backup_longevity_test.go`

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,428 @@
...
+
+	minRunTime := Inst().MinRunTimeMins
+	timeout := (minRunTime) * 60
+
+	start := time.Now().Local()
+	lastInvocationTime := start
+
+	for {
+		// if timeout is 0, run indefinitely
+		if timeout != 0 && int(time.Since(start).Seconds()) > timeout {
```
mkoppal-px, line 147: This will always be false

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+func isDisruptiveTrigger(triggerType string) bool {
+	return disruptiveTriggers[triggerType]
+}
+
+func populateDataFromConfigMap(configData *map[string]string) error {
+	log.Infof("ChaosMap provided: %v", configData)
+	setEmailRecipients(configData)
+	setEmailHost(configData)
```
mkoppal-px, line 279: Handle the error here

tests/backup/backup_longevity_test.go
atrivedi-px: Done

tests/backuptestTriggers.go
atrivedi-px: String are the highlight events/any anomaly during the event which is not an error.

tests/backup/backup_longevity_test.go
atrivedi-px: Done

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,409 @@
...
+	emailTriggerFunction = map[string]func(){
+		EmailReporter: TriggerEmailReporter,
+	}
+
+	BeforeEach(func() {
+		if !populateDone {
+			StartPxBackupTorpedoTest("BackupLongevityTest",
+				"Longevity Run For Backup", nil, 0, ATrivedi, Q1FY24)
+			populateIntervals()
+			//  populateDisruptiveTriggers()
```
lsrinivas-pure, line 75: delete if not required

tests/backup/backup_longevity_test.go
lsrinivas-pure: MinRunTimeMins default value is 0 and in that case this condition will never be true and i.e expected. in case any MinRunTimeMins is given like 60, this condition will be true after 60 mins

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,409 @@
...
+			}
+		})
+	})
+	JustAfterEach(func() {
+		defer EndTorpedoTest()
+		AfterEachTest(contexts)
+	})
+})
+
+func backupEventTrigger(wg *sync.WaitGroup,
```
lsrinivas-pure, line 131: can we use same testTriggerfunction in longevity_test passing backup trigger functions

tests/backup/backup_longevity_test.go
```diff
@@ -0,0 +1,409 @@
...
+
+			lastInvocationTime = time.Now().Local()
+
+		}
+		time.Sleep(controlLoopSleepTime)
+	}
+	os.Exit(0)
+}
+
+func emailEventTrigger(wg *sync.WaitGroup,
```
lsrinivas-pure, line 167: please see if we can use emailEventTrigger in longevity test

tests/backup/backup_longevity_test.go
lsrinivas-pure: please use same functions on longevity test file. else we cant run px and backup events together

lsrinivas-pure (changes requested):

tests/backup/backup_longevity_test.go
atrivedi-px: This may be required in future, for now only normal events are added but we are planning to add pxb disruptive events in next cycles.

tests/backup/backup_longevity_test.go
atrivedi-px: This is same as longevity test, we copied the function from same

tests/backup/backup_longevity_test.go
atrivedi-px: This is same as longevity test, we copied the function from same and renamed as per backup trigger

tests/backup/backup_longevity_test.go
atrivedi-px: Done

tests/backup/backup_longevity_test.go
lsrinivas-pure: Hi @atrivedi-px I would like to understand if it is same can we use the existing function itself?

tests/backup/backup_longevity_test.go
lsrinivas-pure: please add a ToDO comment

PR #2052, opened 2024-02-05
PWX-35975: added journal check to pool tests

ragsingh-px:
Added a isjournal check for pool expansion tests
Closes # PWX-35975

### DIFF
--- a/tests/basic/pool_function_test.go
+++ b/tests/basic/pool_function_test.go
@@ -116,6 +116,11 @@ var _ = Describe("{PoolExpandSmoky}", func() {
 	//	targetSizeInBytes = originalSizeInBytes + 100*units.GiB
 	//	targetSizeGiB = targetSizeInBytes / units.GiB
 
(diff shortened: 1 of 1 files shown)


tests/basic/pool_function_test.go
```diff
@@ -863,10 +930,14 @@ var _ = Describe("{PoolExpandResizeDiskInMaintenanceMode}", func() {
 			originalSizeInBytes = poolToResize.TotalSize
 			targetSizeInBytes = originalSizeInBytes + 100*units.GiB
 			targetSizeGiB = targetSizeInBytes / units.GiB
-
+			isjournal, err := IsJournalEnabled()
+			log.FailOnError(err, "Failed to check is journal enabled")
+			if isjournal {
+				targetSizeGiB = targetSizeGiB - 3
+			}
```
rkinhalkar-px, line 937: we have to reduce the size after we trigger pool expansion

PR #2068, opened 2024-02-09
Modifying stork test case as per current behaviour

atrivedi-px:
Modify testcase behaviour to fail backups and restores if custom admin namespace is not present

Closes #PB-5520


kshithijiyer-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #2073, opened 2024-02-12
Initial tests and driver checkin for Nomad support in Torpedo

dbhatnagar-px:
The following tests are checked in with this PR: 
1. CreateAndValidateNomadVolume
2. RunFioJobOnNomad
3. RunMultipleFioJobsOnNomad
4. ScaleFioJobOnNomad
5. RunFioJobWithSnapshotOperationsOnNomad

As of now we are only supporting fio apps in Nomad. With next few PRs, we will create support for more apps. 

There is no Nomad pipeline yet so have run these tests locally. Will create pipeline soon in next few PRs as well.

### DIFF
--- a/tests/basic/basic_test.go
+++ b/tests/basic/basic_test.go
@@ -45,29 +45,38 @@ func TestBasic(t *testing.T) {
 var _ = BeforeSuite(func() {
 	dash = Inst().Dash
 	log.Infof("Init instance")
-	InitInstance()
-	dash.TestSetBegin(dash.TestSet)
-	EnableAutoFSTrim()
+	value, exists := os.LookupEnv("NOMAD_ADDR")
+	if !exists {
+		InitInstance()
+		dash.TestSetBegin(dash.TestSet)
+		EnableAutoFSTrim()
+	} else {
+		log.Infof("Value set for Nomad cluster is: %v", value)
+		dash.TestSetBegin(dash.TestSet)
+	}
 })
 
 var _ = AfterSuite(func() {
-	TestLogger = CreateLogger("SystemCheck.log")
-	defer dash.TestSetEnd()
-	defer CloseLogger(TestLogger)
-	defer dash.TestCaseEnd()
-	// making sure validate clean up executed even if systemcheck failed
-	defer func() {
-		if wantAllAfterSuiteActions || wantAfterSuiteValidateCleanup {
-			dash.TestCaseBegin("Validate Cleanup", "Validating clean up", "", nil)
-			ValidateCleanup()
-		}
-	}()
+	_, exists := os.LookupEnv("NOMAD_ADDR")
+	if !exists {
+		TestLogger = CreateLogger("SystemCheck.log")
+		defer dash.TestSetEnd()
+		defer CloseLogger(TestLogger)
+		defer dash.TestCaseEnd()
+		// making sure validate clean up executed even if systemcheck failed
+		defer func() {
+			if wantAllAfterSuiteActions || wantAfterSuiteValidateCleanup {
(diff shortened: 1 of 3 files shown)


tests/basic/basic_test.go
```diff
@@ -45,29 +45,38 @@ func TestBasic(t *testing.T) {
...
-			dash.TestCaseBegin("Validate Cleanup", "Validating clean up", "", nil)
-			ValidateCleanup()
-		}
-	}()
+	_, exists := os.LookupEnv("NOMAD_ADDR")
+	if !exists {
+		TestLogger = CreateLogger("SystemCheck.log")
+		defer dash.TestSetEnd()
+		defer CloseLogger(TestLogger)
+		defer dash.TestCaseEnd()
```
mkoppal-px, line 65: Don't you want to send the test results to aetos in case of nomad?

tests/basic/nomad_test.go
```diff
@@ -0,0 +1,415 @@
...
+	. "github.com/portworx/torpedo/tests"
+	"time"
+)
+
+var _ = Describe("{ListNodesOnNomadCluster}", func() {
+	var client *nomad.NomadClient
+	var err error
+	JustBeforeEach(func() {
+		StartTorpedoTest("ListNodesOnNomadCluster", "List nodes in Nomad Cluster", nil, 0)
+		client, err = nomad.NewNomadClient()
```
lsrinivas-pure, line 18: since client is being used in all nomad tests. can we move it before suite?

tests/basic/nomad_test.go
```diff
@@ -0,0 +1,415 @@
...
+			log.Infof("Volume %v created and validated successfully", volumeID)
+		} else {
+			log.Errorf(" Volume %v not created successfully")
+		}
+	})
+	JustAfterEach(func() {
+		for _, volume := range volumes {
+			err = client.DeleteVolume(volume.ID)
+			if err != nil {
+				log.Errorf("Volume %v could not be deleted with err %v", volume.ID, err)
```
lsrinivas-pure, line 68: dont you want to use failonerror?

tests/basic/nomad_test.go
```diff
@@ -0,0 +1,415 @@
...
+				log.Errorf("Fio job is not running")
+				return
+			}
+			log.Infof("Allocation and fio-task is running successfully as of now. Will check in 1 minute again.")
+			time.Sleep(checkInterval)
+		}
+		log.Infof("Fio ran successfully for the duration of the test")
+	})
+	AfterEach(func() {
+		err = client.DeleteJob("fio")
```
lsrinivas-pure, line 120: seems this clean up is repeating. can we have cleanup function and call it everywhere

tests/basic/basic_test.go
dbhatnagar-px: Yes - already fixed this in next PR that will have better handling here. Raising that in a couple of days

tests/basic/nomad_test.go
dbhatnagar-px: Ok, good idea. Will handle this in next PR.

tests/basic/nomad_test.go
dbhatnagar-px: Yes - I am changing the cleanup logic in my next PR. This one by one logic will be moved to loop based.

tests/basic/nomad_test.go
dbhatnagar-px: Good catch - already changing this logic in my next PR that also involve more tests.

lsrinivas-pure (approved):
LGTM

PR #2076, opened 2024-02-12
PTX-22333 Fix autopilot validation for backup tests

nikolaypopov:
Fix autopilot validation for backup tests by not looking for PX and just getting list of all deployment and finding autopilot deployment and returning its namespace.

See run here https://jenkins.pwx.dev.purestorage.com/job/Users/job/Aniket/job/AutopilotAll/131/console
```
12:02:24  2024-02-12 20:02:23 +0000:[DEBUG] [k8s.(*K8s).GetAutopilotNamespace:#6548] - Found [autopilot] deployment in [kube-system] namespace
```

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -6532,21 +6532,28 @@ func (k *K8s) GetPortworxNamespace() (string, error) {
 	return "", fmt.Errorf("failed to determine which namespace Portowrx is deployed in, if deployed at all")
 }
 
-// GetAutopilotNamespace looks for StorageCluster object and returns the namespace of the object
-// If StorageCluster object is not found, will try to determine namespace by checking portworx-service location
+// GetAutopilotNamespace list all deployments in all namespaces and looks for autopilot deployment and returns its namespace
 func (k *K8s) GetAutopilotNamespace() (string, error) {
-	stc, err := k8sOperator.ListStorageClusters("")
+	var autopilotNamespace string
+	deployments, err := k8sApps.ListDeployments("", metav1.ListOptions{})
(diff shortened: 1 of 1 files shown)


Santhoshairody (approved):

drivers/scheduler/k8s/k8s.go
```diff
@@ -6532,21 +6532,28 @@ func (k *K8s) GetPortworxNamespace() (string, error) {
...
+				log.Debugf("Found [%s] deployment in [%s] namespace", deployment.Name, deployment.Namespace)
+				autopilotNamespace = deployment.Namespace
+			}
+		}
 	}
-	if len(stc.Items) > 0 {
-		log.Debugf("Found StorageCluster object, will assume Portworx and Autopilot are deployed in [%s] namespace", stc.Items[0].Namespace)
-		return stc.Items[0].Namespace, nil
+
+	if len(autopilotNamespace) == 0 {
```
kshithijiyer-px, line 6552: I tired running this what I am seeing is that the job is failing here. This check needs to be changed to check if PX as well as autopilot is there if not then only fails. This check currently fails if autopilot is not there even when there is no PX. 

The job which we are running here doesn't install both and hence it's causing it to fail. 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/Kshithij/job/kiyer-iks-patch-version/18/console

kshithijiyer-px (changes requested):
We need to change to check to see 2 things, If px is there then only check if autopilot is there. So your check will become:

len(autopilotNamespace) == 0 && len(pxNamespace) != 0

nikolaypopov:
Not needed anymore, fix was done in another PR

PR #2090, opened 2024-02-17
PTX-22439 Fixing IsPKS() return condition

nikolaypopov:
Fixing IsPKS() return condition, due to strconv.ParseBool: parsing "": invalid syntax
```
18:11:27   [38;5;9m[FAILED] Unexpected error:
18:11:27       <*strconv.NumError | 0xc0017a42d0>: 
18:11:27       strconv.ParseBool: parsing "": invalid syntax
18:11:27       {
18:11:27           Func: "ParseBool",
18:11:27           Num: "",
18:11:27           Err: <*errors.errorString | 0xc47ef60>{s: "invalid syntax"},
18:11:27       }
```


tests/basic/telemetry_test.go
```diff
@@ -88,10 +88,9 @@ func IsPKS() bool {
 	if isOpBased {
 		log.Info("This is PX Operator based install, checking if StorageCluster has PKS annotation set to true...")
 		spec, err := Inst().V.GetDriver()
-		Expect(err).ToNot(HaveOccurred())
+		log.FailOnError(err, "Failed to get StorageCluster object")
 		isPks, err := strconv.ParseBool(spec.Annotations["portworx.io/is-pks"])
```
lsrinivas-pure, line 92: can we change it like this 

val, ok:= spec.Annotations["portworx.io/is-pks"]
if ok {
isPks, err := strconv.ParseBool(val)
log.FailOnError(err, "Failed to check PKS")
return isPKS
}

lsrinivas-pure (approved):

tests/basic/telemetry_test.go
nikolaypopov: This `strconv.ParseBool(spec.Annotations["portworx.io/is-pks"])` already covers that fact that this annotation exists, if it doesn't, I am assuming its not PKS.
Basically, if this annotation doesn't exists or its value is not `true` or `false`, PX itself would fail to deploy as it needs this annotation to deploy properly on PKS.

lsrinivas-pure: ya but for non PKS set up this annotation will be not there and we will get error while parsing to bool. since here anyway you are checking for err == nil, it should be fine

PR #2098, opened 2024-02-20
Fixing namespace label issues

vpinisetti-px:
To fix the namespace label issues

Closes # https://portworx.atlassian.net/browse/PB-5345

Vanilla: where the default labels are 3

https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/4556/consoleFull

OCP 4.13: where the default labels are 7

https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/4584/console

### DIFF
--- a/tests/backup/backup_restore_basic_test.go
+++ b/tests/backup/backup_restore_basic_test.go
@@ -1682,6 +1682,7 @@ var _ = Describe("{AddMultipleNamespaceLabels}", func() {
 	var (
 		batchSize                                int
 		desiredNumLabels                         int
+		defaultLabelMapLength                    int
 		backupLocationUID                        string
 		cloudCredUID                             string
 		clusterUid                               string
@@ -1699,6 +1700,7 @@ var _ = Describe("{AddMultipleNamespaceLabels}", func() {
 		cloudCredUidList                         []string
 		scheduledAppContexts                     []*scheduler.Context
 		scheduleRestoreMapping                   map[string]string
+		defaultLabelMap                          map[string]string
 		fetchedLabelMap                          map[string]string
 		labelMap                                 map[string]string
 		err                                      error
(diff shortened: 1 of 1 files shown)


tests/backup/backup_restore_basic_test.go
```diff
@@ -1742,9 +1749,9 @@ var _ = Describe("{AddMultipleNamespaceLabels}", func() {
 			log.InfoD("Verifying number of labels added to namespace")
 			fetchedLabelMap, err = Inst().S.GetNamespaceLabel(bkpNamespaces[0])
 			dash.VerifyFatal(err, nil, fmt.Sprintf("Fetching the labels added to namespace %s", bkpNamespaces[0]))
-			length := len(fetchedLabelMap)
-			dash.VerifyFatal(err, nil, fmt.Sprintf("Fetching labels %v for namespace %v", length-3, bkpNamespaces[0]))
-			dash.VerifyFatal(length-3, desiredNumLabels, fmt.Sprintf("Verifying number of added labels to desired labels for namespace %v", bkpNamespaces[0]))
+			updatedLabelMapLength := len(fetchedLabelMap)
+			dash.VerifyFatal(err, nil, fmt.Sprintf("Length of updated labels for namespace %v : %v", updatedLabelMapLength, bkpNamespaces[0]))
```
kphalgun-px, line 1753: How about?

```go
log.InfoD("Length of updated labels for namespace %s: %v", bkpNamespaces[0], updatedLabelMapLength)
```

tests/backup/backup_restore_basic_test.go
```diff
@@ -1731,6 +1733,11 @@ var _ = Describe("{AddMultipleNamespaceLabels}", func() {
 			ValidateApplications(scheduledAppContexts)
 		})
 		Step("Adding labels to namespaces in multiple of 10 until 1000", func() {
+			log.InfoD("Verifying default labels in the namespace %s", bkpNamespaces[0])
+			defaultLabelMap, err = Inst().S.GetNamespaceLabel(bkpNamespaces[0])
+			dash.VerifyFatal(err, nil, fmt.Sprintf("Default labels presented in the namespace %v : %v", bkpNamespaces[0], defaultLabelMap))
```
mkoppal-px, line 1738: Use `log.FailOnError` here

tests/backup/backup_restore_basic_test.go
```diff
@@ -1731,6 +1733,11 @@ var _ = Describe("{AddMultipleNamespaceLabels}", func() {
...
+			defaultLabelMapLength = len(defaultLabelMap)
+			dash.VerifyFatal(err, nil, fmt.Sprintf("Lenth of default labels in the namespace %v : %v", bkpNamespaces[0], defaultLabelMapLength))
```
mkoppal-px, line 1740: nit: `Length`

mkoppal-px:
This issue was mainly seen on OCP/IBM. Please run the tests against those platforms.

vpinisetti-px:
> Do you really need a number of labels? Why can we fetch the list at start and then just add it to the desired labels and just match the two lists?

@kshithijiyer-px This is required as per test steps. Even after comparison two maps will end up with label count as per the test steps to verify all 1000 labels got applied or not and also I believe this is much simpler solution.

kshithijiyer-px:
> > Do you really need a number of labels? Why can we fetch the list at start and then just add it to the desired labels and just match the two lists?
> 
> @kshithijiyer-px This is required as per test steps. Even after comparison two maps will end up with label count as per the test steps to verify all 1000 labels got applied or not and also I believe this is much simpler solution.

@vpinisetti-px The reason of asking to check if the 1000 labels are the same labels is that programmatically it's easy to check if 2 maps have the same value.

kshithijiyer-px (approved):
LGTM

kphalgun-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #2118, opened 2024-02-27
repro clousnap restore

lsrinivas-pure:
Cloudsnap restore test is failing when restore happens multiple times as previously created `volumesnapshotrestores.stork.libopenstorage.org `crd still exists.
This fix will delete `volumesnapshotrestores.stork.libopenstorage.org` CRD once successfully validated.

Closes # PTX-22599


dbhatnagar-px (approved):
Lgtm

PR #2123, opened 2024-02-27
Fix Detect portworx pod restart logic for non-portworx volume driver

jainsmit:
Fix Detect portworx pod restart logic for non-portworx volume driver

Move the code from k8s scheduler to portworx volume driver. 

Validating portworx pod restart counts only when pxd (portworx) volume driver is used.

Closes # PTX-22369


kshithijiyer-px (approved):
LGTM

nikolaypopov (approved):

drivers/volume/portworx/portworx.go
```diff
@@ -450,6 +455,25 @@ func (d *portworx) init(sched, nodeDriver, token, storageProvisioner, csiGeneric
 	} else {
 		torpedovolume.StorageProvisioner = provisioners[torpedovolume.DefaultStorageProvisioner]
 	}
+
+	// Update node PxPodRestartCount during init
+	k8s, err := scheduler.Get(sched)
```
kphalgun-px, line 460: How about? or something else instead of k8s

```
schedDriver, err := scheduler.Get(sched)
```

drivers/volume/portworx/portworx.go
```diff
@@ -450,6 +455,25 @@ func (d *portworx) init(sched, nodeDriver, token, storageProvisioner, csiGeneric
...
+	if err != nil {
+		return fmt.Errorf("unable to get portworx pods restart count. Error: [%v]", err)
+	}
+
+	for pod, value := range pxPodRestartCountMap {
+		n, err := node.GetNodeByIP(pod.Status.HostIP)
+		if err != nil {
+			return err
+		}
+		n.PxPodRestartCount = value
```
kphalgun-px, line 475: I think we are missing node.UpdateNode.

jainsmit: why UpdateNode is needed when you are updating the object structure field ?

kphalgun-px: I think the node.GetNodeByIP actually returns a copy of the node object, rather than a pointer to the nodeRegistry map in the node package. The node.UpdateNode can help update the nodeRegistry map.

jainsmit: Yes you are right, good catch . Thanks !!

drivers/volume/portworx/portworx.go
jainsmit: Done

kphalgun-px (approved):
LGTM

PR #2124, opened 2024-02-28
Autopilot changes for OCP prometheus

Santhoshairody:
PX prometheus need not to be the requirements from OCP 4.14.x . Hence removing this restriction.
PTX-22584
Closes #
PTX-22584


drivers/scheduler/k8s/k8s.go
```diff
@@ -3872,16 +3872,10 @@ func (k *K8s) ValidateVolumes(ctx *scheduler.Context, timeout, retryInterval tim
 			autopilotLabels["name"] = "autopilot"
 			autopilotPods, err := k8sCore.GetPods(autopilotDefaultNamespace, autopilotLabels)
```
nikolaypopov, line 3873: Can we find any there is a function that gets Portworx namespace or write one and use it here instead of hardcoded `kube-system`, as it is not default namespace anymore

nikolaypopov (changes requested):

drivers/scheduler/k8s/k8s.go
Santhoshairody: > Can we find any there is a function that gets Portworx namespace or write one and use it here instead of hardcoded `kube-system`, as it is not default namespace anymore

Addressed now

drivers/scheduler/k8s/k8s.go
```diff
@@ -3841,27 +3844,20 @@ func (k *K8s) ValidateVolumes(ctx *scheduler.Context, timeout, retryInterval tim
...
-			if err != nil {
-				return err
-			}
-			autopilotEnabled = autopilotEnabled && !(len(autopilotPods.Items) == 0) && !(len(prometheusPods.Items) == 0)
-			if autopilotEnabled {
+			if autopilotEnabledOnPvc {
+				isAutopilotPodAvailable, err := k.IsAutopilotEnabled()
+				if err != nil {
+					return err
+				} else if isAutopilotPodAvailable == false {
```
nikolaypopov, line 3858:
```suggestion
				} else if !isAutopilotPodAvailable {
```

drivers/scheduler/k8s/k8s.go
```diff
@@ -6676,16 +6672,88 @@ func (k *K8s) addAnnotationsToPVC(pvc *corev1.PersistentVolumeClaim, annotations
...
+			namespaces = append(namespaces, svc.Namespace)
+		}
+	}
+	if len(namespaces) > 0 {
+		// If portworx-service is deployed in 2 namespaces, return the none kube-system namespace
+		if len(namespaces) == 2 {
+			log.Debugf("Found [%s] service in 2 different namespaces %s", portworxServiceName, namespaces)
+			for _, namespace := range namespaces {
+				if namespace != "kube-system" {
+					log.Debugf("When Portworx deployed outside of [kube-system] namespace, it also creates [%s] service in the [kube-system ] namespace as well as in the namespace it is deployed in", portworxServiceName)
```
nikolaypopov, line 6740: nit: extra space `[kube-system ]`
```suggestion
					log.Debugf("When Portworx deployed outside of [kube-system] namespace, it also creates [%s] service in the [kube-system] namespace as well as in the namespace it is deployed in", portworxServiceName)
```

nikolaypopov (approved):
LGTM

PR #2126, opened 2024-02-28
Added code to get backup status before proceeding for restore for test case SwapShareBackup

snigdha-px:
Testcase SwapShareBackup is failing intermittently . When trying to restore the shared backup, it fails with error that backup is not in success state
We do not have sufficient logs to see what state the backup is in and the reason for that state
hence added GetBackupStatusWithReason before restoring the backup
Closes #https://portworx.atlassian.net/browse/PB-5663

log : https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/4641/consoleFull
https://aetos.pwx.purestorage.com/resultSet/testSetID/558632/testCaseID/3353536

### DIFF
--- a/tests/backup/backup_share_test.go
+++ b/tests/backup/backup_share_test.go
@@ -4085,6 +4085,8 @@ var _ = Describe("{SwapShareBackup}", func() {
 			log.FailOnError(err, "Failed to share backup %s", backupName)
 		})
 		Step(fmt.Sprintf("validate the backup shared %s is present in user context %s", backupName, users[1]), func() {
+			ctxNonAdmin, err := backup.GetNonAdminCtx(users[1], CommonPassword)
+			log.FailOnError(err, "Fetching non admin ctx")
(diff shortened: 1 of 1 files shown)


atrivedi-px (approved):
LGTM

mkoppal-px (approved):
LGTM

kshithijiyer-px (approved):
LGTM

kphalgun-px:
How about we log the backup status and reason in CreateRestore?

snigdha-px:
> How about we log the backup status and reason in CreateRestore?

Backup status is validated in all the other testcases as we use CreateBackupAndValidate
Only for this testcase backup status is not validated after sharing , hence added
Adding in "CreateRestore" will be duplicate verification

kphalgun-px:
> Only for this testcase backup status is not validated after sharing , hence added

I wanted to log the backup status and reason to help with future debugging. If this is the only test case that does not validate the shared backup, then I guess it is fine.

PR #2129, opened 2024-02-28
Commenting out cluster get nodes api calls during platform upgrade

jainsmit:
Commenting out cluster get nodes api calls during platform upgrade

getClusterNodesInfo doesn't works for most of the distribution and need to find better way to validate during platform upgrade

Closes # PTX-22497

Minor change


nikolaypopov (approved):
LGTM, just confirm with @lsrinivas-pure before merging

PR #2150, opened 2024-03-05
[PTX-22697] Fix Travis For Longevity

kphalgun-px:
This PR fixes Travis for longevity tests and introduces a workaround by setting the TEST_SUITE to 'bin/longevity.test' if FOCUS_TESTS is set to any of the longevity tests.

Closes #PTX-22697

Since Travis will only work once this PR is merged into the master branch, it seems there is no way to test this beforehand.

### DIFF
--- a/deployments/deploy-ssh.sh
+++ b/deployments/deploy-ssh.sh
@@ -251,6 +251,16 @@ done
 
 echo "checking if we need to override test suite: ${TEST_SUITE}"
 
+# TODO: Remove this after all longevity jobs switch to 'bin/longevity.test' for TEST_SUITE.
+case $FOCUS_TESTS in
+  Longevity|UpgradeLongevity|BackupLongevity)
+    TEST_SUITE="bin/longevity.test"
+    echo "Warning: Based on the FOCUS_TESTS ('$FOCUS_TESTS'), the TEST_SUITE ('$TEST_SUITE') is set to 'bin/longevity.test'"
+    ;;
+  *)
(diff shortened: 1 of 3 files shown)


atrivedi-px (approved):
LGTM

deployments/deploy-ssh.sh
```diff
@@ -251,6 +251,16 @@ done
 
 echo "checking if we need to override test suite: ${TEST_SUITE}"
 
+# TODO: Remove this after all longevity jobs switch to 'bin/longevity.test' for TEST_SUITE.
+case $FOCUS_TESTS in
+  Longevity|UpgradeLongevity|BackupLongevity)
```
atrivedi-px, line 256: Could we have a wildcard entry *Longevity* so that even if any new test is added TEST_SUITE is set to `bin/longevity.test` for now.

kphalgun-px: I thought so, but if someone includes the 'Longevity' keyword in any of their tests, it will incorrectly set the TEST_SUITE.

mkoppal-px: Shouldn't the `TEST_SUITE` parameter be passed from the pipeline?

kphalgun-px: It is a parameter in all the pipelines but it is difficult to list all longevity pipelines as of now.

lsrinivas-pure (approved):
LGTM

PR #2154, opened 2024-03-05
[PTX-22692]Fixing panic issue when we add new storage nodes in OCP

lsrinivas-pure:
- Fixing panic issue when we add new storage nodes in OCP
- Adding extra checks for getting vm objects

Closes #PTX-22692


mkoppal-px (approved):
LGTM

PR #2163, opened 2024-03-06
[PTX-22648]fixing volume delete event and adding logic to verify FA/FB backend a…

lsrinivas-pure:
…fter vols deletion

- Fixing volumedelete event which is failing other events if any volume deletion is failed
- Add validation for pure volumes backend validation once px volume is deleted
- fixed volumeclone event ,which is failing due to expected parent volume name is not set
Closes #PTX-22648

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -1681,7 +1681,8 @@ func TearDownContext(ctx *scheduler.Context, opts map[string]bool) {
 		})
 
 		if !ctx.SkipVolumeValidation {
-			ValidateVolumesDeleted(ctx.App.Key, vols)
+			err = ValidateVolumesDeleted(ctx.App.Key, vols)
+			log.FailOnError(err, "Failed to delete volumes for app %s", ctx.App.Key)
 		}
 
 		// Delete Cluster Scope objects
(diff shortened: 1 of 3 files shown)


tests/common.go
```diff
@@ -1681,7 +1681,8 @@ func TearDownContext(ctx *scheduler.Context, opts map[string]bool) {
 		})
 
 		if !ctx.SkipVolumeValidation {
-			ValidateVolumesDeleted(ctx.App.Key, vols)
+			err = ValidateVolumesDeleted(ctx.App.Key, vols)
+			log.FailOnError(err, "Failed to deleting volumes for app %s", ctx.App.Key)
```
vumesh-px, line 1685: nit: 
```suggestion
			log.FailOnError(err, "Failed to delete volumes for app %s", ctx.App.Key)
```

tests/testTriggers.go
```diff
@@ -3466,14 +3468,62 @@ func TriggerVolumeDelete(contexts *[]*scheduler.Context, recordChan *chan *Event
 				UpdateOutcome(event, err)
 			})
 
-			ValidateVolumesDeleted(ctx.App.Key, vols)
+			err := ValidateVolumesDeleted(ctx.App.Key, vols)
+			UpdateOutcome(event, err)
+			log.Infof("waiting for 5 mins allowing vols to delete")
+			time.Sleep(5 * time.Minute)
```
vumesh-px, line 3474: instead of static 5mins wait, can we have a while loop for 5min which continuously checks for the existence of the LUN, if not exit quicker.

vumesh-px (changes requested):

tests/common.go
lsrinivas-pure: fixed

tests/testTriggers.go
lsrinivas-pure: fixed

vumesh-px (approved):
LGTM

aravindranath14 (approved):
LGTM

PR #2176, opened 2024-03-08
PTX-22748 Fix OCP Prometheus prereq methods and move it to openshift scheduler

nikolaypopov:
**Description:** Fix OCP Prometheus prereq methods and move it to `openshift` scheduler

**NOTE:** Also additional refactoring and removing some old code that is not needed anymore

**Jenkins job:** https://jenkins.pwx.dev.purestorage.com/job/Torpedo/view/Torpedo%20NextPX/job/tp-ocp-upgrade/572/console


lsrinivas-pure (approved):
LGTM

PR #2184, opened 2024-03-11
increase stats update time from 2 min to 5 min

vprabhakar-px:
timeout increase from 2 min to 5 min , this is verified with private branch


lsrinivas-pure (approved):
LGTM

PR #2185, opened 2024-03-11
PWX-36081 : Adding check for PodDisruptionBudget value during kubernetes upgrade

svijaykumar-px:
**What this PR does / why we need it**: This PR adds validation to the poddisruptionbudget value during kubernetes upgrade in a cluster. It constantly monitors the pdb value throughout the upgrade to ensure the value of minAvailable does not change and only nodes lesser than allowed disruptions are down.

Closes # PWX-36081

The code has been tested using [this](https://jenkins.pwx.dev.purestorage.com/job/Users/job/svijaykumar/job/tp-ocp-upgrade-copy/18/console) jenkins job where upgrade of OCP cluster is performed.


nikolaypopov:
I would like to run this also on some none OCP cluster upgrades, OCP upgrades are pretty pretty straight forward, but AKS, EKS, GKE upgrades are more complex and I have seen things that been added work on OCP but not on other cloud upgrades. Let me know if you need to help to run the fix in the private image on some of these before merging, otherwise it looks good to me

nikolaypopov (approved):
Ran AKS and EKS upgrade jobs with this code in private image, both passed, LGTM

PR #2233, opened 2024-03-21
[PTX-22978] Print Storage pool show after SP operations in longevity

sayalasomayajula-px:
adds sv pool show after pool related operations in longevity

Closes # PTX-22978


tests/common.go
```diff
@@ -607,6 +607,24 @@ func PrintCommandOutput(cmnd string) {
 	log.Infof(output)
 
 }
+func PrintSvPoolStatus() {
+	PrintSPCommandOutput("pxctl sv pool show")
```
lsrinivas-pure, line 611: we need to run this command on the node where pool expansion happened. take node as input and run on the node

tests/testTriggers.go
```diff
@@ -2106,6 +2106,8 @@ func TriggerStorageFullPoolExpansion(contexts *[]*scheduler.Context, recordChan
 
 				}
 			}
+			log.InfoD("Printing The storage pool status after pool expansion")
```
lsrinivas-pure, line 2109: we have other tests adddisk and resizedisk. please add this in those tests also.
Also add this for tests in storage_pool_test.go also

lsrinivas-pure (changes requested):

tests/common.go
sayalasomayajula-px: Changed the function to take node and execute sv pool show on the node

tests/testTriggers.go
sayalasomayajula-px: Added the log accordingly and sv pool show will now happen on the nodes on which storage pool operation happen

tests/common.go
```diff
@@ -614,6 +614,23 @@ func PrintCommandOutput(cmnd string) {
 
 }
 
+func PrintSvPoolStatus(node node.Node) {
+	PrintCommandOutputOnNode("pxctl sv pool show", node)
+}
+func PrintCommandOutputOnNode(cmnd string, n node.Node) {
```
kphalgun-px, line 620: How about using `runCmdGetOutput` in `tests/common.go`? Also, could we add a line between these two functions?

tests/testTriggers.go
```diff
@@ -2095,6 +2095,8 @@ func TriggerStorageFullPoolExpansion(contexts *[]*scheduler.Context, recordChan
 						if newPoolSize >= expectedSizeWithJournal {
 							isExpansionSuccess = true
 						}
+						log.InfoD("Printing The storage pool status after pool expansion")
```
kphalgun-px, line 2098: How about including node details like Name or Id in the log message?

tests/common.go
sayalasomayajula-px: You are right the runCmdGetOutput is doing samething, hence replaced with the following subroutine

tests/testTriggers.go
sayalasomayajula-px: Added the name for all the log statements.

tests/testTriggers.go
```diff
@@ -5537,6 +5539,20 @@ func TriggerMetadataPoolResizeDisk(contexts *[]*scheduler.Context, recordChan *c
...
 
 	})
+
+	stNodes := node.GetStorageNodes()
+	expectedCapacity := (len(stNodes) / 2) + 1
+	poolsToExpand := make([]*opsapi.StoragePool, 0)
+	for _, stNode := range stNodes {
+		eligibility, err := GetPoolExpansionEligibility(&stNode)
+		log.InfoD(fmt.Sprintf("error getting the pool expansion eligibility :%s", err))
+		if len(poolsToExpand) <= expectedCapacity {
```
kphalgun-px, line 5549: Might be unnecessary.

tests/testTriggers.go
```diff
@@ -5537,6 +5539,20 @@ func TriggerMetadataPoolResizeDisk(contexts *[]*scheduler.Context, recordChan *c
 		wg.Wait()
 
 	})
+
+	stNodes := node.GetStorageNodes()
+	expectedCapacity := (len(stNodes) / 2) + 1
+	poolsToExpand := make([]*opsapi.StoragePool, 0)
+	for _, stNode := range stNodes {
+		eligibility, err := GetPoolExpansionEligibility(&stNode)
+		log.InfoD(fmt.Sprintf("error getting the pool expansion eligibility :%s", err))
```
kphalgun-px, line 5548: How about log.Errorf?

sayalasomayajula-px: removed the code and added new one to retirve node from a pool uuid

tests/testTriggers.go
sayalasomayajula-px: removed the code and added new one to retirve node from a pool uuid

lsrinivas-pure (approved):
LGTM

PR #2240, opened 2024-03-22
add extra checks to get volumes for virtual machines

Rohit-PX:
* In OCP 4.14 extra checks might be required to get volumes being used by virtual machines
* PVC owner references are not enough in OCP 4.14 to know if volume belongs to a VM
* The following extra checks will be used if required:
  * Check Data volume owner references
  * Check virt-launcher pod owner references

Closes # https://portworx.atlassian.net/browse/PTX-22957

### DIFF
--- a/drivers/scheduler/k8s/k8s.go
+++ b/drivers/scheduler/k8s/k8s.go
@@ -4328,8 +4328,41 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 				// check if the pvc has our VM as the owner
 				want := false
 				for _, ownerRef := range pvc.OwnerReferences {
-					if ownerRef.Kind == vm.Kind && ownerRef.Name == vm.Name {
-						want = true
+					if ownerRef.Kind == vm.Kind {
+						if ownerRef.Name == vm.Name {
+							want = true
+						}
+					}
+					// If volume is still not found, try the datavolume for the virtual machine
+					if !want && ownerRef.Kind == "DataVolume" {
+						kvClient := k8sKubevirt.GetKubevirtClient()
+						dataVol, err := kvClient.CdiClient().CdiV1beta1().DataVolumes(vm.Namespace).Get(context.TODO(), ownerRef.Name, metav1.GetOptions{})
+						if err != nil && !k8serrors.IsNotFound(err) {
+							return nil, fmt.Errorf("error getting datavolume for vm: %s in namespace: %s. error: %v", vm.Name, vm.Namespace, err)
+						} else {
+							if dataVol != nil {
+								for _, ownerRef := range dataVol.OwnerReferences {
+									if ownerRef.Name == vm.Name {
+										want = true
+									}
+								}
+							}
+						}
+					}
+				}
+				// If volume is still not found, try the virt-launcher pod as well
+				if !want {
+					virtLauncherPodsList, err := k8sCore.GetPods(vm.Namespace, nil)
+					if err != nil {
+						return nil, fmt.Errorf("failed to get pods in namespace: %s: %w", vm.Namespace, err)
+					}
+					for _, virtLauncher := range virtLauncherPodsList.Items {
+						for _, ownerRef := range virtLauncher.OwnerReferences {
+							if ownerRef.Kind == "VirtualMachineInstance" && ownerRef.Name == vm.Name {
+								// since VMI name is same as VM name
+								want = true
+							}
+						}
 					}
 				}
 				if !want {


Rohit-PX:
Successfully ran the entire suite here: https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%20Cloud%20Jobs/job/stork-ocp-kubevirt-deploy-2/30/console

drivers/scheduler/k8s/k8s.go
```diff
@@ -4328,8 +4328,40 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+					if ownerRef.Kind == vm.Kind {
+						if ownerRef.Name == vm.Name {
+							want = true
+						}
+					}
+					// If volume is still not found, try the datavolumes
+					if !want && ownerRef.Kind == "DataVolume" {
+						// get datavolume
+						kvClient := k8sKubevirt.GetKubevirtClient()
+						dvList, err := kvClient.CdiClient().CdiV1beta1().DataVolumes(vm.Namespace).List(context.TODO(), metav1.ListOptions{})
```
pureneelesh, line 4340: We should get only one datavolume that is referenced in the ownerRef. Otherwise, we may include volumes of another VM in the same namespace (in case there are multiple VMs in the same namespace).

drivers/scheduler/k8s/k8s.go
```diff
@@ -4328,8 +4328,40 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+								if ownerRef.Name == vm.Name {
+									want = true
+								}
+							}
+						}
+					}
+				}
+				// If volume is still not found, try the virt-launcher pod as well
+				if !want {
+					virtLauncherPodsList, err := k8sCore.GetPods(vm.Namespace, nil)
```
pureneelesh, line 4355: Similar comment here. We should only consider the pods that are using this PVC. Skip the other pods.

drivers/scheduler/k8s/k8s.go
```diff
@@ -4328,8 +4328,40 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+					if err != nil {
+						return nil, fmt.Errorf("failed to get pods in namespace: %s: %w", vm.Namespace, err)
+					}
+					for _, virtLauncher := range virtLauncherPodsList.Items {
+						for _, ownerRef := range virtLauncher.OwnerReferences {
+							if ownerRef.Kind == "VirtualMachineInstance" && ownerRef.Name == vm.Name {
```
pureneelesh, line 4361: add a comment here that the name of VMI is same as the VM name.

drivers/scheduler/k8s/k8s.go
Rohit-PX: I am not sure I understand, even though we are getting all datavolumes in the VM namespace, in line# 4346 we are checking the datavolume owner reference against the VM name. That way we are ensuring we get only the datavolume that corresponds to the VM in question.

pureneelesh: Ah, yes. So, there is no correctness issue. Do you still want to just get a single data volume instead of listing multiple? We already know the name of the datavolume from the ownerRef. It is odd to use the "list" API when we can use "get".

drivers/scheduler/k8s/k8s.go
pureneelesh: No change needed here since we do need to list all pods in the namespace.

pureneelesh (approved):

drivers/scheduler/k8s/k8s.go
Rohit-PX: Fixed.

Rohit-PX:
Successful run after final changes:

```
--- PASS: TestKubevirt (1393.24s)
    --- PASS: TestKubevirt/kubeVirtHypercOneLiveMigration (1389.75s)
PASS
```

drivers/scheduler/k8s/k8s.go
```diff
@@ -4328,8 +4328,39 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+					if ownerRef.Kind == vm.Kind {
+						if ownerRef.Name == vm.Name {
+							want = true
+						}
+					}
+					// If volume is still not found, try the datavolumes
+					if !want && ownerRef.Kind == "DataVolume" {
+						// get datavolume
+						kvClient := k8sKubevirt.GetKubevirtClient()
+						dataVol, err := kvClient.CdiClient().CdiV1beta1().DataVolumes(vm.Namespace).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
```
pureneelesh, line 4340: pvc.Name -> ownerRef.Name

drivers/scheduler/k8s/k8s.go
```diff
@@ -4328,8 +4328,39 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+						if err != nil {
+							return nil, fmt.Errorf("error getting datavolumes in namespace: %s", vm.Namespace)
```
pureneelesh, line 4342: fix the comment as well

drivers/scheduler/k8s/k8s.go
```diff
@@ -4328,8 +4328,39 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
+						if ownerRef.Name == vm.Name {
+							want = true
+						}
+					}
+					// If volume is still not found, try the datavolumes
+					if !want && ownerRef.Kind == "DataVolume" {
+						// get datavolume
+						kvClient := k8sKubevirt.GetKubevirtClient()
+						dataVol, err := kvClient.CdiClient().CdiV1beta1().DataVolumes(vm.Namespace).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
+						if err != nil {
```
pureneelesh, line 4341: ignore NotFound error

drivers/scheduler/k8s/k8s.go
```diff
@@ -4333,17 +4334,18 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
 							want = true
 						}
 					}
-					// If volume is still not found, try the datavolumes
+					// If volume is still not found, try the datavolume for the vi
```
pureneelesh: typo: vi

drivers/scheduler/k8s/k8s.go
```diff
@@ -4333,17 +4334,18 @@ func (k *K8s) GetVolumes(ctx *scheduler.Context) ([]*volume.Volume, error) {
...
 						kvClient := k8sKubevirt.GetKubevirtClient()
-						dataVol, err := kvClient.CdiClient().CdiV1beta1().DataVolumes(vm.Namespace).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
-						if err != nil {
-							return nil, fmt.Errorf("error getting datavolumes in namespace: %s", vm.Namespace)
-						}
-						for _, ownerRef := range dataVol.OwnerReferences {
-							if ownerRef.Name == vm.Name {
-								want = true
+						dataVol, err := kvClient.CdiClient().CdiV1beta1().DataVolumes(vm.Namespace).Get(context.TODO(), ownerRef.Name, metav1.GetOptions{})
+						if err != nil || k8serrors.IsNotFound(err) {
```
pureneelesh, line 4342: the second condition is redundant. Did you mean to exit out when the error is something other than NotFound?

drivers/scheduler/k8s/k8s.go
Rohit-PX: Fixed.

pureneelesh (approved):

Rohit-PX:
```
-- PASS: TestKubevirt (914.09s)
    --- PASS: TestKubevirt/kubeVirtHypercOneLiveMigration (910.58s)
PASS
```

lsrinivas-pure (approved):
LGTM

PR #2251, opened 2024-03-26
Adding verification for  Pool Resize and Pool rebalance  

Santhoshairody:
Adding verification for  Pool Resize and Pool rebalance .
 
Currently test fails while validating events , with this code check for ARO to complete instead of event based verification. 
This also guarantees that AROs have been created. 
Closes #
PTX-22372


drivers/scheduler/k8s/k8s.go
```diff
@@ -6423,6 +6423,141 @@ func (k *K8s) ValidateAutopilotRuleObjects() error {
 	return nil
 }
 
+// VerifyPoolResizeARO() created and resize completed
+func (k *K8s) VerifyPoolResizeARO(ruleName apapi.AutopilotRule) (bool, error) {
+	var ruleTriggered bool
+	namespace, err := k.GetAutopilotNamespace()
+	if err != nil {
+		log.Warnf("NO Namespace")
```
aghodke1312, line 6431: you can remove the warn. Return false, fmt.Errorf("No Namespace")
+ NO -> No

drivers/scheduler/k8s/k8s.go
```diff
@@ -6423,6 +6423,141 @@ func (k *K8s) ValidateAutopilotRuleObjects() error {
...
+	expectedAroStates := []apapi.RuleState{
+		apapi.RuleStateTriggered,
+		apapi.RuleStateActiveActionsPending,
+		apapi.RuleStateActiveActionsInProgress,
+		apapi.RuleStateActiveActionsTaken,
+	}
+
+	listAutopilotRuleObjects, err := k8sAutopilot.ListAutopilotRuleObjects(namespace)
+	if err != nil {
+		log.Warnf("No ARO  available")
```
aghodke1312, line 6444: same here.

drivers/scheduler/k8s/k8s.go
```diff
@@ -6423,6 +6423,141 @@ func (k *K8s) ValidateAutopilotRuleObjects() error {
...
+		return false, err
+	}
+	if len(listAutopilotRuleObjects.Items) == 0 {
+		log.Warnf("the list of autopilot rule objects is empty, please make sure that you have an appropriate autopilot rule")
```
aghodke1312, line 6448: same

aghodke1312 (approved):

PR #2277, opened 2024-03-29
[PTX-23107]Add helper function for bind mount of kubevirt vm

spai-px:
This PR adds helper function to verify bind mount of a vm.

Closes # PTX-23107


drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		}
+		// Check if all the replicas are attached to the same node
+		volInspect, err := volumeDriver.InspectVolume(vol.ID)
+		if err != nil {
+			return false, fmt.Errorf("failed to inspect volume [%s]: %w", vol.ID, err)
+		}
+		replicaset := volInspect.ReplicaSets
+
+		// check if the replicaset size is same as the global replicaset size
+		if len(replicaset) != len(globalReplicSet) {
```
pureneelesh, line 654: typo: globalReplicSet missing a in Replica

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+			return false, fmt.Errorf("replicaset count mismatch for volume [%s]", vol.ID)
+		}
+		// The criteria to call the bind mount successful is to check if the replicaset of all volumes should be same
+		// check if the replicaset values is same as the global replicaset
+		// Here we are considering globalreplicaset to be source of truth for comparision
+		for _, rs := range replicaset {
+			found := false
+		outerLoop:
+			for _, rsNode := range rs.Nodes {
```
pureneelesh, line 663: need to verify that len(rs.Nodes) matches. This is the HA factor.

len(replicaSet) is aggr level which is usually 1 in all our tests.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+				for _, grs := range globalReplicSet {
+					for _, grsNode := range grs.Nodes {
+						if grsNode == rsNode {
+							found = true
+							break outerLoop
+						}
+					}
+				}
```
pureneelesh, line 671: consider writing a helper function isNodeInReplicaSet(rsNode, globalReplicSet)

tests/basic/kubevirt_test.go
```diff
@@ -45,3 +47,58 @@ var _ = Describe("{AddNewDiskToKubevirtVM}", func() {
...
+
+var _ = Describe("{KubeVirtLiveMigration}", func() {
+
+	JustBeforeEach(func() {
+		StartTorpedoTest("KubeVirtLiveMigration", "Live migrate a kubevirtVM", nil, 0)
+	})
+
+	var contexts []*scheduler.Context
+
+	itLog := "Live migrate a kubevirtVM"
```
pureneelesh, line 59: kubevirtVM should be 2 words

tests/basic/kubevirt_test.go
```diff
@@ -45,3 +47,58 @@ var _ = Describe("{AddNewDiskToKubevirtVM}", func() {
...
 	})
 })
+
+var _ = Describe("{KubeVirtLiveMigration}", func() {
+
+	JustBeforeEach(func() {
+		StartTorpedoTest("KubeVirtLiveMigration", "Live migrate a kubevirtVM", nil, 0)
+	})
+
+	var contexts []*scheduler.Context
```
pureneelesh, line 57: nit: contexts -> appCtxs for readability

tests/basic/kubevirt_test.go
```diff
@@ -45,3 +47,58 @@ var _ = Describe("{AddNewDiskToKubevirtVM}", func() {
...
+			log.InfoD(stepLog)
+			for _, ctx := range contexts {
+				success, err := kubevirt.StartAndWaitForVMIMigration(ctx, context1.TODO(), Inst().S)
+				log.FailOnError(err, "Failed to live migrate kubevirt VM")
+				dash.VerifyFatal(success, true, "Failed to live migrate kubevirt VM?")
+			}
+		})
+
+		for _, ctx := range contexts {
+			bindMount, err := kubevirt.VerifyBindMount(ctx, Inst().S, Inst().V)
```
pureneelesh, line 95: this is a wrong expectation for the test. Live migration does not move the attachment. So, the migrated VM will be nfs-mounted. If this test is passing something is wrong with VerifyBindMount function.

vendor/modules.txt
```diff
@@ -1332,6 +1332,7 @@ github.com/portworx/sched-ops/k8s/errors
 github.com/portworx/sched-ops/k8s/externalsnapshotter
 github.com/portworx/sched-ops/k8s/externalstorage
 github.com/portworx/sched-ops/k8s/kubevirt
+github.com/portworx/sched-ops/k8s/kubevirt-dynamic
```
pureneelesh, line 1335: This is fine, but it might be easier to use the typed client using kubevirt.GetKubeVirtClient().

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
 
 	return nil
 }
+
+// StartAndWaitForVMIMigration starts the VM migration and waits for the VM to be in running state in the new node
+func StartAndWaitForVMIMigration(virtualMachine *scheduler.Context, ctx context1.Context, schedDriver scheduler.Driver) (bool, error) {
+
+	vms, err := GetAllVMsFromScheduledContexts([]*scheduler.Context{virtualMachine})
+	vmiNamespace := vms[0].Namespace
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	vmiName := vms[0].Name
+
+	if err != nil {
```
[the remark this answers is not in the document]
spai-px: Moved the error before accessing vms namespace and name.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		return false, err
+	}
+
+	// Start the VM migration
+	migration, err := kubevirtdy.Instance().CreateVirtualMachineInstanceMigration(ctx, vmiNamespace, vmiName)
+	if err != nil {
+		log.Infof("Failed to create VM migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		return true, err
```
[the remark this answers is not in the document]
spai-px: My bad changed it to false.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		migr, err = kubevirtdy.Instance().GetVirtualMachineInstanceMigration(ctx, vmiNamespace, migration.Name)
+		if err != nil {
+			return "", false, fmt.Errorf("failed to get migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+		if !migr.Completed {
+			return "", true, fmt.Errorf("waiting for migration to complete for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+
+		// get volume from app context
+		volume, err := schedDriver.GetVolumes(virtualMachine)
```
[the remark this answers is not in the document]
spai-px: Done moved this call before we wait for kubevirt migration to end.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		if err != nil {
+			return "", false, fmt.Errorf("failed to get volume for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+
+		// wait until there is only one pod in the running state
+		testPod, err := getVMPod(virtualMachine, volume[0])
+		if err != nil {
+			return "", false, err
+		}
```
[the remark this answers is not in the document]
spai-px: Yeah it's better to add one more validation to check if the pod has been scheduled on a different node than the previous node. Added a check to see if node before and after live migration are different.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		log.InfoD("VM pod live migrated to pod: [%s] in namespace [%s] is in running state", testPod.Name, testPod.Namespace)
+		return "", false, nil
+
+	}
+	_, err = task.DoRetryWithTimeout(t, defaultVmMountCheckTimeout, defaultVmMountCheckRetryInterval)
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+			found = pod
+		}
+	}
+	if found.Name == "" {
+		return nil, fmt.Errorf("failed to find a running pod for volume %s", vol.ID)
+	}
+	return &found, nil
+}
+
+func VerifyBindMount(vmi *scheduler.Context, driver scheduler.Driver, volumeDriver volume.Driver) (bool, error) {
```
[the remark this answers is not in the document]
spai-px: Right will make it consistent across. Will be using virtualMachineCtx moving forward.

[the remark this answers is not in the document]
spai-px: Yeah that sounds right. Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		return false, fmt.Errorf("no volumes found for VM [%s] in namespace [%s]", vmi.App.Key, vmi.App.NameSpace)
+	}
+
+	// Get the node where pod is scheduled
+	vms, err := GetAllVMsFromScheduledContexts([]*scheduler.Context{vmi})
+	if err != nil {
+		return false, err
+	}
+
+	vm := vms[0]
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	log.InfoD("pod name: %s, namespace: %s", vmi.App.Key, vmi.App.NameSpace)
+
+	vmNodeName, err := GetNodeOfVM(vm)
+	if err != nil {
+		return false, err
+	}
+	log.Infof("Pod [%s] is deployed on node [%s]", vmi.App.Key, vmNodeName)
```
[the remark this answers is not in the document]
spai-px: Done

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+			}
+			nodeIpAttachedOn := volInspect.AttachedOn
+			nodeNameAttachedOn, err := node.GetNodeByIP(nodeIpAttachedOn)
+			if err != nil {
+				return "", false, fmt.Errorf("failed to get node name by IP [%s]: %w", nodeIpAttachedOn, err)
+			}
+			log.Infof("Volume [%s] is attached on node [%s]", vol.ID, nodeNameAttachedOn.Name)
+			if err != nil {
+				return "", false, fmt.Errorf("failed to check if volume [%s] is mounted on node [%s]: %w", vol.ID, vmNodeName, err)
+			}
```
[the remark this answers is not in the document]
spai-px: My bad removed.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+			if !(nodeNameAttachedOn.Name == vmNodeName) {
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+				return "", true, fmt.Errorf("volume [%s] is not mounted on node [%s]", vol.ID, vmNodeName)
```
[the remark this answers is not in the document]
spai-px: Done

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+			}
+			return "", false, nil
+		}
+		_, err = task.DoRetryWithTimeout(t, 1*time.Minute, 10*time.Second)
```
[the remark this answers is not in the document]
spai-px: So in case there is a live migration the volumes and the vm's will not be present in the same node then this would fail. So according to @dbhatnagar-px we can have a different function to check for rebalance of volumes and see if the volume gets moved to the node where vm is present. Maybe we can handle this there Removing retry here.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		if err != nil {
+			return false, err
+		}
+		// Check if all the replicas are attached to the same node
```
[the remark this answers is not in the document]
spai-px: Yeah but according to @dbhatnagar-px the volume placement should be such that all the replicas of all the volumes also must be in the same node. therefore I am comparing replicaset of one volume with the other volumes.

drivers/kubevirt/kubevirt.go
spai-px: Yes done.

drivers/kubevirt/kubevirt.go
spai-px: Done

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+				for _, grs := range globalReplicSet {
+					for _, grsNode := range grs.Nodes {
+						if grsNode == rsNode {
+							found = true
+							break outerLoop
+						}
+					}
+				}
+			}
+			if found == false {
```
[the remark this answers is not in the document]
spai-px: Done.

tests/basic/kubevirt_test.go
spai-px: Done

tests/basic/kubevirt_test.go
spai-px: Done

tests/basic/kubevirt_test.go
spai-px: I was just just testing the function and excepted behaviour is for the function to fail so will remove it once done with testing.

vendor/modules.txt
spai-px: Will work on this..

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +511,181 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		}
+		replicaset := volInspect.ReplicaSets
+
+		// check if the replicaset size is same as the global replicaset size
+		if len(replicaset) != len(globalReplicSet) {
+			return false, fmt.Errorf("replicaset count mismatch for volume [%s]", vol.ID)
+		}
+		// The criteria to call the bind mount successful is to check if the replicaset of all volumes should be same
+		// check if the replicaset values is same as the global replicaset
+		// Here we are considering globalreplicaset to be source of truth for comparision
```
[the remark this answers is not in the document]
spai-px: @dbhatnagar-px The condition for bind mount is that volume collocation should be present right?? That's why this logic has been put here. please confirm.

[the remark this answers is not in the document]
pureneelesh: We can (and probably should) verify bind-mount by exec'ing into the pod like the stork test is doing. The checks here are not sufficient. For example, the volumes can be collocated and all attached to the same replica node. But still it can be NFS-mount in case of "localhost NFS".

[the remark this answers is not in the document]
spai-px: Yeah this was supposed to be added but I missed it. Went through the stork test code and have replicated the same here with few modifications required for torpedo. Please have a look.

tests/basic/kubevirt_test.go
```diff
@@ -25,15 +27,21 @@ var _ = Describe("{AddNewDiskToKubevirtVM}", func() {
...
+			bindMount, err := kubevirt.IsVMBindMounted(appCtx, Inst().S, Inst().V, true)
+			log.FailOnError(err, "Failed to verify bind mount")
+			dash.VerifyFatal(bindMount, true, "Failed to verify bind mount")
+		}
+
 		stepLog = "Add one disk to the kubevirt VM"
 		Step(stepLog, func() {
 			log.InfoD(stepLog)
-			success, err := kubevirt.AddDisksToKubevirtVM(contexts, numberOfVolumes, "0.5Gi", Inst().N)
+			success, err := kubevirt.AddDisksToKubevirtVM(appCtxs, numberOfVolumes, "0.5Gi", Inst().N)
```
pureneelesh, line 44: test should verify that the new disk is bind-mounted as well

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+func getVMDiskMountType(pod *corev1.Pod, vmDisk *volume.Volume, volumeDriver volume.Driver, diskName string) (string, error) {
+	podNamespacedName := pod.Namespace + "/" + pod.Name
+	log.Infof("Checking the mount type of %s in pod %s", vmDisk, podNamespacedName)
+
+	// Sample output if the volume is bind-mounted: (vmDisk.diskName is "rootdisk" in this example)
+	// $ kubectl exec -it virt-launcher-fedora-communist-toucan-jfw7n -- mount
+	// ...
+	// /dev/pxd/pxd365793461222635857 on /run/kubevirt-private/vmi-disks/rootdisk type ext4 (rw,relatime,seclabel,discard)
+	// ...
+	volInspect, err := volumeDriver.InspectVolume(vmDisk.ID)
```
pureneelesh, line 784: check err here

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	// Sample output if the volume is nfs-mounted: (vmDisk.diskName is "rootdisk" in this example)
+	// $ kubectl exec -it virt-launcher-fedora-communist-toucan-bqcrp -- mount
+	// ...
+	// 172.30.194.11:/var/lib/osd/pxns/365793461222635857 on /run/kubevirt-private/vmi-disks/rootdisk type nfs (...)
+	// ...
+	nfsMountRE := regexp.MustCompile(fmt.Sprintf(":/var/lib/osd/pxns/%s on .*%s type nfs",
+		volInspect.Id, diskName))
+
+	cmd := []string{"mount"}
+	output, err := core.Instance().RunCommandInPod(cmd, pod.Name, "compute", pod.Namespace)
```
pureneelesh, line 798: why "compute"? Does "" not work to use the default container?

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	if err != nil {
+		return "", fmt.Errorf("failed to run command %v inside the pod %s", cmd, podNamespacedName)
```
pureneelesh, line 800: add "%w:", err

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+
+// isNodeInReplicaSet verifies if the replicaset nodes are present in global replicaset
+func IsNodeInReplicaSet(replicaset []*api.ReplicaSet, globalReplicSet []*api.ReplicaSet) (bool, error) {
+	for _, rs := range replicaset {
+		found := false
+	outerLoop:
+		for _, rsNode := range rs.Nodes {
+			for _, grs := range globalReplicSet {
+				if len(grs.Nodes) != len(rs.Nodes) {
+					return false, fmt.Errorf("replicaset mismatch node count mismatch")
```
pureneelesh, line 756: include more context in the error message

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+				}
+				for _, grsNode := range grs.Nodes {
+					if grsNode == rsNode {
+						found = true
+						break outerLoop
```
pureneelesh, line 761: this looks suspect. We are breaking the outerLoop after finding only one node in the first replicaset. Maybe restructure this code to make it more readable and easier to catch bugs like this?

consider using a simple strategy used by https://github.com/libopenstorage/stork/blob/2962e2b26215f2d32b65f2d1dbaa2c72b348fa38/test/integration_test/kubevirt_hyperconv_test.go#L790

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	vmi, err := kubevirt.Instance().GetVirtualMachineInstance(context1.TODO(), virtualMachineCtx.Name, virtualMachineCtx.Namespace)
+	if err != nil {
+		return "", err
+	}
+	log.InfoD("NodeName: %s", vmi.Status.NodeName)
+	return vmi.Status.NodeName, nil
+}
+
+// isNodeInReplicaSet verifies if the replicaset nodes are present in global replicaset
+func IsNodeInReplicaSet(replicaset []*api.ReplicaSet, globalReplicSet []*api.ReplicaSet) (bool, error) {
```
pureneelesh, line 749: no need for 2 return params. err == nil is same as bool == true

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	volInspect, err := volumeDriver.InspectVolume(vols[0].ID)
+	if err != nil {
+		return false, fmt.Errorf("failed to inspect volume [%s]: %w", vols[0].ID, err)
+	}
+	globalReplicSet := volInspect.ReplicaSets
+	log.InfoD("Length of replicaset: %d", len(globalReplicSet))
+
+	// Check if the volume is mounted on the node
+	for _, vol := range vols {
+		// Check if the volume is mounted on the node
```
pureneelesh, line 652: for readability, consider moving the code inside the for loop in a separate function that checks only one volume

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		}
+
+		// Check if all the replicas are attached to the same node
+		replicaset := volInspect.ReplicaSets
+
+		// check if the replicaset size is same as the global replicaset size
+		if len(replicaset) != len(globalReplicSet) {
+			return false, fmt.Errorf("replicaset count mismatch for volume [%s]", vol.ID)
+		}
+		// The criteria to call the bind mount successful is to check if the replicaset of all volumes should be same
```
pureneelesh, line 674: I still think that the collocation check should be in a separate function. That will simplify the logic here. For bind mount, it is sufficient to verify the output of "mount". No other check is needed.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	if err != nil {
+		return false, err
+	}
+
+	// Get the namespace and name of the VM
+	if len(vms) == 0 {
+		return false, fmt.Errorf("no VMs found for VM [%s] in namespace [%s]", virtualMachineCtx.App.Key, virtualMachineCtx.App.NameSpace)
+	}
+	vmiNamespace := vms[0].Namespace
+	vmiName := vms[0].Name
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+
+	// Start the VM migration
+	migration, err := kubevirtdy.Instance().CreateVirtualMachineInstanceMigration(ctx, vmiNamespace, vmiName)
+	if err != nil {
+		log.Infof("Failed to create VM migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
```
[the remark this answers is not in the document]
spai-px: Done

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		return false, err
+	}
+	log.Infof("VM migration created for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+	log.Infof("Migrating VM [%s] in namespace [%s]", vmiName, vmiNamespace)
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	var migr *kubevirtdy.VirtualMachineInstanceMigration
+
+	// get volume from app context
+	volume, err := schedDriver.GetVolumes(virtualMachineCtx)
+	if err != nil {
+		return false, err
+	}
+
+	//Get the node where the vm is scheduled before the migration
+	nodeName, err := GetNodeOfVM(vms[0])
```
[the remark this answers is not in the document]
spai-px: Yes this is the node before migration, moved it before we create a migration object.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		if err != nil {
+			return "", false, fmt.Errorf("failed to get migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+		if !migr.Completed {
+			return "", true, fmt.Errorf("waiting for migration to complete for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+
+		if err != nil {
+			return "", false, fmt.Errorf("failed to get volume for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
```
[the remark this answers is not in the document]
spai-px: Removed the error.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+
+		// wait until there is only one pod in the running state
+		testPod, err := getVMPod(virtualMachineCtx, volume[0])
+		if err != nil {
+			return "", false, err
+		}
```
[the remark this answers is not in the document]
spai-px: Yeah you are right, we have to wait so changed it to true.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+
+		//Get the node where the vm is scheduled after the migration
+		nodeNameAfterMigration, err := GetNodeOfVM(vms[0])
+		if err != nil {
+			return "", false, err
+		}
+		if nodeName == nodeNameAfterMigration {
+			return "", true, fmt.Errorf("VM pod live migrated [%s] in namespace [%s] but is still on the same node [%s]", testPod.Name, testPod.Namespace, nodeName)
+		}
```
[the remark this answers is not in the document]
spai-px: Yeah changed it to false.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		log.Infof("Failed to create VM migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		return false, err
+	}
+	log.Infof("VM migration created for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+	log.Infof("Migrating VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+	// wait for completion
+	var migr *kubevirtdy.VirtualMachineInstanceMigration
+
+	// get volume from app context
+	volume, err := schedDriver.GetVolumes(virtualMachineCtx)
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		testPod, err := getVMPod(virtualMachineCtx, volume[0])
+		if err != nil {
+			return "", false, err
+		}
+
+		//Get the node where the vm is scheduled after the migration
+		nodeNameAfterMigration, err := GetNodeOfVM(vms[0])
+		if err != nil {
+			return "", false, err
+		}
```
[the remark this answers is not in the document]
spai-px: Yeah we can get it from the testPod itself. Changed it to `nodeNameAfterMigration := testPod.Spec.NodeName`.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		if nodeName == nodeNameAfterMigration {
+			return "", true, fmt.Errorf("VM pod live migrated [%s] in namespace [%s] but is still on the same node [%s]", testPod.Name, testPod.Namespace, nodeName)
+		}
+		log.InfoD("VM pod live migrated to node: [%s] in namespace [%s] is in running state", nodeNameAfterMigration, testPod.Namespace)
```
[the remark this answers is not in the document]
spai-px: Yeah changed the log message to `log.InfoD("VM pod live migrated to node: [%s]", nodeNameAfterMigration)`

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
 
 	return nil
 }
+
+// StartAndWaitForVMIMigration starts the VM migration and waits for the VM to be in running state in the new node
+func StartAndWaitForVMIMigration(virtualMachineCtx *scheduler.Context, ctx context1.Context, schedDriver scheduler.Driver) (bool, error) {
```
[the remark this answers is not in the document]
spai-px: Yeah bool is unecessary since we can catch it from the error itself changed the signature of the function.

drivers/kubevirt/kubevirt.go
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+			return false, fmt.Errorf("failed to inspect volume [%s]: %w", vol.ID, err)
+		}
+		nodeIpAttachedOn := volInspect.AttachedOn
+		nodeNameAttachedOn, err := node.GetNodeByIP(nodeIpAttachedOn)
+		if err != nil {
+			return false, fmt.Errorf("failed to get node name by IP [%s]: %w", nodeIpAttachedOn, err)
+		}
+		log.Infof("Volume [%s] is attached on node [%s]", vol.ID, nodeNameAttachedOn.Name)
+		if nodeNameAttachedOn.Name != vmNodeName {
+			return false, fmt.Errorf("volume [%s] is not attached on node [%s]", vol.ID, vmNodeName)
```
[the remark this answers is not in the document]
spai-px: Done added more context.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		}
+
+		// Check if all the replicas are attached to the same node
```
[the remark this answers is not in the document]
spai-px: Done.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		replicaset := volInspect.ReplicaSets
+
+		// check if the replicaset size is same as the global replicaset size
+		if len(replicaset) != len(globalReplicSet) {
+			return false, fmt.Errorf("replicaset count mismatch for volume [%s]", vol.ID)
```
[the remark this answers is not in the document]
spai-px: Added logs to the mismatched volume.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		}
+		// The criteria to call the bind mount successful is to check if the replicaset of all volumes should be same
+		// check if the replicaset values is same as the global replicaset
+		// Here we are considering globalreplicaset to be source of truth for comparision
+		replicaColocated, err := IsNodeInReplicaSet(replicaset, globalReplicSet)
```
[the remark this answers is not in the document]
spai-px: Done. Changed name of the function.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,313 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		if err != nil {
+			return false, err
+		}
+		if !replicaColocated {
+			return false, fmt.Errorf("replicaset mismatch for volume [%s]", vol.ID)
```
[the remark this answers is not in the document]
spai-px: Done.

tests/basic/kubevirt_test.go
spai-px: Added check for this as well.

drivers/kubevirt/kubevirt.go
spai-px: Done.

drivers/kubevirt/kubevirt.go
spai-px: Changed it to "" will check if this works.

drivers/kubevirt/kubevirt.go
spai-px: Added err

drivers/kubevirt/kubevirt.go
spai-px: Done. Added more info.

drivers/kubevirt/kubevirt.go
spai-px: Removed this and implemented a better readable function, using the same strat used above.

drivers/kubevirt/kubevirt.go
spai-px: Done. made the function return only error.

drivers/kubevirt/kubevirt.go
spai-px: Moved the logic inside the for loop to another function named IsVolumeBindMounted

vendor/modules.txt
spai-px: we will upgrade later, as of now its ok.

drivers/kubevirt/kubevirt.go
spai-px: Moved volume colocation logic to another function `VerifyVolumeReplicaCollocation`.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,330 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		return err
+	}
+
+	// Get the namespace and name of the VM
+	if len(vms) == 0 {
+		return fmt.Errorf("no VMs found for VM [%s] in namespace [%s]", virtualMachineCtx.App.Key, virtualMachineCtx.App.NameSpace)
+	}
+	vmiNamespace := vms[0].Namespace
+	vmiName := vms[0].Name
+	if len(vms) > 1 {
```
dbhatnagar-px, line 533: Please move this check one line above

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,330 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		migr, err = kubevirtdy.Instance().GetVirtualMachineInstanceMigration(ctx, vmiNamespace, migration.Name)
+		if err != nil {
+			return "", false, fmt.Errorf("failed to get migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+		if !migr.Completed {
+			return "", true, fmt.Errorf("waiting for migration to complete for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+
+		// wait until there is only one pod in the running state
+		testPod, err := getVMPod(virtualMachineCtx, volume[0])
```
dbhatnagar-px, line 571: Right now it will work. Please add future task to identify and use boot volume only here. Add a TODO comment with details here

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,330 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		if err != nil {
+			return "", true, err
+		}
+
+		//Get the node where the vm is scheduled after the migration
+		nodeNameAfterMigration := testPod.Spec.NodeName
```
dbhatnagar-px, line 577: This could be a race condition in a heavily loaded setup. So Keep a timeout of 5 mins for the `nodeNameAfterMigration` has changed from previous name. If it has changed then we can assume safely that new target has been selected. If not, then 5 mins is the max time we will check for it.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,330 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		return "", false, nil
+	}
+	_, err = task.DoRetryWithTimeout(t, defaultVmMountCheckTimeout, defaultVmMountCheckRetryInterval)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getVMPod(virtualMachineCtx *scheduler.Context, vol *volume.Volume) (*corev1.Pod, error) {
```
dbhatnagar-px, line 592: getVirtLauncherPodForVM

dbhatnagar-px, line 592: Add a comment for each method as well

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,330 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	if len(vols) == 0 {
+		return false, fmt.Errorf("no volumes found for VM [%s] in namespace [%s]", virtualMachineCtx.App.Key, virtualMachineCtx.App.NameSpace)
+	}
+
+	// Get the node where pod is scheduled
+	vms, err := GetAllVMsFromScheduledContexts([]*scheduler.Context{virtualMachineCtx})
+	if err != nil {
+		return false, err
+	}
+	if len(vms) != 1 {
```
dbhatnagar-px, line 629: Add a comment for now that it will work only for 1 VM per spec/context. Ok to merge as of now but pls create a future ticket

drivers/kubevirt/kubevirt.go
spai-px: Done added a single check, which passes only if len(vms)=1 else fail.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,330 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	nodeName, err := GetNodeOfVM(vms[0])
+	if err != nil {
+		return err
+	}
+	log.Infof("VM [%s] in namespace [%s] is scheduled on node [%s]", vmiName, vmiNamespace, nodeName)
+
+	// Start the VM migration
+	migration, err := kubevirtdy.Instance().CreateVirtualMachineInstanceMigration(ctx, vmiNamespace, vmiName)
+	if err != nil {
+		log.Errorf("Failed to create VM migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
```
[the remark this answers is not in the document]
spai-px: Ok done.

drivers/kubevirt/kubevirt.go
spai-px: Done. Created a ticket to do this. https://purestorage.atlassian.net/browse/PTX-23166

drivers/kubevirt/kubevirt.go
spai-px: Done added 5 minute wait for node_name check

drivers/kubevirt/kubevirt.go
spai-px: Changed function name getVmpod to `GetVirtLauncherPodForVM`

spai-px: Done.

drivers/kubevirt/kubevirt.go
spai-px: Done. Created a ticker https://purestorage.atlassian.net/browse/PTX-23167

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,350 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	if err != nil {
+		return err
+	}
+	log.Infof("VM migration created for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+
+	// wait for completion
+	var migr *kubevirtdy.VirtualMachineInstanceMigration
+
+	// get volumes from app context
+	vol, err := schedDriver.GetVolumes(virtualMachineCtx)
```
pureneelesh, line 555: vol -> vols

drivers/kubevirt/kubevirt.go
pureneelesh: @dbhatnagar-px : I don't understand your comment and why we have to wait some more after the migration object says that the migration is completed. At that point, we can simply return error if the node name has not changed. No retried are necessary. We should only keep retrying until the "migration" object says that the migration is still in progress (we are already retrying this). That should take care of longer wait time in a loaded setup.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,350 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		if err != nil {
+			return "", false, fmt.Errorf("failed to get migration for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+		if !migr.Completed {
+			return "", true, fmt.Errorf("waiting for migration to complete for VM [%s] in namespace [%s]", vmiName, vmiNamespace)
+		}
+
+		// wait until there is only one pod in the running state
+		//TODO https://purestorage.atlassian.net/browse/PTX-23166 - This is a temporary fix to get the pod of the VM
+		testPod, err := GetVirtLauncherPodForVM(virtualMachineCtx, vol[0])
```
pureneelesh, line 571: We should be able to use any volume to get the virt launcher pod. It does not have to be the root volume. Only exception I can think of is a "hotplug" volume. It may return the hotplug pod.

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,350 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+		}
+	}
+	if found.Name == "" {
+		return nil, fmt.Errorf("failed to find a running pod for volume %s", vol.ID)
+	}
+	return &found, nil
+}
+
+// IsVMBindMounted checks if the volumes are bind mounted to the VM
+func IsVMBindMounted(virtualMachineCtx *scheduler.Context, driver scheduler.Driver, volumeDriver volume.Driver, initialCheck bool) (bool, error) {
```
pureneelesh, line 635: nit: rename initialCheck -> wait since that parameter is used to decide whether to wait for the bind mount or not

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,350 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	// The criteria to call the bind mount successful is to check if the replicaset of all volumes should be same and should be locally attached to the node
+	// check if the replicaset values is same as the global replicaset
+	// Here we are considering globalreplicaset to be source of truth for comparison
+
+	for _, vol := range vols {
+		err := IsVolumeBindMounted(volumeDriver, virtualMachineCtx, vmNodeName, vol, initialCheck)
+		if err != nil {
+			return false, err
+		}
+		err = VerifyVolumeReplicaCollocation(vol, globalReplicSet, volumeDriver)
```
pureneelesh, line 675: VerifyVolumeReplicaCollocation -> AreVolumeReplicasCollacated to be consistent with the naming style of IsVolumeBindMounted

drivers/kubevirt/kubevirt.go
```diff
@@ -507,3 +515,350 @@ func AddPVCsToVirtualMachine(vm kubevirtv1.VirtualMachine, pvcs []*corev1.Persis
...
+	if err != nil {
+		return fmt.Errorf("failed to inspect volume [%s]: %w", vol.ID, err)
+	}
+
+	replicaset := volInspect.ReplicaSets
+
+	// check if the replicaset size is same as the global replicaset size
+	if len(replicaset) != len(globalReplicSet) {
+		return fmt.Errorf("replicaset count mismatch for volume [%s] and for volume [%s]", vol.ID, volInspect.Id)
+	}
```
pureneelesh, line 858: this check should be part of ReplicaSetsMatch function?

pureneelesh (approved):

drivers/kubevirt/kubevirt.go
spai-px: Done moved it to ReplicaSetsMatch

drivers/kubevirt/kubevirt.go
spai-px: Done.

drivers/kubevirt/kubevirt.go
spai-px: Done.

drivers/kubevirt/kubevirt.go
dbhatnagar-px: You are right Neelesh. We have removed this check from here. Thanks.

drivers/kubevirt/kubevirt.go
dbhatnagar-px: We need to check for boot volume as data disks could be attached to other VMs as well - running on some other node.

dbhatnagar-px (approved):
Lgtm

PR #2315, opened 2024-04-05
Fixed the kubevirt tests for NFS backup location

mkoppal-px:
Few kubevirt tests were failing when running with NFS backup location because the `GetVolumes` function was trying to get all PVCs in the namespace when it finds a spec of type Virtual Machine. In case of NFS, backup creates a pvc for mounting the nfs backup location and even that was being inspected but because it was in terminating state it was failing.

Added logic to exclude the volumes created by px-backup in application namespace.

Closes #

https://aetos.pwx.purestorage.com/resultSet/testSetID/602647
Ignore the failed test cases because they have failed because of long running tests which will be fixed in the next PR


kphalgun-px (approved):
LGTM

kshithijiyer-px (approved):
LGTM

dbhatnagar-px (approved):
Lgtm

PR #2331, opened 2024-04-09
Db data validation kubevirt

dbhatnagar-px:

Closes #

### DIFF
--- a/tests/kubevirt_helper.go
+++ b/tests/kubevirt_helper.go
@@ -31,6 +31,7 @@ const (
 var (
 	defaultVmMountCheckTimeout       = 15 * time.Minute
 	defaultVmMountCheckRetryInterval = 30 * time.Second
+	k8sKubevirt                      = kubevirt.Instance()
 )
 
 // AddDisksToKubevirtVM is a function which takes number of disks to add and adds them to the kubevirt VMs passed (Please provide size in Gi)
@@ -484,7 +485,6 @@ func AreVolumeReplicasCollocated(vol *volume.Volume, globalReplicSet []*api.Repl
 }
 
 func CreateConfigMap() error {
-	k8sCore := core.Instance()
 	// Check if a config map named kubevirt-creds exist
 	configMap, err := k8sCore.GetConfigMap("kubevirt-creds", "default")
 	log.Infof("configMap: %v", configMap)
@@ -557,7 +557,6 @@ func ValidateFileIntegrityInVM(virtualMachines []*scheduler.Context, namespace s
 }
 
 func ListEvents(namespace string) error {
-	k8sCore = core.Instance()
 	eventList, err := k8sCore.ListEvents(namespace, metav1.ListOptions{})
 	if err != nil {
 		log.Infof("Failed to list events in namespace %s: %v", namespace, err)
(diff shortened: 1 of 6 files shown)


tests/kubevirt_helper.go
```diff
@@ -477,24 +485,163 @@ func AreVolumeReplicasCollocated(vol *volume.Volume, globalReplicSet []*api.Repl
...
+
+			_, err = k8sKubevirt.UpdateVirtualMachine(vm)
+			if err != nil {
+				return fmt.Errorf("failed to update VM [%s]: %w", vm.Name, err)
+			}
+
+			err = RestartKubevirtVM(v.Name, v.Namespace, true)
+			if err != nil {
+				return err
+			}
```
kphalgun-px, line 635: Is restart required for Hotplug volumes?

tests/kubevirt_helper.go
```diff
@@ -477,24 +485,163 @@ func AreVolumeReplicasCollocated(vol *volume.Volume, globalReplicSet []*api.Repl
...
+			log.InfoD("Sleep for 5mins for vm to come up")
+			time.Sleep(5 * time.Minute)
+			NewDiskCountOutput, err := GetNumberOfDisksInVM(v)
```
kphalgun-px, line 638: How about we handle error?

tests/kubevirt_helper.go
```diff
@@ -477,24 +485,163 @@ func AreVolumeReplicasCollocated(vol *volume.Volume, globalReplicSet []*api.Repl
...
+	log.Infof("Events in namespace %s:", namespace)
+	for _, event := range eventList.Items {
+		log.Infof("Time: %v, Event: %s, Type: %s, Reason: %s, Object: %s/%s, Message: %s",
+			event.FirstTimestamp, event.Name, event.Type, event.Reason, event.InvolvedObject.Kind, event.InvolvedObject.Name, event.Message)
+	}
+	return nil
+}
+
+// HotAddPVCsToKubevirtVM hot adds disk to a running VM
+func HotAddPVCsToKubevirtVM(virtualMachines []*scheduler.Context, numberOfDisks int, size string) error {
```
kphalgun-px, line 579: Can we make use of AddPVCsToVirtualMachin? We may need to include a Hotpluggable parameter.

tests/kubevirt_helper.go
```diff
@@ -477,24 +485,163 @@ func AreVolumeReplicasCollocated(vol *volume.Volume, globalReplicSet []*api.Repl
...
+func ValidateFileIntegrityInVM(virtualMachines []*scheduler.Context, namespace string) error {
+	log.Infof("Validating file integrity in namespace %s", namespace)
+	validateFilesCmd := "cd ~/testfiles && md5sum -c ~/file_checksums.md5"
+	for _, appCtx := range virtualMachines {
+		vms, err := GetAllVMsFromScheduledContexts([]*scheduler.Context{appCtx})
+		if err != nil {
+			return err
+		}
+		for _, v := range vms {
+			_, err := RunCmdInVM(v, validateFilesCmd, context1.TODO())
```
kphalgun-px, line 554: RunCmdInVM expects the VM to have a username specified in the label, but it seems that `kubevirt-fio-pvc-clone` has the username set as an annotation. Can you please verify this?

dbhatnagar-px: It's there in labels as well.

tests/kubevirt_helper.go
dbhatnagar-px: As of now yes - this is the behavior via API.

tests/kubevirt_helper.go
dbhatnagar-px: Good catch - I missed it. Let me do that

lsrinivas-pure (approved):
LGTM

PR #2365, opened 2024-04-16
PTX-23399: Replace pkill with kill.

pnookala-px:
pkill the arguments are process name.
kill is for killing a process pid. We want to killl a process using pid, so use kill.


lsrinivas-pure (approved):
LGTM

PR #2378, opened 2024-04-17
PTX-23457: minor nits

pnookala-px:
1. Fix kill test that was accidentally overwritten.
2. Decommission tests should not validate.


tests/basic/legacy_shared_sharedv4_service_migration_test.go
```diff
@@ -531,8 +531,10 @@ var _ = Describe("{LegacySharedToSharedv4ServiceNodeDecommission}", func() {
 			for _, ctx := range contexts {
 				checkMapOfPods(podMap, ctx)
 			}
-			ValidateApplications(contexts)
+			// Don't validate apps.
```
lsrinivas-pure, line 534: can you use this function PrereqForNodeDecomm for moving replicas before docomm

pnookala-px: Will address this in new ticket.

lsrinivas-pure (approved):
LGTM

PR #2380, opened 2024-04-17
PTX-23460 Vendor update of libopenstorage/operator package for Telemetry output check when PX in debug mod fix

nikolaypopov:
Vendor update of libopenstorage/operator package for Telemetry output check when PX in debug mod fix


Pure-AdamuKaapan (approved):
LGTM! Thanks for the quick turnaround

PR #2417, opened 2024-04-24
Ptx 20883 1

vprabhakar-px:
FBDA takes time to update stats , currently time set is 5 min , but after checking seems like update stats is taking more time . hence increasing timeout by 15 min for Stats to update ( which worked fine )


lsrinivas-pure (approved):
LGTM

PR #2424, opened 2024-04-25
Improve node reboot kubevirt

Rohit-PX:
* Add extra checks post reboot of nodes to verify that the VM pod has moved to a different node other than the node that was rebooted
* All the while validate that bind mount is still preserved

Closes # https://purestorage.atlassian.net/browse/PTX-23439


tests/kubevirt_helper.go
```diff
@@ -762,3 +763,19 @@ func DeployVMTemplatesAndValidate() error {
 	_, err = task.DoRetryWithTimeout(waitForCompletedAnnotations, importerPodCompletionTimeout, importerPodRetryInterval)
 	return err
 }
+
+func AddToVMNodeMap(vmName, nodeName string, vmNodeMap *map[string]string) {
+	(*vmNodeMap)[vmName] = nodeName
+}
+
+func ValidateAndUpdateVMNode(vmName, nodeName string, vmNodeMap *map[string]string) error {
```
pureneelesh, line 771: There should not be any need to pass map as a pointer unless the function is returning a brand new map.

tests/kubevirt_helper.go
```diff
@@ -762,3 +763,19 @@ func DeployVMTemplatesAndValidate() error {
 	_, err = task.DoRetryWithTimeout(waitForCompletedAnnotations, importerPodCompletionTimeout, importerPodRetryInterval)
 	return err
 }
+
+func AddToVMNodeMap(vmName, nodeName string, vmNodeMap *map[string]string) {
```
pureneelesh, line 767: nit: maybe get rid of this function. It does not seem to do much.

tests/kubevirt_helper.go
pureneelesh, line 771: nit: rename the function to indicate that it is expecting node to change e.g. ValidateVMNodeChanged

pureneelesh (approved):

tests/kubevirt_helper.go
Rohit-PX: Fixed.

tests/kubevirt_helper.go
Rohit-PX: Fixed.

tests/kubevirt_helper.go
Rohit-PX: Fixed.

PR #2425, opened 2024-04-25
PTX-23652: be resilient to failures of GetVolumes() call

pnookala-px:
GetVolumes() call is supposed to be made when the pods in the context are stable. In case of migration, the pods can get restarted during the migration process and the call to gEtVolumes() can fail. Be resilient to these failures and don't fail the test case if this call fails, instead treat the failure as having a shared volume. This call will happen again and will get the correct count in further iterations.

### DIFF
--- a/tests/basic/legacy_shared_sharedv4_service_migration_test.go
+++ b/tests/basic/legacy_shared_sharedv4_service_migration_test.go
@@ -90,21 +90,23 @@ func setMigrateLegacySharedToSharedv4Service(on bool) {
 func getLegacySharedVolumeCount(contexts []*scheduler.Context) int {
 	count := 0
 	for _, ctx := range contexts {
-		vols, err := Inst().S.GetVolumes(ctx)
-		if err == nil {
-			for _, v := range vols {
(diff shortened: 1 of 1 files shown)


tests/basic/legacy_shared_sharedv4_service_migration_test.go
```diff
@@ -91,13 +91,20 @@ func getLegacySharedVolumeCount(contexts []*scheduler.Context) int {
...
-		log.FailOnError(err, "error geting volumes used by app")
-		for _, v := range vols {
-			vol, err := Inst().V.InspectVolume(v.ID)
-			log.FailOnError(err, "Failed to inspect volume %v", v.ID)
-			if vol.Spec.Shared {
-				count++
+		if err == nil {
+			for _, v := range vols {
+				vol, err := Inst().V.InspectVolume(v.ID)
+				log.FailOnError(err, "Failed to inspect volume %v", v.ID)
```
lsrinivas-pure, line 97: can we try retry logic here?

lsrinivas-pure (approved):
LGTM

PR #2432, opened 2024-04-26
Adding check to ignore Volume attaches in case of pure file volume

vprabhakar-px:
getting file attachment only if PVC is not pure_file


dbhatnagar-px (approved):
Lgtm

PR #2455, opened 2024-05-03
Add test case to bring down network interface of FA Controller

vprabhakar-px:
bring down network interface and bring back

### DIFF
--- a/tests/basic/pure_test.go
+++ b/tests/basic/pure_test.go
@@ -2879,3 +2879,157 @@ var _ = Describe("{FBDAMultiTenancyBasicTest}", func() {
 		})
 	})
 })
+
+var _ = Describe("{FADAPodRecoveryDisableDataPortsOnFA}", func() {
+
+	/*
+			PTX : https://purestorage.atlassian.net/browse/PTX-23763
+		Verify that FA Pods Recovers after bringing back network Interface down on all FA's
+
+	*/
+	JustBeforeEach(func() {
+		log.Infof("Starting Torpedo tests ")
+		StartTorpedoTest("FADAPodRecoveryDisableDataPortsOnFA",
+			"Verify Pod Recovers from RO mode after Bounce after blocking network interface from FA end",
+			nil, 0)
+	})
+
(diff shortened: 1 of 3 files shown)


tests/basic/pure_test.go
```diff
@@ -2879,3 +2879,157 @@ var _ = Describe("{FBDAMultiTenancyBasicTest}", func() {
...
+			"Verify Pod Recovers from RO mode after Bounce after blocking network interface from FA end",
+			nil, 0)
+	})
+
+	itLog := "FADAPodRecoveryDisableDataPortsOnFA"
+	It(itLog, func() {
+
+		var contexts []*scheduler.Context
+		var k8sCore = core.Instance()
+		//podNodes := []*corev1.Pod{}
```
lsrinivas-pure, line 2902: delete if not required

tests/common.go
```diff
@@ -12170,3 +12170,128 @@ func WaitForVolumeClean(vol *volume.Volume) error {
...
+		return nil, fmt.Errorf("Unable to get Px Pure Secret")
+	}
+
+	if len(pxPureSecret.Arrays) > 0 {
+		return pxPureSecret.Arrays, nil
+	}
+	return nil, fmt.Errorf("Failed to list FA Arrays ")
+}
+
+func FlashArrayGetIscsiPorts() (map[string][]string, error) {
```
lsrinivas-pure, line 12193: can you move all these to pureutils as these are specifc to FA/FB

vprabhakar-px: sure , will move it to pure utils

tests/basic/pure_test.go
vprabhakar-px: done

tests/common.go
vprabhakar-px: @lsrinivas-pure , this is the common function that is derived from functions listed in pure utils , I guess this should be in common utils . 
will be using this functions on multiple scripts belongs to FA going forward

lsrinivas-pure (approved):

PR #2485, opened 2024-05-09
[PWX-37028] Added FBDAMultiTenancyUpdatePureNFSEnpoint

vinayakshnd:

Closes #

### DIFF
--- a/drivers/volume/aws/aws.go
+++ b/drivers/volume/aws/aws.go
@@ -2,10 +2,14 @@ package aws
 
 import (
 	"fmt"
+
 	"github.com/libopenstorage/openstorage/api"
 
+	"github.com/portworx/torpedo/drivers/node"
+	"github.com/portworx/torpedo/drivers/volume"
 	torpedovolume "github.com/portworx/torpedo/drivers/volume"
 	"github.com/portworx/torpedo/drivers/volume/portworx/schedops"
+	"github.com/portworx/torpedo/pkg/errors"
 	"github.com/portworx/torpedo/pkg/log"
 )
 
@@ -52,6 +56,24 @@ func (d *aws) InspectCurrentCluster() (*api.SdkClusterInspectCurrentResponse, er
 func (i *aws) DeleteSnapshotsForVolumes(volumeNames []string, clusterProviderCredential string) error {
 	log.Warnf("DeleteSnapshotsForVolumes function has not been implemented for volume driver - %s", i.String())
 	return nil
+}
+
+// UpdateFBDANFSEndpoint updates the NFS endpoint for a given FBDA volume
+func (d *aws) UpdateFBDANFSEndpoint(volumeName string, newEndpoint string) error {
+	log.Warnf("UpdateFBDANFSEndpoint function has not been implemented for volume driver - %s", d.String())
+	return &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "InspectVolume()",
+	}
+}
+
+// ValidatePureFBDAMountSource checks that, on all the given nodes, all the provided FBDA volumes are mounted using the expected IP
+func (d *aws) ValidatePureFBDAMountSource(nodes []node.Node, vols []*volume.Volume, expectedIP string) error {
+	log.Warnf("ValidatePureFBDAMountSource function has not been implemented for volume driver - %s", d.String())
+	return &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "ValidatePureFBDAMountSource()",
+	}
 }
 
 func (d *aws) Init(sched, nodeDriver, token, storageProvisioner, csiGenericDriverConfigMap string) error {
--- a/drivers/volume/aws/aws.go
+++ b/drivers/volume/aws/aws.go
@@ -63,7 +63,7 @@ func (d *aws) UpdateFBDANFSEndpoint(volumeName string, newEndpoint string) error
 	log.Warnf("UpdateFBDANFSEndpoint function has not been implemented for volume driver - %s", d.String())
 	return &errors.ErrNotSupported{
 		Type:      "Function",
-		Operation: "InspectVolume()",
+		Operation: "UpdateFBDANFSEndpoint()",
 	}
 }
 
--- a/drivers/volume/ocp/ocp.go
+++ b/drivers/volume/ocp/ocp.go
@@ -2,7 +2,10 @@ package ocp
 
 import (
 	"fmt"
+
 	"github.com/libopenstorage/openstorage/api"
+	"github.com/portworx/torpedo/drivers/node"
+	"github.com/portworx/torpedo/drivers/volume"
 	torpedovolume "github.com/portworx/torpedo/drivers/volume"
 	"github.com/portworx/torpedo/drivers/volume/portworx"
 	"github.com/portworx/torpedo/drivers/volume/portworx/schedops"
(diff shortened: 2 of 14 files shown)


drivers/volume/ocp/ocp.go
```diff
@@ -101,6 +104,24 @@ func (o *ocp) InspectVolume(name string) (*api.Volume, error) {
 	}
 }
 
+// UpdateFBDANFSEndpoint updates the NFS endpoint for a given FBDA volume
```
vprabhakar-px, line 107: we support FA with OCP, can we also implement the functions here ?

tests/basic/pure_test.go
```diff
@@ -2880,6 +2863,196 @@ var _ = Describe("{FBDAMultiTenancyBasicTest}", func() {
...
+
+		// save the original custom app configs
+		origCustomAppConfigs = make(map[string]scheduler.AppConfig)
+		for appName, customAppConfig := range Inst().CustomAppConfig {
+			origCustomAppConfigs[appName] = customAppConfig
+		}
+
+		// update the custom app config with empty string for Pure NFS endpoint
+		// So that app uses NFS endpoint from pure.json secret.
+		Inst().AppList = []string{customConfigAppName}
```
lsrinivas-pure, line 2885: can you add logic to revert to original app_list . If we run this test along with other tests we might need to apps provided as part of the test run for other tests

drivers/volume/ocp/ocp.go
pnookala-px: This is FBDA endpoint. It can be practically be on any environment, for the purpose of this test case, we don't need to do this.

tests/basic/pure_test.go
```diff
@@ -2880,6 +2863,196 @@ var _ = Describe("{FBDAMultiTenancyBasicTest}", func() {
...
 
+var _ = Describe("{FBDAMultiTenancyUpdatePureNFSEnpoint}", func() {
+	var contexts []*scheduler.Context
+	var customConfigAppName string
+	var origCustomAppConfigs, customAppConfigs map[string]scheduler.AppConfig
+	testName := "fbda-mt-update-endp"
+
+	JustBeforeEach(func() {
+		StartTorpedoTest("FBDAMultiTenancyUpdatePureNFSEnpoint", "Validate Pure NFS endpoint can be changed using pxctl", nil, 0)
+		Step("setup credential necessary for cloudsnap", createCloudsnapCredential)
```
pnookala-px, line 2874: We don't need to setup cloudsnap for Fbda tests.

tests/basic/pure_test.go
```diff
@@ -2880,6 +2863,196 @@ var _ = Describe("{FBDAMultiTenancyBasicTest}", func() {
...
+		// remove any keys that are not present in the orig map
+		for appName := range Inst().CustomAppConfig {
+			if _, ok := origCustomAppConfigs[appName]; !ok {
+				delete(Inst().CustomAppConfig, appName)
+			}
+		}
+		log.Infof("JustAfterEach restoring Inst().CustomAppConfig = %v", Inst().CustomAppConfig)
+		err := Inst().S.RescanSpecs(Inst().SpecDir, Inst().V.String())
+		Expect(err).NotTo(HaveOccurred(), "Failed to rescan specs from %s", Inst().SpecDir)
+		Step("delete credential used for cloudsnap", deleteCloudsnapCredential)
```
pnookala-px, line 2968: cloudsnap need not be setup and tore down. It is waste of time in this test case.

tests/basic/pure_test.go
```diff
@@ -2880,6 +2863,196 @@ var _ = Describe("{FBDAMultiTenancyBasicTest}", func() {
...
+		origCustomAppConfigs = make(map[string]scheduler.AppConfig)
+		for appName, customAppConfig := range Inst().CustomAppConfig {
+			origCustomAppConfigs[appName] = customAppConfig
+		}
+
+		// update the custom app config with empty string for Pure NFS endpoint
+		// So that app uses NFS endpoint from pure.json secret.
+		Inst().AppList = []string{customConfigAppName}
+		Inst().CustomAppConfig[customConfigAppName] = scheduler.AppConfig{
+			StorageClassPureNfsEndpoint: "",
```
pnookala-px, line 2887: In the test configure pxctl cluster options update --fb-stats-expiry-duration to 1. This will make sure we don't sleep too much time.

drivers/volume/aws/aws.go
```diff
@@ -54,6 +58,24 @@ func (i *aws) DeleteSnapshotsForVolumes(volumeNames []string, clusterProviderCre
 	return nil
 }
 
+// UpdateFBDANFSEndpoint updates the NFS endpoint for a given FBDA volume
+func (d *aws) UpdateFBDANFSEndpoint(volumeName string, newEndpoint string) error {
+	log.Warnf("UpdateFBDANFSEndpoint function has not been implemented for volume driver - %s", d.String())
+	return &errors.ErrNotSupported{
+		Type:      "Function",
+		Operation: "InspectVolume()",
```
kphalgun-px, line 66: How about?

```go
	Operation: "UpdateFBDANFSEndpoint()",
```

drivers/volume/portworx/portworx.go
```diff
@@ -1900,6 +1900,29 @@ func (d *portworx) ValidateVolumeInPxctlList(volumeName string) error {
...
+	cmd := fmt.Sprintf("%s --pure_nfs_endpoint %s %s", pxctlVolumeUpdate, newEndpoint, volumeName)
+	_, err := d.nodeDriver.RunCommandWithNoRetry(
+		nodes[0],
+		cmd,
+		node.ConnectionOpts{
+			Timeout:         crashDriverTimeout,
+			TimeBeforeRetry: defaultRetryInterval,
+		})
+	if err != nil {
+		return fmt.Errorf("failed setting FBDA NFS endpoint for volume '%s' to '%s', Err: %v", volumeName, newEndpoint, err)
```
kphalgun-px, line 1914: Can we also include node name or ID in the log message?

tests/basic/pure_test.go
```diff
@@ -2880,6 +2863,196 @@ var _ = Describe("{FBDAMultiTenancyBasicTest}", func() {
...
+			}
+
+			// Check that we are using the correct address
+			dash.VerifyFatal(strings.Contains(line, newNFSEndpoint), true,
+				fmt.Sprintf("Verify mount line for volume [%s] contains new NFS endpoint '%s'", foundVol, n.Name))
+		}
+	}
+}
+
+func scaleAppToZero(ctx *scheduler.Context) (map[string]int32, error) {
```
kphalgun-px, line 3010: I wonder if this scale function works on deployment based apps like `vdbench-fbda-ep`.

drivers/volume/ocp/ocp.go
vinayakshnd: As answered by @pnookala-px this is not required .

tests/basic/pure_test.go
vinayakshnd: @pnookala-px actually ValidateApplicationsPureSDK() function (as part of -ve tests) invokes cloudsnaps for FBDA volumes and expect the cloudsnaps to fail with proper error saying cloudsnaps are not supported with FBDA volumes.  For this -ve test, it requires creds to be created and then cloudsnap to fail with proper error and not because of creds not present error.

tests/basic/pure_test.go
vinayakshnd: actually ValidateApplicationsPureSDK() function (as part of -ve tests) invokes cloudsnaps for FBDA volumes and expect the cloudsnaps to fail with proper error saying cloudsnaps are not supported with FBDA volumes. For this -ve test, it requires creds to be created and then cloudsnap to fail with proper error and not because of creds not present error.

tests/basic/pure_test.go
vinayakshnd: @kphalgun-px this uses k8s scheduler driver's `ScaleApplication` function which supports deployments also. 
https://github.com/portworx/torpedo/blob/master/drivers/scheduler/k8s/k8s.go#L5221

tests/basic/pure_test.go
vinayakshnd: @pnookala-px I talked to Vishwas about this and learnt that our test framework should use the same default timeout that is set out of box and customer will also have that in their env.

drivers/volume/aws/aws.go
vinayakshnd: Done

drivers/volume/portworx/portworx.go
vinayakshnd: Done

tests/basic/pure_test.go
kphalgun-px: Thank you.

kphalgun-px (approved):
LGTM

tests/basic/pure_test.go
vinayakshnd: Removed modifing of AppList it self.

vprabhakar-px (approved):
LGTM

dbhatnagar-px (approved):
Lgtm

PR #2526, opened 2024-05-21
adding more logs for PDB validation

svijaykumar-px:
**What this PR does / why we need it**: Adding more logs to PDB validation to help debug the issues better. This code adds a log when the PDB minAvailable value changes when not expected to. It also raises a warning when more nodes are down than expected logging value of storage nodes that are currently up.

Closes # PTX-24237

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/devans10/pugo/flasharray"
@@ -982,6 +983,7 @@ func ValidatePDB(pdbValue int, allowedDisruptions int, initialNumNodes int, isCl
 
 	Step("Validate PDB minAvailable for px storage", func() {
 		if currentPdbValue != pdbValue {
+			log.Warnf("Current PDB minAvailable has changed to %d", currentPdbValue)
 			err := fmt.Errorf("PDB minAvailable value has changed. Expected: %d, Actual: %d", pdbValue, currentPdbValue)
 			processError(err, errChan...)
 		}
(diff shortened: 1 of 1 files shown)


tests/common.go
```diff
@@ -1002,6 +1004,13 @@ func ValidatePDB(pdbValue int, allowedDisruptions int, initialNumNodes int, isCl
 		} else {
 			currentNumNodes := len(nodes.([]*opsapi.StorageNode))
 			if allowedDisruptions < initialNumNodes-currentNumNodes {
+				log.Warnf("Number of nodes down is more than allowed disruptions [ %d ], total storage nodes : [ %d ] and current count of storage nodes are: [ %d ]", allowedDisruptions, initialNumNodes, currentNumNodes)
+				var currNodesName []string
+				for _, node := range nodes.([]*opsapi.StorageNode) {
+					currNodesName = append(currNodesName, node.SchedulerNodeName)
```
jainsmit, line 1010: Would it return storage nodes which are down or where a node is deleted but node-id is showing down ? Can we also add kubernetes worker node list here.

The issue which we have seen recently was a sequential upgrade and values are not matching..

svijaykumar-px: This will return the list of kubernetes nodes where portworx is currently up. Should I return nodes that are down instead? In that case it will be the kubernetes nodes where initially portworx was up and running but currently down.

jainsmit: We need all storage nodes where PX is up and down. Also need list of worker nodes currently present in a cluster so that will know where PX is up or down or node delete from the cluster because of the upgrade..

svijaykumar-px: I am still not clear, when you say storage node where PX is up/down you mean the node ID? or the kubernetes node name where that px node is run? It is easier to debug and understand using kubernetes node names.
So let me know if this sounds good -

1.  kubernetes node names where PX is running (can also include the node ID)
2. kubernetes node where PX is down (can include node ID)
3. list of all kubernetes nodes

jainsmit: Sorry about the confusion. Yes it sounds good. Need both px node id and K8S node name. More like output similar to `pxctl status` command.  

When PDB is not matching what I need to debug is PDB value set in STC and storage nodes(nodes id and name)  in a cluster. Later need to identify if some nodes are recreated or deleted during Anthos upgrade (It can be find by listing k8s nodes)

PR #2541, opened 2024-05-27
PWX-37401 Test stopping Portworx on a node for 10 minutes, then starting it again

sayalasomayajula-px:
Deploys fada, facd apps and stops a node randomly for 10 mins and check if apps are deployed or not and node is up after 10 mins

Closes #
PWX-37401

PASS LOGS:https://jenkins.pwx.dev.purestorage.com/blue/organizations/jenkins/Users%2Fkrishna%2Ftp-pxe-byoc/detail/tp-pxe-byoc/3144/pipeline
https://jenkins.pwx.dev.purestorage.com/blue/organizations/jenkins/Users%2Fkrishna%2Ftp-pxe-byoc/detail/tp-pxe-byoc/3179/pipeline/

### DIFF
--- a/tests/basic/pure_test.go
+++ b/tests/basic/pure_test.go
@@ -4598,6 +4598,12 @@ var _ = Describe("{CreateCloneOfTheFADAVolume}", func() {
 })
 
 var _ = Describe("{DeployFADAAndFBDAAppsAndStopPortworx}", func() {
+	/*
+		https://purestorage.atlassian.net/browse/PTX-37401
+		1.Deploy FADA and FACD apps and parallely and stop portworx for 10 mins
+		2.After 10 mins make it up and check if the pods are running
+		3.Destroy the apps
+	*/
 	JustBeforeEach(func() {
 		StartTorpedoTest("DeployFADAAndFBDAAppsAndStopPortworx",
 			"Deploy FADA and FBDA apps and parallely stop portworx for 10 mins and after 10 min make it up and check if the pods are running",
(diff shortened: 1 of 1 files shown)


lsrinivas-pure (approved):
LGTM

tests/basic/pure_test.go
```diff
@@ -4340,3 +4340,86 @@ var _ = Describe("{CreateCloneOfTheFADAVolume}", func() {
...
+
+var _ = Describe("{DeployFADAAndFBDAAppsAndStopPortworx}", func() {
+	/*
+		https://purestorage.atlassian.net/browse/PTX-37401
+		1.Deploy FADA and FACD apps and parallely and stop portworx for 10 mins
+		2.After 10 mins make it up and check if the pods are running
+		3.Destroy the apps
+	*/
+	JustBeforeEach(func() {
+		StartTorpedoTest("DeployFADAAndFBDAAppsAndStopPortworx",
```
Pure-AdamuKaapan, line 4352: This test shouldn't be bound specifically to FADA and FBDA, nor should it have a set app list. Can we make this more like the existing tests, where it will keep the same node reboot logic but work with more apps?

Pure-AdamuKaapan (changes requested):
The test logic itself looks good (the stopping, starting, validate node up, validate apps), but let's decouple it from specific applications and from FADA/FBDA. All the methods we're using here should work with any driver I believe, not just Pure.

Pure-AdamuKaapan (approved):
LGTM!

PR #2551, opened 2024-05-29
Helper methods to add cluster wide PSA in Rancher

snigdha-px:
Added the following helper methods for Rancher PSA:
1)UpdateClusterWidePSA
2)GetPodSecurityAdmissionConfigurationTemplateList
3)GetCurrentClusterWidePSA
4)GetRKEClusterList
5)CreateCustomPodSecurityAdmissionConfigurationTemplate

Closes #https://purestorage.atlassian.net/browse/PB-7152
https://purestorage.atlassian.net/browse/PB-7154

Testcase "DummyPSATestcase" is added temporarily to verify the helper methods
Will remove it once I start writing the testcase consuming these helper methods

https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/5230/


tests/backup_helper.go
```diff
@@ -3762,6 +3762,23 @@ func IsPresent(dataSlice interface{}, data interface{}) bool {
 	return false
 }
 
+// AppendList append only unique items of one list into another list
```
kshithijiyer-px, line 3765: How about this?
```
func AppendList[T comparable](dataSlice1 []T, dataSlice2 []T) []T {
	// Estimate the required size of the map
	totalLength := len(dataSlice1) + len(dataSlice2)
	map1 := make(map[T]bool, totalLength)
	list1 := make([]T, 0, totalLength) // Initialize list1 with the total capacity

	for _, element := range dataSlice1 {
		if !map1[element] {
			map1[element] = true
			list1 = append(list1, element)
		}
	}
	for _, element := range dataSlice2 {
		if !map1[element] {
			map1[element] = true
			list1 = append(list1, element)
		}
	}
	return list1
}
```

snigdha-px: For smaller number of data, I think this will not make too much of a difference
Also we reduced one for loop but added 2 if statement which will run each time in for loop
Let me know your thoughts @kshithijiyer-px

drivers/scheduler/rke/rke.go
kphalgun-px, line 1: How about we return error at the end?

Instead of

```go
func (r *Rancher) GetCurrentClusterWidePSA(clusterName string) (error, string)
```

How about?

```go
func (r *Rancher) GetCurrentClusterWidePSA(clusterName string) (string, error)
```

Similarly for GetPodSecurityAdmissionConfigurationTemplateList

tests/backup_helper.go
kshithijiyer-px: I feel it's better to keep it optimised from day one we never know where all we'll end up using it. If we use it in scale tests this might become helpful there. I'll leave the decision to you if you wish to change it this way. or not.

drivers/scheduler/rke/rke.go
snigdha-px: done

tests/backup_helper.go
snigdha-px: done

kshithijiyer-px (approved):
LGTM

drivers/scheduler/rke/rke.go
```diff
@@ -459,6 +460,78 @@ func (r *Rancher) SetASGClusterSize(perZoneCount int64, timeout time.Duration) e
...
+			Exemptions: psaTemplateExemptions,
+		},
+	}
+	log.InfoD("Create custom PSA template: %v with template defaults %v and template exemptions %v", psaTemplateName, psaTemplateDefaults, psaTemplateExemptions)
+	_, err := r.client.PodSecurityAdmissionConfigurationTemplate.Create(PSAInterface)
+	return err
+}
+
+// GetRKEClusterList returns the list of RKE clusters added to Rancher
+func (r *Rancher) GetRKEClusterList() (error, []string) {
```
kphalgun-px, line 479: How about we put error in the end?

snigdha-px: done

drivers/scheduler/rke/rke.go
```diff
@@ -459,6 +460,78 @@ func (r *Rancher) SetASGClusterSize(perZoneCount int64, timeout time.Duration) e
...
+			clusterInterface := &rancherClient.Cluster{
+				Name: cluster.Name,
+				DefaultPodSecurityAdmissionConfigurationTemplateName: psaName,
+			}
+			_, err = r.client.Cluster.Update(&cluster, clusterInterface)
+			return err
+		}
+	}
+	return fmt.Errorf("cluster with cluster name %s is not present", clusterName)
+}
```
kphalgun-px, line 533: Is it necessary to call `r.client.Cluster.List(nil)` when we can directly use `r.client.Cluster.Update`, assuming the latter knows how to handle cases where the cluster does not exist?

How about?

```go
// UpdateClusterWidePSA adds the cluster wide PSA to the given cluster
func (r *Rancher) UpdateClusterWidePSA(clusterName string, psaName string) error {
	log.Infof("Updating cluster wide PSA %v for cluster %v", psaName, clusterName)
	clusterInterface := &rancherClient.Cluster{
		DefaultPodSecurityAdmissionConfigurationTemplateName: psaName,
	}
	_, err := r.client.Cluster.Update(clusterName, clusterInterface)
	if err != nil {
		log.Errorf("failed to update cluster %v with PSA %v: %v", clusterName, psaName, err) // the error message might not contain clusterName and psaName
		return err
	}
	return nil
}
```

kphalgun-px (approved):
LGTM

drivers/scheduler/rke/rke.go
kphalgun-px: My bad. I did not realize that we passing cluster object.

snigdha-px: Sure

ak-px (approved):
LGTM

mkoppal-px (approved):
Approving based on others review

PR #2554, opened 2024-05-30
Modify TC's for Azure backup location

sabrarhussaini:
As a part of this PR, the below modifications are made;
1. Helper method **IsAzureBlobEmpty**- is added to verify if the azure backup location is empty.
2. Two TC's were modified, to reduce the character length for backups and restores as they were exceeding 63 character limit.


tests/common.go
```diff
@@ -6649,6 +6652,42 @@ func IsGCPBucketEmpty(bucketName string) (bool, error) {
 	return false, nil
 }
 
+// IsAzureBlobEmpty returns true if bucket empty else false
+func IsAzureBlobEmpty(containerName string) (bool, error) {
+	_, _, _, _, accountName, accountKey := GetAzureCredsFromEnv()
+	azureRegion := os.Getenv("AZURE_ENDPOINT")
+	urlStr := fmt.Sprintf("https://%s.blob.core.windows.net/%s", accountName, containerName)
+	if azureRegion == "CHINA" {
```
sn-px, line 6660: can we add this into constant "CHINA"

sabrarhussaini: @sn-px Please advise, as it is just the name of the ENV variable which is passed from Jenkins file.

ak-px (approved):
LGTM

tests/common.go
sabrarhussaini: @sn-px Addressed.

kshithijiyer-px (approved):

sn-px (approved):
LGTM

mkoppal-px (approved):
LGTM

PR #2566, opened 2024-06-03
Adding function to modify maintenance job frequency

mkoppal-px:
Adding function to modify maintenance job frequency. This is required for Partial Backup Success automation.

Closes #PB-7089

I have added a dummy test to verify the function. Will remove it before merging.
[Jenkins BYOC Run](https://jenkins.pwx.dev.purestorage.com/job/portworx-backup/job/system-tests/job/byoc/job/px-backup-on-demand-system-test-byoc/5256/console)

### DIFF
--- a/tests/backup_helper.go
+++ b/tests/backup_helper.go
@@ -8752,3 +8752,59 @@ func ValidatePVCCleanup(pvcBefore, pvcAfter []string) error {
 
 	return nil
 }
+
+type MaintenanceJobType string
+
+const (
+	FullMaintenanceJob  MaintenanceJobType = "full-maintenance-repo"
+	QuickMaintenanceJob MaintenanceJobType = "quick-maintenance-repo"
+)
+
+// ModifyMaintenanceJobFrequency modifies the frequency of the cron job
+func ModifyMaintenanceJobFrequency(backupLocationName string, frequencyInMinutes int, maintenanceJobType MaintenanceJobType) error {
+	k8sBatch := batch.Instance()
+	cronJob, err := GetCronJobByBackupLocation(backupLocationName, maintenanceJobType)
+	if err != nil {
+		return err
+	}
+	cronJob.Spec.Schedule = fmt.Sprintf("*/%d * * * *", frequencyInMinutes)
+	_, err = k8sBatch.UpdateCronJob(cronJob)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetCronJobByBackupLocation takes the backup location name and the job type and returns the cron job
+func GetCronJobByBackupLocation(backupLocationName string, maintenanceJobType MaintenanceJobType) (*batchv1.CronJob, error) {
+	k8sBatch := batch.Instance()
+	ctx, err := backup.GetAdminCtxFromSecret()
+	if err != nil {
+		return nil, err
+	}
+	pxBackupNamespace, err := backup.GetPxBackupNamespace()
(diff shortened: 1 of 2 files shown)


tests/backup_helper.go
```diff
@@ -8752,3 +8752,59 @@ func ValidatePVCCleanup(pvcBefore, pvcAfter []string) error {
...
+	cronJob.Spec.Schedule = fmt.Sprintf("*/%d * * * *", frequencyInMinutes)
+	_, err = k8sBatch.UpdateCronJob(cronJob)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetCronJobByBackupLocation takes the backup location name and the job type and returns the cron job
+func GetCronJobByBackupLocation(backupLocationName string, maintenanceJobType MaintenanceJobType) (*batchv1.CronJob, error) {
```
kphalgun-px, line 8779: Can we take ctx as an argument?

kphalgun-px (approved):
LGTM

lsrinivas-pure (approved):
LGTM

tests/backup_helper.go
```diff
@@ -8752,3 +8752,59 @@ func ValidatePVCCleanup(pvcBefore, pvcAfter []string) error {
...
+
+type MaintenanceJobType string
+
+const (
+	FullMaintenanceJob  MaintenanceJobType = "full-maintenance-repo"
+	QuickMaintenanceJob MaintenanceJobType = "quick-maintenance-repo"
+)
+
+// ModifyMaintenanceJobFrequency modifies the frequency of the cron job
+func ModifyMaintenanceJobFrequency(backupLocationName string, frequencyInMinutes int, maintenanceJobType MaintenanceJobType) error {
```
ak-px, line 8764: How about having complete cron in parameter , considering the hourly schedules.

tests/backup_helper.go
```diff
@@ -8752,3 +8752,59 @@ func ValidatePVCCleanup(pvcBefore, pvcAfter []string) error {
...
+	k8sBatch := batch.Instance()
+	cronJob, err := GetCronJobByBackupLocation(backupLocationName, maintenanceJobType)
+	if err != nil {
+		return err
+	}
+	cronJob.Spec.Schedule = fmt.Sprintf("*/%d * * * *", frequencyInMinutes)
```
kshithijiyer-px, line 8770: Should we support all the cron params as it will be helpful later in case of longevity and scale test frameworks?

tests/backup_helper.go
mkoppal-px: This is not user specific. So ctx does not make sense here. `ctx` is need to fetch all the available backup locations which can will be given by admin context.

tests/backup_helper.go
mkoppal-px: If you want hourly schedule we need to give it in minutes. I did this to reduce the input parameters. Ideally we would want this to be in minutes.

tests/backup_helper.go
mkoppal-px: In the next PR I will be providing a function specific to update the image and memory of the cron job.

tests/backup_helper.go
kphalgun-px: Makes sense. Thank you.

kshithijiyer-px (approved):
LGTM

ak-px (approved):
LGTM

PR #2567, opened 2024-06-03
PTX-24441 - Skip PDB allowed disruptions checking for anthos upgrade

svijaykumar-px:
**What this PR does / why we need it**: Anthos does not follow PDB value set by portworx and sometimes brings down more nodes. Till the issue is solved, skipping validation of nodes down for anthos upgrades.


jainsmit (approved):
LGTM !! Please test it once

nikolaypopov (approved):

PR #2575, opened 2024-06-06
Adds ability to set FADA pod name in torpedo flag

Pure-AdamuKaapan:
To test FADA multitenancy, it would be helpful to be able to add pod names to existing specs, instead of making whole separate apps just for those. This adds a variable to add the pod name into storage classes as we create them, in certain cases.


deployments/deploy-ssh.sh
```diff
@@ -547,6 +551,7 @@ spec:
             "--pure-volumes=$IS_PURE_VOLUMES",
             "--pure-fa-snapshot-restore-to-many-test=$PURE_FA_CLONE_MANY_TEST",
             "--pure-san-type=$PURE_SAN_TYPE",
+            "--pure-fada-pod=$PURE_FADA_POD",
```
lsrinivas-pure, line 554: @Pure-AdamuKaapan one quick question, will there be a scenario where we need to test multiple apps with different pod names?

vprabhakar-px (approved):
LGTM

vumesh-px (approved):
This is okay for having a single pod for all apps in the entire run. 
Had an offline conversation with @lsrinivas-pure  and @sayalasomayajula-px, we will also need a way for CustomAppConfig (which could be outside this PR)

deployments/deploy-ssh.sh
Pure-AdamuKaapan: Yes but we aren't tackling that yet, we can figure out the specifics later (as noted in the comment from Vishal)

pp511 (approved):

PR #2582, opened 2024-06-07
PTX-24281:moving repls on the node before pool deletion

lsrinivas-pure:
moving repls on the node before pool deletion

Closes #PTX-24281

### DIFF
--- a/tests/common.go
+++ b/tests/common.go
@@ -9163,6 +9163,26 @@ func ExitPoolMaintenance(stNode node.Node) error {
 
 // DeleteGivenPoolInNode deletes pool with given ID in the given node
 func DeleteGivenPoolInNode(stNode node.Node, poolIDToDelete string, retry bool) (err error) {
+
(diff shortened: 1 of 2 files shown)


tests/common.go
```diff
@@ -9156,6 +9156,26 @@ func ExitPoolMaintenance(stNode node.Node) error {
 
 // DeleteGivenPoolInNode deletes pool with given ID in the given node
 func DeleteGivenPoolInNode(stNode node.Node, poolIDToDelete string, retry bool) (err error) {
+
+	//Moving repls on the node before deletion
```
mkoppal-px, line 9160: nit: Leave a space after `//` else GoLand will create a diff sometimes

mkoppal-px (approved):
LGTM except for a nit

tests/common.go
lsrinivas-pure: fixed

PR #2589, opened 2024-06-10
Automating KDMP backup failure

mkoppal-px:
We need to automate KDMp backup failure to test partial backup success feature.

Closes #PB-7087

[Jenkins BYOC with the sample test case](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Mithun/job/Quick%20BYOC/107/)


tests/backup_helper.go
```diff
@@ -8948,3 +8950,28 @@ func UpdateMaintenanceCronJob(backupLocation string, maintenanceJobType Maintena
 	_, err = k8sBatch.UpdateCronJob(cronJob)
 	return err
 }
+
+// DeleteAllDataExportCRs deletes the data export CR as soon as it is found in the given namespace
+func DeleteAllDataExportCRs(namespace string) error {
```
ak-px, line 8955: I think its better have to filter options as well, in case if single namespace of multiple volumes and with same volume type and if we delete at namespace level the backup will be completely failed.

mkoppal-px: @ak-px I thought about this, I will create another function called `DeleteDataExportCRs` which will take a map[namespace]pvcName
Can I do this in next PR or shall I add this function in the same PR?

ak-px: Currently this method won't block us as we are using the different vol type app. so it's ok to do in separate PR.

ak-px (approved):
LGTM

lsrinivas-pure (approved):
LGTM

PR #2593, opened 2024-06-11
Disables FADA cloudsnap validation temporarily

Pure-AdamuKaapan:
See PWX-37628: the pxctl cloudsnap method currently succeeds for FADA volumes when it should fail. To unblock testing, disabling this specific check for now.

### DIFF
--- a/tests/basic/pure_test.go
+++ b/tests/basic/pure_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+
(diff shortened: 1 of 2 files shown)


lsrinivas-pure (approved):
LGTM
can you raise a PTX to enable it again once issue is fixed. Thanks

Pure-AdamuKaapan:
@lsrinivas-pure dev will reenable it once we have investigated, as part of PWX-37628.

PR #2608, opened 2024-06-18
Added a function to fail Px volume backup by stopping cloudsnap

mkoppal-px:

- [x] Added a function `StopCloudsnapBackup` to fail Px volumes and stop the snapshot
- [x] Added a new spec for mysql which also added some random data to the volume. Needed this to stop the snapshot, otherwise the snapshot gets completed quickly and we may fail to stop it
- [x] Added a dummy test case to test the `StopCloudsnapBackup` function. This will be removed later.
- [x] Tested the function by running it in goroutine to fail multiple volumes while the backup is going on

Closes #PB-7224

[BYOC jenkins run](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Mithun/job/Quick%20BYOC/130/)
This run has been aborted because the backup ends up with partial success as expected, but we do not have validation for the partial success state. So had to abort.


dbhatnagar-px (approved):
Lgtm

lsrinivas-pure (approved):
LGTM

ak-px (approved):
LGTM

PR #2611, opened 2024-06-18
Fixing OKE platform upgrade

lsrinivas-pure:
Fixing OKE platform upgrade test
Closes #PTX-24419

https://jenkins.pwx.dev.purestorage.com/job/Torpedo/view/Operator%20Release%20Cloud%20Dashboard/job/tp-nextpx-oke-upgrade-hops-op/42/console

### DIFF
--- a/drivers/scheduler/oke/oke.go
+++ b/drivers/scheduler/oke/oke.go
@@ -17,13 +17,12 @@ import (
 
 	"github.com/libopenstorage/cloudops"
 	oracleOps "github.com/libopenstorage/cloudops/oracle"
-	"github.com/oracle/oci-go-sdk/v65/core"
 	"github.com/portworx/torpedo/drivers/node"
 	kube "github.com/portworx/torpedo/drivers/scheduler/k8s"
 )
 
 const (
-	// Schedular is the name of the oke driver
+	// SchedName is the name of the oke driver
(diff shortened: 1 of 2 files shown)


drivers/scheduler/oke/oke.go
```diff
@@ -577,6 +764,30 @@ func init() {
...
+
+	stdout, stderr, err := osutils.ExecShell(cmd)
+	if err != nil {
+		return "", fmt.Errorf("failed to get instance list. stderr: %v, err: %v,stdout: %v", stderr, err, stdout)
+	}
+
+	instances := &Instance{}
+	err = json.Unmarshal([]byte(stdout), instances)
+	if err != nil {
+		return "", err
```
nikolaypopov, line 778: Can we add some more verbose error here like you did above, I faced many issues in spawn due to OKE errors are not very straight forwardc (did a huge overhaul to add verbosity, which helped me track down some issues) and because we had very little prints in the functions, it was impossible to chase down where the error was actually coming from

dbhatnagar-px (approved):
Lgtm

drivers/scheduler/oke/oke.go
lsrinivas-pure: fixed it

PR #2617, opened 2024-06-21
Fetching RKE endpoint from the env

snigdha-px:

Closes #

### DIFF
--- a/drivers/scheduler/rke/rke.go
+++ b/drivers/scheduler/rke/rke.go
@@ -80,7 +80,7 @@ func (r *Rancher) GetRancherClusterParametersValue() (*RancherClusterParameters,
 	// TODO Rancher URL for cloud cluster will not be fetched from master node IP
 	masterNodeName := node.GetMasterNodes()[0].Name
 	log.Infof("The master node here is %v", masterNodeName)
-	endpoint := "https://" + masterNodeName + "/v3"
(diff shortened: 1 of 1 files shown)


drivers/scheduler/rke/rke.go
```diff
@@ -97,17 +97,20 @@ func (r *Rancher) UpdateRancherClient(clusterName string) error {
...
+			return fmt.Errorf("env variable DESTINATION_RKE_TOKEN or DESTINATION_RKE_URL should not be empty")
 		}
 	} else if clusterName == "source-config" {
 		rkeToken = os.Getenv("SOURCE_RKE_TOKEN")
-		if rkeToken == "" {
-			return fmt.Errorf("env variable SOURCE_RKE_TOKEN should not be empty")
+		endpoint = os.Getenv("SOURCE_RKE_URL")
+		log.Infof("Source endpoint: %v", endpoint)
+		if rkeToken == "" || endpoint == "" {
+			return fmt.Errorf("env variable SOURCE_RKE_TOKEN or SOURCE_RKE_URL should not be empty")
```
vpinisetti-px, line 113: @snigdha-px Do we really need the check here. I believe we can create only one Rancher cluster and add both source & destination clusters to same Rancher cluster. In that case we no need to export two env variables and no check required.
@sudas-px Need your inputs here.

vpinisetti-px (changes requested):

PR #2620, opened 2024-06-21
Moving the IBMCLI to nexus

aghodke1312:

Closes #


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM golang:1.21.6-alpine AS build
 LABEL maintainer="harsh@portworx.com"
 ARG MAKE_TARGET
+ARG NEXUS_URL
```
mborodin-px, line 4: we don't need to parametrize this anymore, just define it here

Santhoshairody (approved):

mborodin-px (approved):

PR #2627, opened 2024-06-24
PWX-37657: Changed DeployAppsAndStopPortworx to be serial instead of parallel

Pure-AdamuKaapan:
The original intent of this test was to ensure that a drive set failover is successful - in the parallel state it's more testing failures during volume provision which is not the intent. Refactored to make it serial.

PWX-37657


pp511 (approved):
lgtm

PR #2647, opened 2024-07-02
PTX-24898 Restart the px while managment interface is down and enable interface during start of px

sayalasomayajula-px:
…nable it to check if px is up

This PR  Stops PX and then makes management interface down and starts the PX after that it makes mgmt interface Up and check if PX is up

Closes #
https://purestorage.atlassian.net/browse/PTX-24898

https://jenkins.pwx.dev.purestorage.com/job/Users/job/krishna/job/tp-pxe-byoc/3753/display/redirect


lsrinivas-pure (approved):
LGTM

PR #2661, opened 2024-07-08
Adding invalid provisioner for CreateInvalidVolumeSnapshotClass(

ak-px:
Adding invalid provisioner for to method CreateInvalidVolumeSnapshotClass.
In some providers like IKS, the CSI driver is not adhering to an invalid parameter in SnapshotClass and not causing the CSI snapshot to fail. Hence made the driver name itself as invalid.

Closes #


mkoppal-px (approved):
LGTM

PR #2662, opened 2024-07-08
Fix for SSIE tests

lsrinivas-pure:
- Fix for VolumeHAPoolOpsNoKVDBleaderDown , failing due to invalid app contexts.
- Fix for longevity kvdb failover test due to stale node registry values
- Fix for volume size using pvc rather than spec

Closes #PTX-25004,PTX-24997,PTX-24959


dbhatnagar-px (approved):
Lgtm

PR #2663, opened 2024-07-08
PTX-25021 Add API acknowledgement for OCP 4.16

nikolaypopov:
PTX-25021 Add API acknowledgement for OCP 4.16


lsrinivas-pure (approved):
LGTM

PR #2665, opened 2024-07-10
PWX-36514: Adding validation for node PDB during k8s upgrade

svijaykumar-px:
**What this PR does / why we need it**: From operator 24.2.0, there will no longer be cluster wide PDB, instead there will only be per node PDB. Thus adding validation for node PDBs during k8s cluster upgrades. The code will validate that only maximum of numNodes-quorum nodes have PDB minAvailable 0 at a time and are down. This count is the total number of nodes that can be upgraded at a time. 
The quorum value by default is n/2+1 but can be changed based on valid `portworx.io/storage-pdb-min-available` value.

Closes # PWX-36514, 

**Special notes for your reviewer**: validation for maintaining volume quorum will be done as a part of another PTX

### DIFF
--- a/tests/basic/upgrade_cluster_test.go
+++ b/tests/basic/upgrade_cluster_test.go
@@ -114,7 +114,7 @@ var _ = Describe("{UpgradeCluster}", func() {
 
 				err = Inst().S.UpgradeScheduler(version)
 				if err != nil {
-					err = Inst().S.RefreshNodeRegistry()
+					err := Inst().S.RefreshNodeRegistry()
 					log.FailOnError(err, "Refresh Node Registry failed")
 					err = Inst().V.RefreshDriverEndpoints()
 					log.FailOnError(err, "Refresh Driver Endpoints failed")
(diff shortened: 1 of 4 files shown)


tests/basic/upgrade_cluster_test.go
```diff
@@ -74,24 +74,23 @@ var _ = Describe("{UpgradeCluster}", func() {
...
-		storageNodes := node.GetStorageNodes()
-		if len(storageNodes) == 0 {
-			log.FailOnError(fmt.Errorf("No storage nodes found"), "error getting px storage nodes")
-		}
-		pxver, err := Inst().V.GetDriverVersionOnNode(storageNodes[0])
-		if err != nil {
-			log.Errorf("error getting px version on node [%s]: [%v]", storageNodes[0].Name, err)
+		pxVersionString, err := Inst().V.GetDriverVersion()
+		var pxVersion *version.Version
+		if err == nil {
```
lsrinivas-pure, line 79: fail if if err else pxVersion will be nil and we will nil pointer in line 88

lsrinivas-pure (approved):
LGTM

PR #2673, opened 2024-07-11
Adding APIs for defrag schedule

jainsmit:
Adding APIs for defrag schedules. These APIs are required to implement defrag schedule longevity triggers

Closes # PTX-24479

These are generic defrag schedule apis

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -216,6 +216,8 @@ type portworx struct {
 	licenseFeatureManager pxapi.PortworxLicensedFeatureClient
 	autoFsTrimManager     api.OpenStorageFilesystemTrimClient
 	portworxServiceClient pxapi.PortworxServiceClient
+	defragManager         api.OpenStorageFilesystemDefragClient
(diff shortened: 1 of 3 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -6316,4 +6317,70 @@ func (d *portworx) UpdateSkinnySnapReplNum(repl string) error {
 		break
 	}
 	return nil
-}
\ No newline at end of file
+}
+
+// CreateDefragSchedule create defrag schedule for provided defrag job
+func (d *portworx) CreateDefragSchedule(startTime string, defragJob *api.DefragJob) error {
```
Rohit-PX, line 6323: We can return the `DefragSchedule` object here, that way the function can be used in tests where we want to `edit` or `delete` the schedule if needed.

Rohit-PX (approved):
LGTM except for the comment.

drivers/volume/portworx/portworx.go
jainsmit: Yeah right, Incorporated it.

PR #2683, opened 2024-07-16
Adding token for pxctl commands if security is enabled

lsrinivas-pure:
- Fix for running pxctl commands with admin context when security is enabled
Closes #PTX-24835

Result: https://aetos.pwx.purestorage.com/resultSet/testSetID/699250


dbhatnagar-px (approved):
Lgtm

PR #2691, opened 2024-07-18
Fix ScaleMongoDBWhileBackupAndRestore failure on OCP with OCS

mkoppal-px:
The test case `ScaleMongoDBWhileBackupAndRestore` was failing on OCP with OCS platform while restoring a KDMP backup. This was because we were trying to restore multiple backups in the same destination namespace at the same time in parallel. This was causing the KDMP restore to stay in progress for ever. And this is an invalid use case for any platform. So now, I will be restoring the 5 backups taken into 5 different destination namespaces.

Closes #PB-7596

[Jenkins Run](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Mithun/job/BYOC-%20For%20PSO/98/)


lsrinivas-pure (approved):
LGTM

kshithijiyer-px (approved):
LGTM

PR #2692, opened 2024-07-18
Upgrade cluster test is panicking when printing pxctl status 

jainsmit:
…pgrade test failure

Upgrade cluster test is panicking when printing pxctl status  during upgrade test failure.

Node were deleted and re-created during scheduler upgrade, test update the node registry but node also has to be updated weather nodes are storage nodes or not.

It was failing during `node.GetStorageNodes()[0]` call in common.go (line number 772)

Closes # PTX-21736

Trivial fix
https://jenkins.pwx.dev.purestorage.com/job/Torpedo/job/tp-anthos-upgrade/450/console

### DIFF
--- a/tests/basic/upgrade_cluster_test.go
+++ b/tests/basic/upgrade_cluster_test.go
@@ -2,6 +2,10 @@ package tests
 
 import (
 	"fmt"
+	"net/url"
+	"strings"
+	"time"
+
 	oputil "github.com/libopenstorage/operator/pkg/util/test"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -17,9 +21,6 @@ import (
 	"github.com/portworx/torpedo/pkg/log"
 	. "github.com/portworx/torpedo/tests"
 	corev1 "k8s.io/api/core/v1"
-	"net/url"
-	"strings"
-	"time"
 )
 
 var _ = Describe("{UpgradeCluster}", func() {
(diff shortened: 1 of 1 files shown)


Rohit-PX (approved):

tests/basic/upgrade_cluster_test.go
```diff
@@ -90,6 +91,7 @@ var _ = Describe("{UpgradeCluster}", func() {
 
 				err = Inst().S.UpgradeScheduler(version)
 				if err != nil {
+					err = Inst().S.RefreshNodeRegistry()
```
nikolaypopov, line 94: Do you need to check this error? If not we can just do `Inst().S.RefreshNodeRegistry()`

jainsmit: Actually Anthos scheduler is invoking "RefreshNodeRegistry()"  after the upgrade completed. This method only update new nodes IPs and names but doesn't update the flag `IsStorageDriverInstalled` because of this when `GetStorageNodes()` method invoked after upgrade, it return empty list (As part of `PrintCommandOutput()` method call `node.GetStorageNodes()[0]`)

jainsmit: Sorry I need to use this method `Inst().V.RefreshDriverEndpoints()`

tests/basic/upgrade_cluster_test.go
```diff
@@ -90,6 +91,7 @@ var _ = Describe("{UpgradeCluster}", func() {
 
 				err = Inst().S.UpgradeScheduler(version)
 				if err != nil {
+					err = Inst().V.RefreshDriverEndpoints()
```
lsrinivas-pure, line 94: can we add refershnode registry also?
I see only anthos is doing it but if we add it here it will common for all the platforms

jainsmit: sure, done !!

PR #2693, opened 2024-07-19
PTX-25257 Fix deleting admin security context even if cmd fails

nikolaypopov:
Fix deleting admin security context even if cmd fails


jainsmit (approved):
LGTM !!

aghodke1312 (approved):

PR #2695, opened 2024-07-21
ignoring dmthin configuration

vprabhakar-px:

Closes #

### DIFF
--- a/tests/basic/storage_pool_test.go
+++ b/tests/basic/storage_pool_test.go
@@ -2873,6 +2873,12 @@ var _ = Describe("{ResizeWithJrnlAndMeta}", func() {
 
 	It(stepLog, func() {
 		log.InfoD(stepLog)
+		isDmthin, err := IsDMthin()
+		log.FailOnError(err, "Error while checking cluster type")
+		if isDmthin {
+			Skip("Drive add Journal Device is not supported for DMThin")
+		}
+
 		journalStatus, err := IsJournalEnabled()
 		log.FailOnError(err, "err getting journal status")
 		dash.VerifyFatal(journalStatus, true, "verify journal device is enabled")


tests/basic/storage_pool_test.go
```diff
@@ -2873,6 +2873,12 @@ var _ = Describe("{ResizeWithJrnlAndMeta}", func() {
 
 	It(stepLog, func() {
 		log.InfoD(stepLog)
+		isDmthin, err := IsDMthin()
+		log.FailOnError(err, "Error while checking cluster type")
+		if isDmthin {
+			Skip("Drive add Journal Device is not supported for DMThin")
```
lsrinivas-pure, line 2879: is this message correct?
Also this test is not applocable for dmthin as dmthin will have separate metadata device
this test is for btrfs where pool has hasMetadata:yes

vprabhakar-px: currently Yes, but there is an effort in progress to get  journal devices for DMTHin , currently the code base is disabled but eventually it will be enabled .

lsrinivas-pure: what about metadata device?

vprabhakar-px: metadata device should be present for pool to exists in DMThin , eventually both metadata and journal might be present .

lsrinivas-pure: yes for dmthin we need dedicated metadata device, this test is specific for pool which has both jrnl and metadata device part of it. Let me know if my understanding is wrong

vprabhakar-px: yes, but we are not  specifically checking if there is metadata or journaldevice in the script. we just check if in case of journal present ,we check for volume size 

```		expectedSizeWithJournal := expectedSize
		if isJournalEnabled {
			expectedSizeWithJournal = expectedSizeWithJournal - 3
		}
		if newPoolSize >= expectedSizeWithJournal {
			// storage pool resize has been completed
			return nil, false, nil
		}
```

lsrinivas-pure: i see this test case has to be fixed .. to use pool with metadata.. i see that is not happening now. Approving for now. can you raise a PTX for this fix

lsrinivas-pure (approved):

PR #2705, opened 2024-07-23
Added stork volume snapshot test triggers

gejain:
Added stork volume snapshot test triggers

Closes #https://purestorage.atlassian.net/browse/PTX-24966

Refer trigger run result - https://aetos.pwx.purestorage.com/resultSet/testSetID/705387/testCaseID/4563771

### DIFF
--- a/pkg/asyncdr/asyncdr.go
+++ b/pkg/asyncdr/asyncdr.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	crdv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/k8sutils"
 	"github.com/portworx/sched-ops/k8s/apiextensions"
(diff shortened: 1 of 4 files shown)


tests/testTriggers.go
```diff
@@ -10283,6 +10287,132 @@ func TriggerAsyncDRMigrationSchedule(contexts *[]*scheduler.Context, recordChan
...
+		if err != nil {
+			UpdateOutcome(event, fmt.Errorf("Failed to get pvc list for namespace"))
+			return
+		}
+		err = asyncdr.ValidateSnapshotScheduleCount(pvcNames, ns, schdPol.Name, snapshotType, snapInterval, retain)
+		if err != nil {
+			UpdateOutcome(event, fmt.Errorf("Failed to validate snapshot schedule count"))
+			return
+		}
+	}
```
kphalgun-px, line 10350: It would be helpful while debugging if we print params in the log messages.

kphalgun-px (approved):
LGTM

tests/testTriggers.go
gejain: Added log for PVC and NS. Thanks!

pkg/asyncdr/asyncdr.go
```diff
@@ -281,6 +282,117 @@ func CreateMigrationSchedule(
...
+					PersistentVolumeClaimName: pvcName},
+			},
+			SchedulePolicyName: policyName,
+		},
+	}
+	snapSchedule, err := storkops.Instance().CreateSnapshotSchedule(snapSched)
+	return snapSchedule, err
+}
+
+func CreateSchedulePolicyWithRetain(policyName string, interval int, retain storkapi.Retain) (pol *storkapi.SchedulePolicy, err error) {
```
Rohit-PX, line 312: TODO later: Just have a  generic function called `CreateSchedulePolicy` and based on if `retain != nil` , use it to create SchedulePolicy with or without `retain`.
In my opinion we don't need a function just to create a snapshot will retain policy.

tests/testTriggers.go
```diff
@@ -10283,6 +10287,134 @@ func TriggerAsyncDRMigrationSchedule(contexts *[]*scheduler.Context, recordChan
...
+	defer func() {
+		event.End = time.Now().Format(time.RFC1123)
+		*recordChan <- event
+	}()
+
+	setMetrics(*event)
+
+	var (
+		taskNamePrefix     = "snaptest"
```
Rohit-PX, line 10375:
```suggestion
		taskNamePrefix     = "stork-snaptest"
```

Rohit-PX (approved):
Except these two comments, looks good to me.

pkg/asyncdr/asyncdr.go
gejain: Yes, as there are lot of tests using it, I will refactor this later.

tests/testTriggers.go
gejain: Done

PR #2709, opened 2024-07-29
PTX-25073:  Validate volume quorum for cluster and PX upgrade

nikita-bhatia:
Added validation for checking that volume quorum is maintained during portworx and kubernetes upgrade.

Closes # PTX-25073 PTX-25027

Job Runs Results :

PX upgrade
1. Vanilla k8s with 10 storage nodes, 4 nodes upgraded at time
https://jenkins.pwx.dev.purestorage.com/job/Users/job/krishna/job/tp-pxe-byoc/4188/console
https://jenkins.pwx.dev.purestorage.com/job/Users/job/krishna/job/tp-pxe-byoc/4193/console

Kuberenetes Upgrade
1. OCP cluster with 10 storage nodes


tests/basic/upgrade_cluster_test.go
```diff
@@ -89,6 +89,18 @@ var _ = Describe("{UpgradeCluster}", func() {
 					log.Warnf("PDB validation skipped. Current Px-Operator version: [%s], minimum required: [%s]. Error: [%v].", opver, PDBValidationMinOpVersion, err)
 				}
 
+				// validate volume quorum during upgrade
+				if opver.GreaterThanOrEqual(ParallelUpgradeVersion) {
+					var vQourumError error
```
kphalgun-px, line 94: Do we report is vQourumError is not nil?

tests/basic/upgrade_cluster_test.go
```diff
@@ -89,6 +89,18 @@ var _ = Describe("{UpgradeCluster}", func() {
 					log.Warnf("PDB validation skipped. Current Px-Operator version: [%s], minimum required: [%s]. Error: [%v].", opver, PDBValidationMinOpVersion, err)
 				}
 
+				// validate volume quorum during upgrade
```
jainsmit, line 92: Please make sure this check works for all different distributions

tests/basic/upgrade_cluster_test.go
nikita-bhatia: added the validation of error

tests/common.go
```diff
@@ -13948,6 +13949,76 @@ func GetVolumeNamefromPVC(namespace string) ([]string, error) {
...
+	for _, volID := range volIDs {
+		apiVol, err := Inst().V.InspectVolume(volID)
+		if err != nil {
+			err = fmt.Errorf("error inspecting volume [%s], Err: %v", volID, err)
+			processError(err, errChan...)
+			return
+		}
+		// check if volume is up
+		log.Infof("Volume [%s] status: %v", volID, apiVol.Status)
+		if apiVol.Status != opsapi.VolumeStatus_VOLUME_STATUS_UP {
```
piyush-nimbalkar, line 13996: @nikita-bhatia We don't expose if a volume is in resync or not from our SDK. I believe the volume will show up as STATUS_UP when resync is running. As we don't expose this information from the SDK, we will have to call `pxctl volume inspect` and look at that data to determine if a volume is in quorum and in sync.

tests/basic/upgrade_test.go
```diff
@@ -157,6 +157,20 @@ var _ = Describe("{UpgradeVolumeDriver}", func() {
...
 
+			var vQourumError error
+			opver, err := optest.GetPxOperatorVersion()
+			if err == nil && opver.GreaterThanOrEqual(ParallelUpgradeVersion) {
+				log.Info("Starting volume quorum validation for Portworx upgrade .......")
+
+				stopVolumeQuorumValidationSignal := make(chan struct{})
+				go DoVolumeQuorumValidation(stopVolumeQuorumValidationSignal, &vQourumError)
+				defer func() {
+					close(stopVolumeQuorumValidationSignal)
```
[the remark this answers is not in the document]
nikita-bhatia: We can always validate, but the number of nodes that will be upgraded will not be always same what customer has configured. Since this completely depends on volume placement strategy, like where volume replicas are residing, it is difficult to validate that always specific number of nodes are being upgraded or not.

tests/common.go
```diff
@@ -13948,6 +13949,76 @@ func GetVolumeNamefromPVC(namespace string) ([]string, error) {
...
+				err := fmt.Errorf("volume [%s] does not have any replicas", volID)
+				processError(err, errChan...)
+				return
+			}
+
+			replicaNodes := replicas[0].Nodes
+
+			// if it is repl-2 volume and replicas are on same node, do not fail the test
+			// if it is repl-3 volume and replicas or either on same or 2 different nodes, do not fail the test
+			if len(replicaNodes) == 1 || (len(replicaNodes) == 2 && len(apiVol.ReplicaSets) == 3) {
```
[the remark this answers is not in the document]
nikita-bhatia: since this feature doesn't guarantee up time for repl-1, hence we will skip the failure for same

tests/common.go
```diff
@@ -13948,6 +13949,76 @@ func GetVolumeNamefromPVC(namespace string) ([]string, error) {
...
+			if len(replicas) == 0 {
+				err := fmt.Errorf("volume [%s] does not have any replicas", volID)
+				processError(err, errChan...)
+				return
+			}
+
+			replicaNodes := replicas[0].Nodes
+
+			// if it is repl-2 volume and replicas are on same node, do not fail the test
+			// if it is repl-3 volume and replicas or either on same or 2 different nodes, do not fail the test
```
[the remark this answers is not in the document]
nikita-bhatia: If there is a scenario 2 , then it is a bug on how the volume replicas are present on same node. But IMO it is not a bug that should fail this scenario, because upgrades will happen with volume uptime only if replicas are available on different nodes. Please suggest.

tests/basic/upgrade_cluster_test.go
nikita-bhatia: This is being validated on vanilla k8s and OCP clusters

tests/common.go
nikita-bhatia: In a repl-3 volume, it is seen that even when one of 3 node is being updated, volume inspect shows replication status in "Degraded" state, is that something we want to consider here?
image

nikita-bhatia: As discussed with @piyush-nimbalkar , have added extra validation to validate Runtime sttaus of the volumes along with volume status. This is going to validate that nodes which are down for upgrades are in "clean" RunTime state only.

tests/basic/upgrade_cluster_test.go
```diff
@@ -74,6 +75,10 @@ var _ = Describe("{UpgradeCluster}", func() {
 
 		printDisks(preUpgradeNodeDisksMap)
 
+		storageNodes := node.GetStorageNodes()
+		pxver, _ := Inst().V.GetDriverVersionOnNode(storageNodes[0])
```
lsrinivas-pure, line 79: add a check to validate len(storageNodes) > 0
 also handle the err in GetDriverVersionOnNode

tests/basic/upgrade_test.go
```diff
@@ -157,6 +158,22 @@ var _ = Describe("{UpgradeVolumeDriver}", func() {
 				close(stopSignal)
 			}()
 
+			var vQuorumError error
+			opver, _ := optest.GetPxOperatorVersion()
```
lsrinivas-pure, line 162: same here.. handle the errors

tests/common.go
```diff
@@ -14075,6 +14079,113 @@ func GetVolumeNamefromPVC(namespace string) ([]string, error) {
...
+		if len(apiVol.ReplicaSets) == 0 {
+			err := fmt.Errorf("volume [%s] does not have any replicas", volID)
+			processError(err, errChan...)
+			return
+		}
+
+		replicaNodes := apiVol.ReplicaSets[0].Nodes
+
+		if apiVol.Status != opsapi.VolumeStatus_VOLUME_STATUS_UP {
+			// if it is repl-2 volume and replicas are on same node, do not fail the test
```
lsrinivas-pure, line 14137: nit: fix the comment; skip repl-1 volume validation

tests/common.go
```diff
@@ -14075,6 +14079,113 @@ func GetVolumeNamefromPVC(namespace string) ([]string, error) {
...
+			return
+		}
+
+		runTimeState := apiVol.RuntimeState[0].RuntimeState[VolumeRuntimeStateKey]
+		// check if volume is up and runtime status is clean
+		log.Infof("Volume [%s] status : %v, runtime state: %v", volID, apiVol.Status, runTimeState)
+
+		// if volume is not in clean state, check if all the nodes of it's repilcas are in storage up state
+		if runTimeState != VolumeRuntimeStatusClean {
+			err = fmt.Errorf("volume [%s] runtime state is not clean", volID)
```
lsrinivas-pure, line 14160: can you just log the volume state here as you are validating the node state below

tests/basic/upgrade_cluster_test.go
nikita-bhatia: done

tests/common.go
nikita-bhatia: done

tests/common.go
```diff
@@ -14075,6 +14079,113 @@ func GetVolumeNamefromPVC(namespace string) ([]string, error) {
...
+				nodeStatus, err := Inst().V.GetNodeStatus(nodeInfo)
+				if err != nil {
+					err = fmt.Errorf("error getting node status for node [%s], Err: %v", replicaNodes[i], err)
+					processError(err, errChan...)
+					return
+				}
+
+				// if node is in storage down state and runtime state is not clean, fail the test
+				log.Infof("Node [%s] status: %v", replicaNodes[i], nodeStatus)
+				if reflect.DeepEqual(nodeStatus, opsapi.Status_STATUS_STORAGE_DOWN) {
```
lsrinivas-pure, line 14179:
```suggestion
				dash.VerifyFatal(nodeStatus, opsapi.Status_STATUS_STORAGE_DOWN, fmt.Sprintf("Validating the node [%s ] state where volume [%s] has %s", replicaNodes[i], apiVol.Id, runTimeState))
```

nikita-bhatia:
@lsrinivas-pure updated the logs as suggested. Please review.

tests/basic/upgrade_cluster_test.go
```diff
@@ -89,6 +100,19 @@ var _ = Describe("{UpgradeCluster}", func() {
 					log.Warnf("PDB validation skipped. Current Px-Operator version: [%s], minimum required: [%s]. Error: [%v].", opver, PDBValidationMinOpVersion, err)
 				}
 
+				var vQuorumError error
+				// validate volume quorum during upgrade
+				if opver.GreaterThanOrEqual(ParallelUpgradeOperatorVersion) && pxVersion.GreaterThanOrEqual(ParallelUpgradePXVersion) {
+					log.Info("Starting volume quorum validation for Portworx upgrade .......")
+					stopVolumeQuorumValidationSignal := make(chan struct{})
+					go DoVolumeQuorumValidation(stopVolumeQuorumValidationSignal, &vQuorumError)
```
lsrinivas-pure, line 108: can you a validation for vQuorumError, fail the test if it has any errors.

tests/common.go
```diff
@@ -14075,6 +14079,101 @@ func GetVolumeNamefromPVC(namespace string) ([]string, error) {
...
+			return
+		}
+
+		runTimeState := apiVol.RuntimeState[0].RuntimeState[VolumeRuntimeStateKey]
+		// check if volume is up and runtime status is clean
+		log.Infof("Volume [%s] status : %v, runtime state: %v", volID, apiVol.Status, runTimeState)
+
+		// if volume is not in clean state, check if all the nodes of it's repilcas are in storage up state
+		if runTimeState != VolumeRuntimeStatusClean {
+			dash.Infof("volume [%s] runtime state is %v which is not clean, validating the node state...", volID, runTimeState)
```
lsrinivas-pure, line 14160: use InfoD here instead of Infof as it will post the log to aetos

tests/basic/upgrade_cluster_test.go
```diff
@@ -89,6 +100,19 @@ var _ = Describe("{UpgradeCluster}", func() {
 					log.Warnf("PDB validation skipped. Current Px-Operator version: [%s], minimum required: [%s]. Error: [%v].", opver, PDBValidationMinOpVersion, err)
 				}
 
+				var vQuorumError error
+				// validate volume quorum during upgrade
+				if opver.GreaterThanOrEqual(ParallelUpgradeOperatorVersion) && pxVersion.GreaterThanOrEqual(ParallelUpgradePXVersion) {
```
svijaykumar-px, line 105: `if err == nil && opver.GreaterThanOrEqual(PDBValidationMinOpVersion) {` This part needs to go into an "if-else" condition where we check for cluster wide pdb validations if operator version is lesser than 24.2.0 and 24.1.0 or greater. This is already a part of another PR. - https://github.com/portworx/torpedo/pull/2665

tests/basic/upgrade_test.go
```diff
@@ -157,6 +158,24 @@ var _ = Describe("{UpgradeVolumeDriver}", func() {
 				close(stopSignal)
 			}()
```
svijaykumar-px, line 160: We also need to add validation on how many nodes go down during px upgrade.

tests/basic/upgrade_cluster_test.go
nikita-bhatia: @lsrinivas-pure this is changed back to following code because our intention is not to fail the test when runTime status of healthy node is resync. We are trying to validate here that the node which is being upgraded has runtime state clean.

if reflect.DeepEqual(nodeStatus, opsapi.Status_STATUS_STORAGE_DOWN) {
					err = fmt.Errorf("node [%s] is in %v Runtime state ", replicaNodes[i], runTimeState)
					processError(err, errChan...)
					return
				}

lsrinivas-pure (approved):
LGTM

PR #2714, opened 2024-08-01
Fixing long running flaky kubevirt tests

mkoppal-px:
There were two kubevirt tests which were failing intermittently - `DefaultBackupRestoreWithKubevirtAndNonKubevirtNS` and `CustomBackupRestoreWithKubevirtAndNonKubevirtNS` - fixed these with the following:

- [x] Instead of using `CreateRestoreWithReplacePolicyWithValidation`, we will now be using `CreateRestoreWithReplacePolicy` and `ValidateCustomResourceRestores` because the validation in `CreateRestoreWithReplacePolicyWithValidation` was expecting all the resources in the restored namespace whereas we were doing custom backup of PVCs only
- [x] The `namespaceMapping` map and `restoreContextMap` map had to be cleared for every backup because every backup had a different namespace backed up (multi-app single ns and multi-app multi ns)
- [x] In `CreateRestoreWithReplacePolicy` we now do not determine whether the restore is Partial Success or Success based on replace or retain because we can get a partial success even with replace because of pre-existing cluster level resources
- [x] Added cluster context switch in `ValidateCustomResourceRestores` which was not present
- [x] Created a new function called `DeleteTorpedoApps` which will delete all the namespaces created by torpedo. This was required because there were a few namespaces created as a result of namespace mapping restore which did not have the testrail ID.

Closes #PB-7724

[Jenkins Run with OCP](https://jenkins.pwx.dev.purestorage.com/job/Users/job/Mithun/job/BYOC-%20For%20PSO/147/)


tests/backup/backup_kubevirt_test.go
```diff
@@ -3179,10 +3194,16 @@ var _ = Describe("{CustomBackupRestoreWithKubevirtAndNonKubevirtNS}", Label(Test
 			log.FailOnError(err, "Unable to fetch px-central-admin ctx")
 			log.InfoD("Total backups to restore : %v", backupNames)
 			for i, bkpName := range backupNames {
+				restoreContextMap = make(map[string][]*scheduler.Context)
 				restoreName = fmt.Sprintf("rreplace-%v-%s-%s", i, bkpName, RandomString(6))
 				log.InfoD("Restoring from the backup - [%s]", bkpName)
-				err = CreateRestoreWithReplacePolicyWithValidation(restoreName, bkpName, make(map[string]string), DestinationClusterName, BackupOrgID, ctx, make(map[string]string), 2, scheduledAppContexts)
-				dash.VerifyFatal(err, nil, fmt.Sprintf("Default restore of backups by replacing the existing resources [%s]", restoreName))
+				namespaceList, err := FetchNamespacesFromBackup(ctx, bkpName, BackupOrgID)
```
kphalgun-px, line 3200: Handle error?

mkoppal-px: Done

kphalgun-px (approved):
LGTM

PR #2715, opened 2024-08-02
Enabling storage vmotion as a longevity event

dbhatnagar-px:
Enabling Longevity events for Storage vmotion: 
1. SVMotionSingleNode -> Event to randomly choose a worker node and trigger storage vmotion
2. SVMotionMultipleNodes -> Event to randomly choose 50% of the worker nodes available (or max 20) and trigger storage vmotion for all chosen worker nodes in parallel


lsrinivas-pure (approved):
LGTM

PR #2733, opened 2024-08-13
Add Max Unavailable option for OCP surge upgrades

nilanto07-px:
Set the maxUnavailable for worker MCP to allow surge upgrades.

Closes # PTX-25469

### DIFF
--- a/drivers/scheduler/openshift/openshift.go
+++ b/drivers/scheduler/openshift/openshift.go
@@ -341,6 +341,14 @@ func updateMaxUnavailableWorkerMCP() error {
 			return nil, true, fmt.Errorf("failed to update machine config pool - worker, Err: %v %v", string(output), err)
 		}
 		log.Info(string(output))
+		args = []string{"get", "mcp", "worker", "-o", "jsonpath={.spec.maxUnavailable}"}
+		output, err := exec.Command("oc", args...).CombinedOutput()
+		if err != nil {
+			return nil, true, fmt.Errorf("failed to get maxUnavailable for worker MCP, Err: %v %v", string(output), err)
+		}
+		if string(output) != maxUnavailable {
+			return nil, true, fmt.Errorf("maxUnavailable for worker MCP does not match, expected: %s, got: %s", maxUnavailable, string(output))
+		}
 
 		log.Infof("Successfully updated maxUnavailable for worker MCP to [%s]", maxUnavailable)
 		return nil, false, nil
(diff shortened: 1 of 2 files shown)


dbhatnagar-px (approved):
Lgtm

drivers/scheduler/openshift/openshift.go
```diff
@@ -327,6 +328,29 @@ func (k *openshift) createTorpedoSecurityContextConstraints() (*ocpsecurityv1api
...
 
+func updateMaxUnavailableWorkerMCP() error {
+
+	maxUnavailable := os.Getenv("OCP_SURGE_UPGRADE_VALUE")
+	log.Info("Setting maxUnavailable value for worker MCP to [%s]", maxUnavailable)
+	patchData := fmt.Sprintf(`{"spec":{"maxUnavailable":"%s"}}`, maxUnavailable)
+
+	t := func() (interface{}, bool, error) {
+		var output []byte
+		args := []string{"patch", "mcp", "worker", "--type=merge", "--patch", patchData}
```
nikolaypopov, line 339: Does this update happens instantly? You sure we don't need to check that it actually finished updating it? I also would recommend to use API rather then CLI, I think it is better for Torpedo, since we mostly use API everywhere, unless it is not possible

nilanto07-px: The only method I could find as per Redhat is CLI - https://access.redhat.com/solutions/5453781
Even the admin acks are CLI which are called before starting the upgrade. 

I will add a check to validate if the MCP is patched.

nilanto07-px:
`15:55:52  2024-08-14 10:25:52 +0000:[INFO] [{UpgradeCluster}] [openshift.updateMaxUnavailableWorkerMCP:#334] - Setting maxUnavailable value for worker MCP to [%s] 2
15:55:52  2024-08-14 10:25:52 +0000:[INFO] [{UpgradeCluster}] [openshift.updateMaxUnavailableWorkerMCP.func1:#343] - 
15:55:52  2024-08-14 10:25:52 +0000:[INFO] [{UpgradeCluster}] [openshift.updateMaxUnavailableWorkerMCP.func1:#353] - Successfully updated maxUnavailable for worker MCP to [2]
 `
https://jenkins.pwx.dev.purestorage.com/job/Users/job/krishna/job/tp-pxe-ginkgo-byoc/25/console

nikolaypopov (approved):

PR #2739, opened 2024-08-19
fixing pool size calculation when value is in floating point

lsrinivas-pure:
fixing pool size calculation when value is in floating point
Closes # TX-25685


mkoppal-px (approved):
LGTM

PR #2745, opened 2024-08-21
PTX-25726 update maxUnavailable for OCP MCP to accept integer instead of str

nilanto07-px:
There was a regression introduced due to incorrect formatting of the `maxUnavailable` value passed during OCP upgrade. This PR fixes it by accepting `int` instead of `str`

Closes #PTX-25726

### DIFF
--- a/drivers/scheduler/openshift/openshift.go
+++ b/drivers/scheduler/openshift/openshift.go
@@ -331,8 +331,12 @@ func (k *openshift) createTorpedoSecurityContextConstraints() (*ocpsecurityv1api
 func updateMaxUnavailableWorkerMCP() error {
 
 	maxUnavailable := os.Getenv("OCP_SURGE_UPGRADE_VALUE")
+	if maxUnavailable == "" {
+		log.Warnf("OCP_SURGE_UPGRADE_VALUE is not set!")
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


lsrinivas-pure (approved):
LGTM

nikolaypopov (approved):

drivers/scheduler/openshift/openshift.go
```diff
@@ -332,7 +332,7 @@ func updateMaxUnavailableWorkerMCP() error {
 
 	maxUnavailable := os.Getenv("OCP_SURGE_UPGRADE_VALUE")
 	log.Info("Setting maxUnavailable value for worker MCP to [%s]", maxUnavailable)
-	patchData := fmt.Sprintf(`{"spec":{"maxUnavailable":"%s"}}`, maxUnavailable)
+	patchData := fmt.Sprintf(`{"spec":{"maxUnavailable":%s}}`, maxUnavailable)
```
kphalgun-px, line 335: The patch command might fail if maxUnavailable is empty.

nilanto07-px: Added return in case the value is empty.

kphalgun-px (commented):
LGTM

kphalgun-px (approved):
LGTM

PR #2746, opened 2024-08-22
PTX-25747: Counting only storage nodes for pdb validation

svijaykumar-px:
**What this PR does / why we need it**: Initially to get the total number of storage nodes, the function used included storageless nodes as well. This PR will fix the issue and only count the storagenodes for PDB validation. This includes for both cluster wide and node PDB

Closes # PTX-25747


nilanto07-px (approved):
LGTM

